PR #201, opened 2014-11-26
SLE12-SP1: initialize bootloader during update if proposed from scratch (bnc#899743)

jsrain:
Don't merge anywhere before SP1 development


jsrain:
Again checked the code, proposal client states:
        if Mode.update
          if ["grub2", "grub2-efi"].include? old_bootloader
            Builtins.y2milestone "update of grub2, do not repropose"
            if !BootCommon.was_read || @force_reset
              # blRead(reread, avoid_reading_device_map)
              Bootloader.blRead(true, true)
              BootCommon.was_read = true
            end
          else
            if !BootCommon.was_proposed || @force_reset
              # Repropose the type. A regular Reset/Propose is not enough.
              # For more details see bnc#872081
              BootCommon.setLoaderType(nil)
              Bootloader.Reset
              Bootloader.Propose
            end
          end
        else
          # in installation always propose missing stuff
          Bootloader.Propose
        end
This means that we should be set even for SP1 - if old bootloader is GRUB2, it will be updated, otherwise configuration proposed from scratch.
Please, tell me if you have further comments, otherwise I will make it merge-able.

jreidinger:
I think we need at least add "none" bootloader, as for it we do not want to overwrite user configuration if he really do not want to automanage dependencies. Otherwise it should work.

jsrain:
Good point :-)

jreidinger:
otherwise it looks good.

jreidinger:
Looks good, just please also update brand new testsuite to cover this new "none" case https://github.com/yast/yast-bootloader/blob/master/test/bootloader_proposal_client_test.rb#L144

jreidinger:
good code coverage increased. Just last fail in travis from rubocop:

14.98s$ rubocop

warning: parser/current is loading parser/ruby21, which recognizes

warning: 2.1.5-compliant syntax, but you are running 2.1.2.

Inspecting 66 files

....................................C.............................

Offenses:

src/clients/inst_bootloader.rb:35:31: C: Do not leave space between ! and its argument.

if Mode.installation || ! BootCommon.was_read

jreidinger:
you are quick. LGTM if travis passed

jsrain:
Just wanted to write that I was faster :-)

PR #204, opened 2014-12-02
New inst clients

jreidinger:
Example conversation to new inst clients.
Result is quite good according to codeclimate: https://codeclimate.com/github/yast/yast-bootloader/compare/new_inst_clients


jreidinger:
I successfully test code on real installation.

src/include/bootloader/routines/switcher.rb
```diff
@@ -73,8 +73,8 @@ def blRead(reread, avoid_reading_device_map)
     # Reset bootloader-specific settings
     # @param [Boolean] init boolean true if basic initialization of system-dependent
```
mvidner: This comment is now useless.

src/lib/bootloader/grub2base.rb
```diff
@@ -79,7 +79,7 @@ def Update
 
     # Reset bootloader settings
     # @param [Boolean] unused
```
mvidner: unused -> removed :)

jreidinger:
@mvidner both removed parameter documentation removed.

mvidner:
Thanks, LGTM.

PR #206, opened 2015-01-15
Use grub2-efi on aarch64

andreas-schwab:


jreidinger:
Thanks for contribution. I see two problems:

1) travis do not pass. Problem is in rubocop style checker easy to fix

2) I want to make sure, we want to allow grub2 and grub2-efi on aarch64 or just grub2-efi? if later one is correct, then this change is not correct and I can point where it is wrong.

andreas-schwab:
There's only grub2-efi on aarch64.

jreidinger:
@andreas-schwab Then you change is not correct. I will mark where are problems.

src/modules/BootCommon.rb
```diff
@@ -493,7 +493,7 @@ def getBootloaders
       end
       ret = [getLoaderType(false)]
       ret << "grub2" # grub2 everywhere
```
jreidinger: this is wrong if aarch support only efi, so it should be
```
ret << "grub2" unless Arch.aarch64
```
otherwise user can choose grub2 itself instead of efi variant in UI.

src/modules/BootSupportCheck.rb
```diff
@@ -86,6 +86,11 @@ def CorrectLoaderType
           return true if lt == "grub2"
         end
       end
+
+      if Arch.aarch64
+        return true if lt == "grub2-efi"
+      end
```
jreidinger: again above is problem in `return true if lt == "grub2"` as it is not correct for aarch64 for which it should not be allowed combination.

jreidinger:
Great, looks good, can you please add changes to `package/yast2-bootloader.changes` and increase version in spec, so your fix get to factory? Thanks

jreidinger:
thanks for fix

PR #207, opened 2015-02-03
Use ttyAMA instead of ttyS on aarch64

andreas-schwab:


jreidinger:
LGTM, please add changes and increase version, I then merge it.

jreidinger:
thanks for patch

PR #209, opened 2015-02-09
Add GPT PReP partition as possible location

k0da:


jreidinger:
looks good, can you please add also changelog entry?

jreidinger:
if you specify new version there, then please also increase version in `VERSION` file. It was old way in SLE-11 how to increase version.
Thanks

jreidinger:
thanks, merged

PR #211, opened 2015-02-10
Fix too much options

jreidinger:


jsrain:
LATM (A for awful), anyway, go ahead and merge it.

PR #212, opened 2015-02-10
Fix detection for encrypted partitions (bnc#913540)

ancorgs:


jreidinger:
LGTM, thanks for fixing

PR #214, opened 2015-02-11
Cherry-picking from master to sync OBS and the repo

ancorgs:


ancorgs:
Here are the OBS packages for comparison https://build.opensuse.org/package/show/openSUSE:13.2/yast2-bootloader

jreidinger:
LGTM

PR #215, opened 2015-02-11
Add test for device map dialog and fix discovered bug

jreidinger:


test/device_map_dialog_test.rb
```diff
@@ -0,0 +1,69 @@
...
+      allow(Yast::UI).to receive(:UserInput).and_return(*events)
+      allow(Yast::UI).to receive(:QueryWidget).and_return("/dev/sda")
+      allow(Yast::UI).to receive(:OpenDialog).and_return(true)
+      allow(Yast::UI).to receive(:CloseDialog).and_return(true)
+    end
+
+    it "always returns symbol :back" do
+      mock_ui_events(:ok)
+
+      expect(subject.run).to eq :back
```
ancorgs: I thought we didn't liked omitting parenthesis

jreidinger: for me in rspec this looks better as `eq` is almost `==`...also rubocop do not complain :)

test/device_map_dialog_test.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      device_map = Bootloader::DeviceMap.new(
+        "/dev/sda" => "hd0",
+        "/dev/sdb" => "hd1"
+      )
+      allow(Yast::BootStorage).to receive(:DisksOrder)
+        .and_return(device_map.disks_order)
+    end
+
+    def mock_ui_events(*events)
```
ancorgs: I would reduce this method to just the first line. Putting the rest directly in the before block including some explanation for the QueryWidget one, which is not that obvious:
```ruby
# On every iteration, the event loop checks which disk is selected in the UI
allow(Yast::UI).to receive(:QueryWidget).and_return("/dev/sda")
```

test/device_map_dialog_test.rb
ancorgs: I agree. But I have been adding the parenthesis so far because it's how we always do it. It's a grey area at https://github.com/SUSE/style-guides/blob/master/Ruby.md#method-calls

If we all agree, I'd happily stop using parenthesis with ```eq```, I just want to be sure that is common feeling.

ancorgs:
There is no direct relation between the test and the fix, isn't it?

test/device_map_dialog_test.rb
jreidinger: yes, it is grey area, as for me it is more like DSL and not common method.

jreidinger:
when I write test I observe it as it failed in one case with nil and I cannot find why.

test/device_map_dialog_test.rb
jreidinger: yes, make sense. I am considering that maybe we can add it to rspec helper to ui. In general maybe default setup of mock could be useful.

ancorgs: Well, I guess that before adding something to the rspec helpers we should agree some general approach to UI testing in YaST (we have a very short "tradition" in testing UI).

kobliha: yast-devel is the correct place to discuss this

jreidinger: writed mail to yast-devel. Will remove unrelated calls.

ancorgs:
LGTM

PR #216, opened 2015-02-12
fix redundancy boot proposal if there is more devices (bnc#917025)

jreidinger:


package/yast2-bootloader.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Feb 12 12:45:50 UTC 2015 - jreidinger@suse.com
+
+- fix redundancy boot proposal if there is more devices
```
lslezak: Bad English: ... **are** more devices.

jreidinger: right, thanks for notice

lslezak:
LGTM

jreidinger:
thanks

PR #217, opened 2015-02-13
Boot arch test

jreidinger:


src/modules/BootArch.rb
```diff
@@ -39,44 +41,29 @@ def DefaultKernelParams(resume)
...
-        end
-        ret = Ops.add(Ops.add(ret, features), " ") if features != ""
-        if Builtins.regexpmatch(ret, "^(.* )?splash=[[:lower:]]+( .*)?$")
-          ret = Builtins.regexpsub(
-            ret,
-            "^((.* ))?splash=[[:lower:]]+(( .*)?)$",
-            "\\1 \\3"
-          )
-        end
-        ret = Ops.add(ret, "splash=silent quiet showopts")
```
ancorgs: NP: If you already dupped Kernel.GetCmdLine, why use a different variable ```ret``` here instead of working with ```kernel_cmdline``` directly? It's not a big deal, anyways. Just curious.

test/boot_arch_test.rb
```diff
@@ -0,0 +1,170 @@
...
+Yast.import "BootArch"
+
+describe Yast::BootArch do
+  subject { Yast::BootArch }
+
+  def stub_arch(arch)
+    Yast.import "Arch"
+
+    allow(Yast::Arch).to receive(:architecture).and_return(arch)
+  end
```
ancorgs: I know you are not a big fan of the excessive nesting usually introduced by ```context```, but I'd say that this one of those cases in which the RSpec good practices would suggest using ```let```
```ruby
describe Yast::BootArch do
  subject { Yast::BootArch }
  before do
    allow(Yast::Arch).to receive(:architecture).and_return(arch)
  end

  describe ".method" do
    context "on x86_64" do
      let(:arch) { "x86_64" }
    end
    ...
  end
end
```

But I can live with the current approach as well.

src/modules/BootArch.rb
jreidinger: reason is that `ret` represents result and kernel cmd line is just one part of it.

test/boot_arch_test.rb
```diff
@@ -0,0 +1,170 @@
...
+      expect(subject.VgaAvailable).to eq true
+    end
+
+    it "returns true if it is on i386 architecture" do
+      stub_arch("i386")
+
+      expect(subject.VgaAvailable).to eq true
+    end
+
+    it "otherwise it returns false" do
```
ancorgs:
```ruby
it "returns false otherwise" do
```

test/boot_arch_test.rb
```diff
@@ -0,0 +1,170 @@
...
+        expect(subject.DefaultKernelParams("/dev/sda2")).to include("console=ttyS0")
+      end
+
+      it "adds additional parameters from Product file" do
+        allow(Yast::ProductFeatures).to receive(:GetStringFeature).and_return("console=ttyS0")
+
+        expect(subject.DefaultKernelParams("/dev/sda2")).to include("console=ttyS0")
+      end
+
+      it "removes splash param from command line or product file and add it silent" do
```
ancorgs: Since the removal of splash is based on a regexp, I'd appreciate having testcases for less straightforward scenarios than the parameter being at the beginning of the string.

test/boot_arch_test.rb
```diff
@@ -0,0 +1,170 @@
...
+        expect(subject.DefaultKernelParams("")).to_not include("resume")
+      end
+
+      it "adds splash=silent quit showopts parameters" do
+        expect(subject.DefaultKernelParams("/dev/sda2")).to include(" splash=silent quiet showopts")
+      end
+    end
+
+    context "on s390" do
+      before do
```
ancorgs: I miss tests about what it does NOT do. Like adding parameters from boot command line or manipulating splash

test/boot_arch_test.rb
```diff
@@ -0,0 +1,170 @@
...
+        expect(subject.DefaultKernelParams("/dev/sda2")).to include("hvc_iucv=8 TERM=dumb")
+      end
+
+      it "adds passed parameter as resume device" do
+        expect(subject.DefaultKernelParams("/dev/dasd2")).to include("resume=/dev/dasd2")
+      end
+    end
+
+    context "on other archs" do
+      before do
```
ancorgs: Same than before. I miss tests about things it does not do.

src/modules/BootArch.rb
ancorgs: Right. Makes sense.

test/boot_arch_test.rb
jreidinger: for me it is a bit overkill as many methods are just plain if depending on architecture.

test/boot_arch_test.rb
jreidinger: I add test case for boot command line. Splash is useless as on s390 it cannot be used as it do not boot in graphical mode.

ancorgs:
LGTM

PR #219, opened 2015-02-17
Fix flags

jreidinger:


src/lib/bootloader/mbr_update.rb
```diff
@@ -88,12 +88,32 @@ def install_generic_mbr
...
+
+      # and then set it
       command = "/usr/sbin/parted -s #{disk} set #{part_num} #{flag} on"
       out = Yast::WFM.Execute(Yast::Path.new(".local.bash_output"), command)
       log.info "Command `#{command}` output: #{out}"
       out
     end
 
+    def reset_flag(disk, flag)
+      command = "/usr/sbin/parted -s #{disk} print"
```
aschnell: We once had a bug report dealing with spaces in the device name so better use quotes.

You might considere using the -m switch to get machine parsable output from parted.

src/lib/bootloader/mbr_update.rb
```diff
@@ -88,12 +88,32 @@ def install_generic_mbr
...
+      out = Yast::WFM.Execute(Yast::Path.new(".local.bash_output"), command)
```
aschnell: AFAIK path(".local.bash_output") also works and is shorter.

jreidinger: this class does not live in Yast namespace, so there is not path shortcut, so full path specification is needed.

src/lib/bootloader/mbr_update.rb
jreidinger: good note, I will add it.

src/lib/bootloader/mbr_update.rb
```diff
@@ -88,12 +88,32 @@ def install_generic_mbr
...
+      log.info "Command `#{command}` output: #{out}"
+      return if out["exit"] != 0
+
+      partitions = out["stdout"].lines.grep(/\s#{flag}/)
```
aschnell: This is dangerous since a user can set an arbitrary partition name on GPT, so the output could be:
<pre>Number  Start   End    Size   File system  Name         Flags
 1      1049kB  822MB  821MB  btrfs        legacy_boot  boot</pre>

jreidinger: ok, with machine output it should be easily fixable.

aschnell:
LGTM

jreidinger:
@arvin thanks

PR #220, opened 2015-02-18
add more logging for autoyast import

jreidinger:


src/include/bootloader/routines/autoinstall.rb
```diff
@@ -130,6 +130,7 @@ def AI2Export(ai)
         # FIXME: this should be better handled by exception and show it properly, but it require too big change now
         Popup.Error(_("Unsupported bootloader '%s'. Adapt your AutoYaST profile accordingly.") %
  exp["loader_type"])
+        Builtins.y2error("unsupported bootloader #{exp["loader_type"]}")
```
mchf: Well, I'd prefer not to add new call to ```Builtins```.

mchf:
Despite of one comment LGTM.

src/include/bootloader/routines/autoinstall.rb
jreidinger: autoinstall code do not yet use new logging system and to be honest I am quite scared to touch almost anything in autoyast part of bootloader as it is full of workarounds and backward compatibility :)

PR #221, opened 2015-02-18
Add partial test for bootloader module

jreidinger:
Fix flags

### DIFF
--- a/test/bootloader_test.rb
+++ b/test/bootloader_test.rb
@@ -6,6 +6,15 @@ describe Yast::Bootloader do
   subject { Yast::Bootloader }
 
   describe ".Import" do
+
+    before do
+      # ensure flags are reset
+      Yast::BootCommon.was_read = false
+      Yast::BootCommon.was_proposed = false
+      Yast::BootCommon.changed = false
+      Yast::BootCommon.location_changed = false
+    end
+
     # FIXME: looks useless as import is used in autoinstallation and in such case reset do nothing
     it "resets configuration" do
       expect(subject).to receive(:Reset)
@@ -80,8 +89,6 @@ describe Yast::Bootloader do
       allow(subject).to receive(:Read)
       Yast::BootCommon.was_read = false
       Yast::BootCommon.was_proposed = false
-
-      allow(subject).to receive(:UpdateConfiguration)
     end
 
     it "does nothing if already read" do
@@ -131,6 +138,7 @@ describe Yast::Bootloader do
       expect(subject).to receive(:Read)
       expect(Yast::Stage).to receive(:initial).and_return(true)
       allow(Yast::Mode).to receive(:update).and_return(true)
+      allow(subject).to receive(:UpdateConfiguration)
 
       subject.ReadOrProposeIfNeeded
     end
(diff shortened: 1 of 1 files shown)


test/bootloader_test.rb
```diff
@@ -0,0 +1,146 @@
...
+    it "marks that configuration is read" do
+      subject.Import({})
+
+      expect(Yast::BootCommon.was_read).to eq true
+    end
+
+    it "marks that configuration is already proposed" do
+      subject.Import({})
+
+      expect(Yast::BootCommon.was_proposed).to eq true
```
ancorgs: Shouldn't we check that it was false before?

test/bootloader_test.rb
```diff
@@ -0,0 +1,146 @@
...
+      subject.Import({})
+    end
+
+    it "acts like missing if \"loader_type\" value is empty" do
+      expect(Yast::BootCommon).to receive(:setLoaderType).with(Yast::BootCommon.getLoaderType(true))
+      subject.Import("loader_type" => "")
+    end
+
+    it "pass initrd specific map to initrd module" do
+      initrd_map = {
```
ancorgs: A simpler map would have been enough to check that it's passed :-D

test/bootloader_test.rb
jreidinger: it is by default false.

test/bootloader_test.rb
jreidinger: I want to use some real one, with real values, as it is not trivial to find how it should look like :)

test/bootloader_test.rb
```diff
@@ -0,0 +1,146 @@
...
+    it "reads configuration in update mode" do
+      expect(subject).to_not receive(:Propose)
+      expect(subject).to receive(:Read)
+      expect(Yast::Stage).to receive(:initial).and_return(true)
+      allow(Yast::Mode).to receive(:update).and_return(true)
+
+      subject.ReadOrProposeIfNeeded
+    end
+
+    it "calls .UpdateConfiguration in update mode" do
```
ancorgs: Is this only relevant in update mode? I mean, it would be good/bad if it calls ```.UpdateConfiguration``` in normal mode?

test/bootloader_test.rb
ancorgs: So part of your "tests as documentation" initiative. I see. Good point.

test/bootloader_test.rb
ancorgs: Even though, from a unit testing POV, if we are not enforcing/checking the previous value we cannot actually assert that "it marks that configuration is read".

test/bootloader_test.rb
jreidinger: yes, only in update mode. I think now for grub2 it do nothing, I just want to document it :)

ancorgs: Then I will move the ```allow(subject).to receive(:UpdateConfiguration)``` from the before block and will put it only in the  "reads configuration in update mode". This way it would be more obvious (for instance, I wouldn't have asked).

ancorgs:
LGTM

PR #222, opened 2015-02-18
Fix ppc install

jreidinger:


jreidinger:
verified in real installation on ppc

ancorgs:
Just wondering if the change would not affect nfs installation. I have not gone through BootCommon code, I just worry that not calling BootCommon.DetectDisks can alter the result of BootCommon.getBootDisk. If it's not the case (once again, I have not looked at it so I trust you), LGTM.

jreidinger:
@ancor `BootCommon.DetectDisks` call this detect_disks and then if needed repropose configuration location. but getBootDisk is set by `detect_disks`

PR #223, opened 2015-02-24
temporary disable failing test which failing only in remote build servic...

jreidinger:
...e


ancorgs:
Rubocop is still unhappy.

On the other hand, wouldn't it be better to mark it as pending than simply commenting it out completely? The latter seems easier to forget about. With the former you would see a reminder in every ```rake test:unit```

jreidinger:
@ancorgs agreed, lets make it pending

ancorgs:
LGTM

PR #224, opened 2015-02-25
Storage tests

jreidinger:


package/yast2-bootloader.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Feb 25 21:48:10 UTC 2015 - jreidinger@suse.com
+
+- Fix creation of multipath device map
```
kobliha: Is there a bug number for that? I'm missing the info in PR description: Which products are affected? None? This is important from maintenance point of view.

So, PLS either add bug number or summarize in PR description. Anyway, good PR description does never hurt anyway.

jreidinger: There is no bug, I found it when I play with tests. No product is affected, bug is introduced by one of cleaning when udev device mapping is moved to own class.

kobliha: Then PLS write at least "Fixed creation of a multipath device map" :)

test/boot_storage_test.rb
```diff
@@ -92,4 +92,34 @@
...
+
+    it "returns empty map if there is no multipath" do
+      target_map_stub("storage_lvm.rb")
+
+      # init variables
+      subject.InitDiskInfo
+      expect(subject.multipath_mapping).to eq({})
+    end
+
+    it "return map of kernel disks to multipath devices" do
```
kobliha: "return map of kernel disks to multipath devices"?

1. returns
2. what is a kernel disk?

test/boot_storage_test.rb
```diff
@@ -92,4 +92,34 @@
...
+      # mock getting mount points as it need whole libstorage initialization
+      allow(Yast::Storage).to receive(:GetMountPoints).and_return("/" => "/dev/vda1")
+      # mock for same reason getting udev mapping
+      allow(::Bootloader::UdevMapping).to receive(:to_mountby_device) do |arg|
+        arg
+      end
+    end
+
+    it "returns empty map if there is no multipath" do
+      target_map_stub("storage_lvm.rb")
```
kobliha: Misusing storage_lvm.rb might have a side effect in the future. Someone could change storage_lvm.rb to contain multipath and that will break this test. The probability is not that high, but you hopefully see my point.

jreidinger: yep, it is my own data got from customer or testing, so it won't change in future. I still hope in future to have generator for target map

test/boot_storage_test.rb
jreidinger: thanks, kernel disk I mean kernel name of disk device

jreidinger:
@kobliha updated

test/boot_storage_test.rb
kobliha: Sure, I know it's your "own" data :) But we don't own the code, we own the responsibility. Someone else (not you) could change the data in the future. As I said, the probability is not high.

kobliha:
LGTM

test/boot_storage_test.rb
jreidinger: yeah, "own" is wrong term, I mean it is real data to prevent regressions, so unless target map format is changed, then it should not be changed.

jreidinger:
Thanks for review

PR #228, opened 2015-03-18
Small cleaning

jreidinger:


src/include/bootloader/routines/misc.rb
```diff
@@ -227,32 +220,6 @@ def restoreMBR(device)
...
-          p["mount"] == "swap" && !p["delete"]
-        end
-        partitions.each do |s|
-          # bnc#577127 - Encrypted swap is not properly set up as resume device
-          if s["crypt_device"] && !s["crypt_device"].empty?
-            dev = s["crypt_device"]
-          else
-            dev = s["device"]
-          end
-          ret[dev] = Ops.get_integer(s, ["region", 1], 0) * cyl_size
```
ancorgs: I guess this code was there before partition["size_k"] was introduced in the target map. Since there is no comment stating the contrary, I'll assume it can be safely replaced by v["size_k"].

ancorgs:
LGTM

src/include/bootloader/routines/misc.rb
jreidinger: yep, it is also my suspicious.

jreidinger:
thanks

PR #229, opened 2015-03-24
Propose secure_boot by default only on x86

andreas-schwab:


jreidinger:
what is reason to not propose secure boot on other architectures? ( I think only arm is now affected )

andreas-schwab:
"/usr/sbin/shim-install: no such file or directory" from perl-bootloader

jreidinger:
so it means that it is not valid option on arm?

andreas-schwab:
Not for now.

jreidinger:
It changing only proposal, so in UI it is still possible to enable it?

andreas-schwab:
There is currently no way to toggle secure boot in the UI

jreidinger:
OK, so LGTM

jreidinger:
Please add entry to changelog and increase version in spec ile. Then I merge it.
Thanks

PR #231, opened 2015-04-14
[Review] Request from 'schubi2' @ 'yast/yast-bootloader/review_150414_reading_mountpoints_correctly'

schubi2:
Please review the following changes:
  * 4d14b64 packaging
  * fe1af7e reading mountpoints correctly while clone_system

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Apr 14 14:50:43 CEST 2015 - schubi@suse.de
+
+- While calling AutoYaST clone_system libStorage has to be set
+  to "normal" mode in order to read mountpoints correctly.
+- 3.1.127
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jreidinger:
code wise ok, but why not fix it in storage?

schubi2:
I have asked Arvin. The problem is that in this case the storage module cannot make this decision. We have already had this discussion about one year ago. At that time I have made the same suggestion but you have found another one. This solution does not work anymore. The problem is, that I do not know your solution anymore :-)

schubi2:
So I am open for every other solution ;-)

schubi2:
That's the point here:
https://github.com/yast/yast-storage/blob/master/src/modules/Storage.rb#L4450

jreidinger:
hmm, ok so problem is that bootloader need to know about root partition and on the other hand storage, do not read it in cofig, as it should be proposed for bootloader. So problem is that bootloader proposed default is based on storage that maybe will be different on target machine.
For now I think your fix is OK, but there is no quaranty that final profile will work everywhere.

PR #233, opened 2015-05-20
Disable os-prober for Power

k0da:
os-prober is slow. Disable it on Power. boo#931653

### DIFF
--- a/src/lib/bootloader/grub2base.rb
+++ b/src/lib/bootloader/grub2base.rb
@@ -56,7 +56,8 @@ module Yast
     # Propose global options of bootloader
     def StandardGlobals
       # s390 do not have os_prober, see bnc#868909#c2
-      disable_os_prober = Arch.s390 || ProductFeatures.GetBooleanFeature("globals", "disable_os_prober")
+      # ppc have slow os_prober, see boo#931653
+      disable_os_prober = (Arch.s390||Arch.ppc) || ProductFeatures.GetBooleanFeature("globals", "disable_os_prober")
       {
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/grub2base.rb
```diff
@@ -56,7 +56,7 @@ def pmbr_setup(action, *devices)
     # Propose global options of bootloader
     def StandardGlobals
       # s390 do not have os_prober, see bnc#868909#c2
-      disable_os_prober = Arch.s390 || ProductFeatures.GetBooleanFeature("globals", "disable_os_prober")
+      disable_os_prober = (Arch.s390||Arch.ppc) || ProductFeatures.GetBooleanFeature("globals", "disable_os_prober")
```
jreidinger: Does it make sense? For SLES global feature is set to false, so your change affected only SLED and opensuse ( opensuse is not for ppc ), so only SLED on ppc, do we release it there at all?

k0da: This change is for openSUSE.
opensuse is not for ppc??

jreidinger: I do not see oficial support for it. If it is goal to have support for ppc in opensuse I am fine with this change. just please add reference to ppc bug in comment above similar like I did for s390, so in future we know why it is there and have reference.

k0da: Updated.

We don't need official support. It is done in Ports

jreidinger:
Thanks

PR #235, opened 2015-05-21
ensure libstorage runs in 'normal' mode so it really returns us the m…

wfeldt:
…ountpoints (bsc #929656)


wfeldt:
this seems to fix part 1 of the problem: we how have the mountpoints

wfeldt:
but there's more that's wrong

jreidinger:
For code 11 it looks OK.

wfeldt:
The change will not be needed in sle11-sp4. See discussion in bugzilla.

PR #236, opened 2015-06-01
Grub1 crash

jreidinger:


src/include/bootloader/routines/dialogs.rb
```diff
@@ -128,7 +128,7 @@ def MainDialog
 
       )
       if ret != :back && ret != :abort && ret != :cancel
-        @return_tab = CWMTab.LastTab
+        @return_tab = CWMTab.LastTab || "tab"
```
mvidner: So "tab" is both the id of an existing Tab widget, and a fake name of a non-existing tab?

jreidinger: Yes, it is a bit tricky. In grub2 due to different design it do not exists, so it is replaced by installation tab. So I use it here as default value.

mvidner: So, could we use "fake_tab" here...

src/include/bootloader/routines/dialogs.rb
```diff
@@ -128,7 +128,7 @@ def MainDialog
...
         @return_tab = "installation" if @return_tab.include? "tab" # workaround different tab set for grub2
```
mvidner: and "_tab" here?

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Jun  1 14:20:30 UTC 2015 - jreidinger@suse.com
+
+- do not crash if last tab do not exists like for non-supported
+  bootloader (bnc#923458)
```
mvidner: This is too implementation oriented. What are the user visible circumstances? The bug title "Startup of YaST bootloader fails" is also too generic to be helpful.

src/include/bootloader/routines/dialogs.rb
jreidinger: no as `fake_tab` do not exist, it should be existing id value

src/include/bootloader/routines/dialogs.rb
jreidinger: no as tab is proposed as initial tag elsewhere in code. So different part of code propose as default return_tab  "tab" and here it is changed to grub2 default return tab..it is just relict of old multi-bootloader code that have here condition if `bootloader == "grub2" &&` which is now useless. It deserve more cleaning but it is not part of this pr and I hope it will be cleaned together with perl-bootloader support drop off

package/yast2-bootloader.changes
jreidinger: it should for unsupported bootloader that it do not show red popup with crash :)

jreidinger:
LGTMed on IRC

PR #239, opened 2015-06-03
Fix cleaning

jreidinger:


imobach:
LGTM

mvidner:
@jreidinger This bug points to poor design of `TmpYAMLFile`, where you must not forget the `ensure` and also consider whether the variable has been initialized already. IMHO the right way is to model this after [Tempfile.open](http://www.rubydoc.info/stdlib/tempfile/Tempfile.open):
```ruby
TmpYAMLFile.open(contents = nil) do |y|
   ... y.path ... y. data ...
end
```

jreidinger:
@mvidner agreed, but because this part of communication going to be removed as part of perl-bootloader removal, I do not want to spend there too much time.

mvidner:
@jreidinger hmm, but then in the SLE12 branch I see https://github.com/yast/yast-bootloader/commit/ca87ec86ad6965ab09fcc31bc7a5d16f28fcdfe9 that surfaced in a real bug https://bugzilla.suse.com/show_bug.cgi?id=927237 so this may be worth fixing after all? Let's see after that bug is done.

PR #241, opened 2015-06-04
13.2 backport tmp cleaning

jreidinger:


mvidner:
"Backport" is not very helpful, "backport of #239" is better :)

In the bug they say an update can be released, so you should bump the version.

mvidner:
But if you really don't want to release it then it LGTM.

jreidinger:
@mvidner thanks

PR #243, opened 2015-06-11
Unrestricted boot

jreidinger:
Do not merge yet as grub2 is not yet ready. changelog update will be added later to avoid conflicts.

### DIFF
--- a/src/include/bootloader/grub/helps.rb
+++ b/src/include/bootloader/grub/helps.rb
@@ -83,9 +83,9 @@ module Yast
             "At boot time, modifying or even booting any entry will require the" \
             " password. If <b>Protect Entry Modification Only</b> is checked then " \
             "booting any entry is not restricted but modifying entries requires " \
-            "the password.<br>" \
+            "the password (which is the way GRUB 1 behaved).<br>" \
             "YaST will only accept the password if you repeat it in " \
-            "<b>Retype Password</b>."
+            "<b>Retype Password</b>.</p>"
         ),
         # help text 1/5
         "disk_order"        => _(
--- a/src/include/bootloader/grub/helps.rb
+++ b/src/include/bootloader/grub/helps.rb
@@ -80,12 +80,12 @@ module Yast
         ),
         "password"          => _(
           "<p><b>Protect Boot Loader with Password</b><br>\n" \
-            "Define the password that will be required to access the boot menu. \n" \
-            "YaST will only accept the password if you repeat\n" \
-            "it in <b>Retype Password</b>. If <b>Allow to boot locked default \n" \
-            "entry without password</b> is checked, then without password is \n " \
-            "possible to let bootloader autoboot default entry with its kernel \n" \
-            "parameters.</p>"
+            "At boot time, modifying or even booting any entry will require the" \
+            " password. If <b>Protect Entry Modification Only</b> is checked then " \
+            "booting any entry is not restricted but modifying entries requires " \
+            "the password.<br>" \
+            "YaST will only accept the password if you repeat it in " \
+            "<b>Retype Password</b>."
         ),
         # help text 1/5
         "disk_order"        => _(
--- a/src/include/bootloader/grub/helps.rb
+++ b/src/include/bootloader/grub/helps.rb
@@ -80,8 +80,12 @@ module Yast
         ),
         "password"          => _(
           "<p><b>Protect Boot Loader with Password</b><br>\n" \
-            "Define the password that will be required to access the boot menu. YaST will only accept the password if you repeat\n" \
-            "it in <b>Retype Password</b>.</p>"
+            "Define the password that will be required to access the boot menu. \n" \
+            "YaST will only accept the password if you repeat\n" \
+            "it in <b>Retype Password</b>. If <b>Allow to boot locked default \n" \
+            "entry without password</b> is checked, then without password is \n " \
+            "possible to let bootloader autoboot default entry with its kernel \n" \
+            "parameters.</p>"
         ),
         # help text 1/5
         "disk_order"        => _(
--- a/src/include/bootloader/grub/options.rb
+++ b/src/include/bootloader/grub/options.rb
@@ -40,7 +40,7 @@ module Yast
       event = deep_copy(event)
       if Ops.get(event, "ID") == :use_pas
         enabled = Convert.to_boolean(UI.QueryWidget(Id(:use_pas), :Value))
-        UI.ChangeWidget(Id(:restricted_pw), :Enabled, enabled)
+        UI.ChangeWidget(Id(:unrestricted_pw), :Enabled, enabled)
         UI.ChangeWidget(Id(:pw1), :Enabled, enabled)
         UI.ChangeWidget(Id(:pw2), :Enabled, enabled)
       end
@@ -76,7 +76,7 @@ module Yast
             HBox(
               HSpacing(2),
               # TRANSLATORS: checkbox entry
-              CheckBox(Id(:restricted_pw), _("A&llow to boot locked default entry without password")),
+              CheckBox(Id(:unrestricted_pw), _("A&llow to boot locked default entry without password")),
               HStretch()
             ),
             HBox(
--- a/src/include/bootloader/grub/options.rb
+++ b/src/include/bootloader/grub/options.rb
@@ -40,6 +40,7 @@ module Yast
       event = deep_copy(event)
       if Ops.get(event, "ID") == :use_pas
         enabled = Convert.to_boolean(UI.QueryWidget(Id(:use_pas), :Value))
+        UI.ChangeWidget(Id(:restricted_pw), :Enabled, enabled)
         UI.ChangeWidget(Id(:pw1), :Enabled, enabled)
         UI.ChangeWidget(Id(:pw2), :Enabled, enabled)
       end
@@ -71,14 +72,22 @@ module Yast
           Id(:use_pas),
           _("Prot&ect Boot Loader with Password"),
           true,
-          HBox(
-            HSpacing(2),
-            # text entry
-            Password(Id(:pw1), Opt(:hstretch), _("&Password")),
-            # text entry
-            HSpacing(2),
-            Password(Id(:pw2), Opt(:hstretch), _("Re&type Password")),
-            HStretch()
+          VBox(
+            HBox(
+              HSpacing(2),
+              # TRANSLATORS: checkbox entry
+              CheckBox(Id(:restricted_pw), _("A&llow to boot locked default entry without password")),
+              HStretch()
+            ),
+            HBox(
+              HSpacing(2),
+              # text entry
+              Password(Id(:pw1), Opt(:hstretch), _("&Password")),
+              # text entry
+              HSpacing(2),
+              Password(Id(:pw2), Opt(:hstretch), _("Re&type Password")),
+              HStretch()
+            )
           )
         )
       )
--- a/src/include/bootloader/grub/options.rb
+++ b/src/include/bootloader/grub/options.rb
@@ -76,7 +76,7 @@ module Yast
             HBox(
               HSpacing(2),
               # TRANSLATORS: checkbox entry
-              CheckBox(Id(:unrestricted_pw), _("A&llow to boot locked default entry without password")),
+              CheckBox(Id(:unrestricted_pw), _("P&rotect Entry Modification Only")),
               HStretch()
             ),
             HBox(
--- a/src/lib/bootloader/grub2base.rb
+++ b/src/lib/bootloader/grub2base.rb
@@ -6,7 +6,8 @@ require "bootloader/udev_mapping"
 module Yast
   # Common base for GRUB2 specialized classes
   class GRUB2Base < Module
-    # stored password object
+    # @!attribute password
+    #    @return [::Bootloader::GRUB2Pwd] stored password configuration object
     attr_reader :password
 
     def main
--- a/src/lib/bootloader/grub2base.rb
+++ b/src/lib/bootloader/grub2base.rb
@@ -33,7 +33,7 @@ module Yast
       # for simplified widgets than other
       Yast.include self, "bootloader/grub2/dialogs.rb"
 
-      @password = Bootloader::GRUB2Pwd.new
+      @password = ::Bootloader::GRUB2Pwd.new
     end
 
     # general functions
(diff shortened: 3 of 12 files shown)


mvidner:
> changelog update will be added later to avoid conflicts

Then you should state the intended changelog in the PR description, to get it reviewed. And I wouldn't have to ask for fate/bugzilla refrerence now ;-)

jreidinger:
changelog is something like
```
- allow to select grub1 password protection behavior (fate#318574)
```

mvidner:
> to select grub1 password protection
which version? "GRUB 2 " ?

src/lib/bootloader/grub2base.rb
```diff
@@ -6,6 +6,9 @@
 module Yast
   # Common base for GRUB2 specialized classes
   class GRUB2Base < Module
+    # stored password object
```
mvidner: \# @return [::Bootloader::GRUB2Pwd] stored password object

src/lib/bootloader/grub2pwd.rb
```diff
@@ -1,63 +1,131 @@
...
-  YAST_BASH_PATH = Yast::Path.new(".target.bash_output")
-  PWD_ENCRYPTION_FILE = "/etc/grub.d/42_password"
+Yast.import "Stage"
 
-  def used?
-    Yast.import "FileUtils"
+module Bootloader
+  # class is responsible for detection, encryption and writing of grub2 password protection
+  class GRUB2Pwd
+    attr_accessor :used, :unrestricted
```
mvidner: Please document that they are Boolean.
Password being "used" is IMHO obvious but "unrestricted" is not, and this is a good place to document what we mean by it.

src/lib/bootloader/grub2pwd.rb
```diff
@@ -1,63 +1,131 @@
...
+    alias_method :used?, :used
+    alias_method :unrestricted?, :unrestricted
 
-    Yast::FileUtils.Exists PWD_ENCRYPTION_FILE
-  end
+    def initialize
+      # TODO: offline upgrade should somehow read it when we need grub2 -> grub2 modifications
```
mvidner: grub1 -> grub2 ?

src/lib/bootloader/grub2pwd.rb
```diff
@@ -1,63 +1,131 @@
...
+      if used?
+        enable
+      else
+        disable
+      end
+    end
 
-  def disable
-    return unless used?
+    def password=(value)
```
mvidner: So, I can set a password, but not query its value, only its existence. OK, fair enough. But this is separated from the `used` switch? Shouldn't they be linked?

src/lib/bootloader/grub2pwd.rb
```diff
@@ -1,63 +1,131 @@
...
 
-    if result["exit"] != 0
-      raise "Failed to create encrypted password for grub2. Command output: #{result["stderr"]}"
+      @encrypted_password = pwd_line[/password_pbkdf2 root (\S+)/, 1]
     end
 
-    pwd_line = result["stdout"].split("\n").grep(/password is/).first
-    if !pwd_line
-      raise "INTERNAL ERROR: output do not contain encrypted password. Output: #{result["stdout"]}"
+    def read_used
```
mvidner: rename to `read_used?` to make clear it's a Boolean

src/lib/bootloader/grub2pwd.rb
jreidinger: for grub1 -> grub2 we do repropose, password config is lost and cannot be converted as it use different hashing algorithm

src/lib/bootloader/grub2pwd.rb
jreidinger: reason is that enabling password do not necessary mean, that you specify password. So used != password set...also it mean when in UI you set used = false and then go back and set used = true, it remember old password which is set to widgets.

src/lib/bootloader/grub2pwd.rb
jreidinger: yep, I forgot to write api documentation, my fault.

mvidner:
- Let password protection be configurable between a restricted
  mode (cannot boot at all without password, default GRUB2
  behavior) and an unrestricted mode (can boot but cannot select
  or edit entries, GRUB1 behavior) (FATE#318574).

src/lib/bootloader/grub2pwd.rb
```diff
@@ -1,63 +1,131 @@
...
+      Yast::FileUtils.Exists PWD_ENCRYPTION_FILE
     end
 
-    ret = pwd_line[/^.*password is\s*(\S+)/, 1]
-    if !ret
-      raise "INTERNAL ERROR: output do not contain encrypted password. Output: #{result["stdout"]}"
+    def enable
+      raise "Wrong code: password not written" unless @encrypted_password
+
+      file_content = "#! /bin/sh\n" \
```
mvidner: > \# The files in /etc/grub.d are programs that write GRUB 2 programs on their stdout.
\# So 'exec tail' is a way of saying "just echo the rest of this program as its output".

mvidner: I mean this as a comment in the rb file, not in the config file (or its output!!)

src/lib/bootloader/grub2pwd.rb
mvidner: So the intent is "read the value for the attribute `used` from permanent storage but don't assign it". 

The yast convention is that Read should assign the attributes (and the preceding methods conforms to that).

Maybe we can set a convention that Read without assigning is called Load. -> `load_used`

src/lib/bootloader/grub2pwd.rb
mvidner: So the state "password is set but disabled" can exist, but only in the UI, not in the config files, right?

src/lib/bootloader/grub2pwd.rb
```diff
@@ -1,63 +1,146 @@
...
-  def used?
-    Yast.import "FileUtils"
+module Bootloader
+  # class is responsible for detection, encryption and writing of grub2 password protection
+  class GRUB2Pwd
+    # @!attribute used
+    #   @return [bool] specifies if password protection enabled
+    #
+    # @!attribute unrestricted
+    #   @return [bool] specifies if unrestricted password protection should be used
```
mvidner: "Boolean", not "bool".
Again, somewhere we should say what "unrestricted" means and I think it should be here.

jreidinger: even Boolean is not known class in ruby, there is only TrueClass and FalseClass.
Well, unrestricted is official grub2 term. I expect that this code will be modified only by someone who knows grub2 terminology.

src/include/bootloader/grub/options.rb
```diff
@@ -71,14 +72,22 @@ def passwd_content
...
-            Password(Id(:pw1), Opt(:hstretch), _("&Password")),
-            # text entry
-            HSpacing(2),
-            Password(Id(:pw2), Opt(:hstretch), _("Re&type Password")),
-            HStretch()
+          VBox(
+            HBox(
+              HSpacing(2),
+              # TRANSLATORS: checkbox entry
+              CheckBox(Id(:unrestricted_pw), _("A&llow to boot locked default entry without password")),
```
mvidner: Why are we talking about a "locked" default entry? I thought being locked is a global property for all entries together, controlled by the "Password" checkbox.

test/grub2pwd_test.rb
```diff
@@ -55,52 +221,47 @@ def mock_file_presence(exists)
...
-    it "raise exception if grub2-mkpasswd-pbkdf create password line but without password" do
-      expect(Yast::WFM).to receive(:Execute)
-        .with(kind_of(Yast::Path), /grub2-mkpasswd/)
-        .and_return(
-          "exit"   => 0,
-          "stderr" => "",
-          "stdout" => "password is"
-        )
-      expect(Yast::SCR).to receive(:Write).never
+    it "returns true if password explicitelly set" do
```
mvidner: explicitly

src/lib/bootloader/grub2pwd.rb
mvidner: Yes, but Boolean is a YARD convention whereas bool is not: http://www.rubydoc.info/gems/yard/file/docs/GettingStarted.md#Declaring_Types

Yes, "unrestricted" is an official grub2 term, but it is local to a single entry whereas we use it in a global meaning. Even if I read the docs I will not know what it means is SUSE: https://www.gnu.org/software/grub/manual/html_node/Security.html#Security

src/lib/bootloader/grub2pwd.rb
```diff
@@ -1,63 +1,146 @@
...
+    attr_accessor :used, :unrestricted
+    alias_method :used?, :used
+    alias_method :unrestricted?, :unrestricted
 
-    Yast::FileUtils.Exists PWD_ENCRYPTION_FILE
-  end
+    # Reads or proposes configuration depending on stage
+    def initialize
+      # TODO: offline upgrade should somehow read it when we need grub2 -> grub2 modifications
```
mvidner: IMHO this comment is anticipating future requirements, YAGNI, delete it.

src/lib/bootloader/grub2pwd.rb
```diff
@@ -1,63 +1,146 @@
...
+    end
+
+  private
+
+    YAST_BASH_PATH = Yast::Path.new(".target.bash_output")
+    PWD_ENCRYPTION_FILE = "/etc/grub.d/42_password"
+
+    def propose
+      @used = false
+      @unrestricted = false # TODO: ensure it in FATE
```
mvidner: ... whereas this TODO should be resolved before merging. (and then deleted).

src/include/bootloader/grub/options.rb
jreidinger: I want to emphasis that it is only default entry without possibility to modify kernel parameters

src/lib/bootloader/grub2pwd.rb
jreidinger: fate is already in needinfo state so I need to wait

src/include/bootloader/grub/helps.rb
```diff
@@ -80,8 +80,12 @@ def initialize_bootloader_grub_helps(_include_target)
...
         "password"          => _(
           "<p><b>Protect Boot Loader with Password</b><br>\n" \
-            "Define the password that will be required to access the boot menu. YaST will only accept the password if you repeat\n" \
-            "it in <b>Retype Password</b>.</p>"
+            "Define the password that will be required to access the boot menu. \n" \
+            "YaST will only accept the password if you repeat\n" \
+            "it in <b>Retype Password</b>. If <b>Allow to boot locked default \n" \
+            "entry without password</b> is checked, then without password is \n " \
+            "possible to let bootloader autoboot default entry with its kernel \n" \
+            "parameters.</p>"
```
mvidner: > **Protect Boot Loader with Password**
At boot time, modifying or even booting any entry will require the password. If **Allow to Boot Default Entry** is checked then that action will not require the password.
YaST will only accept the password if you repeat it in **Retype Password.**

(is Entry the correct term? Section?)

arvidjaar: > If Allow to Boot Default Entry is checked then that action will not require the password.
May be it helps to understand what GRUB2 actually implements.

You can enable authentication in GRUB2 (by setting superusers variable to list of user names). As soon as superusers variable is defined, entering CLI, editing menu entry and executing menu entry (i.e. usually booting) requires authentication as one of users in this list.

It is possible to either provide additional list of users that are allowed to execute individual menu entry (but not to edit it) or mark individual menu entry as "unrestricted" so executing it does not require authentication.

What currently is *NOT* possible is to autoboot default entry while keeping password protection for it. Either entry must be marked as unrestricted (and so always allowed to boot, manually or auto) or you need to enter password both for manual or autoboot.

Looking at it, autoboot without password looks like sensible feature request.

jreidinger: Well, request from our customer is to option to have back grub1 pwd protection, I need to verify how exactly it works in grub1 and compare it to testing grub2 package that michael provide to me.

jreidinger: I check grub1 feature and it also allow to boot any boot entry, only lock kernel parameters

jreidinger: and for grub2 it behave same way. @mvidner can you look at it and propose possibly new labels and help text for it, that make better sense? something like
```
label = "Protect entry modification only"
helptext = "<b> Protect entry modification only</b> protects by password only modification of entry like loading different kernel or modify kernel parameters."
```

mvidner: @arvidjaar Thanks for the clarification! We are implementing a GRUB1 compatibility switch and I originally thought it did what I described above. Here's a revised description:

> **Protect Boot Loader with Password**
At boot time, modifying or even booting any entry will require the password. If **Protect Entry Modification Only** is checked then booting any entry is not restricted but modifying entries requires the password.
YaST will only accept the password if you repeat it in **Retype Password.**

src/lib/bootloader/grub2pwd.rb
mvidner: Still pending.

mvidner:
Code looks good but the changelog entry is missing:
> - Let password protection be configurable between a restricted
  mode (cannot boot at all without password, default GRUB2
  behavior) and an unrestricted mode (can boot but cannot edit entries,
  GRUB1 behavior) (FATE#318574).

jreidinger:
for changelog I still waiting for michael to submit new grub2 to SP1, when he add it I also I add it.

src/include/bootloader/grub/helps.rb
```diff
@@ -80,8 +80,12 @@ def initialize_bootloader_grub_helps(_include_target)
         ),
         "password"          => _(
           "<p><b>Protect Boot Loader with Password</b><br>\n" \
-            "Define the password that will be required to access the boot menu. YaST will only accept the password if you repeat\n" \
-            "it in <b>Retype Password</b>.</p>"
+            "At boot time, modifying or even booting any entry will require the" \
+            " password. If <b>Protect Entry Modification Only</b> is checked then " \
+            "booting any entry is not restricted but modifying entries requires " \
+            "the password.<br>" \
```
mvidner: "the password (which is the way GRUB 1 behaved).\<br>"

src/include/bootloader/grub/helps.rb
```diff
@@ -80,8 +80,12 @@ def initialize_bootloader_grub_helps(_include_target)
...
+            "YaST will only accept the password if you repeat it in " \
+            "<b>Retype Password</b>."
```
mvidner: \</p>

mvidner:
LGTM!

PR #245, opened 2015-06-29
Correct boot flag

jreidinger:


package/yast2-bootloader.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Jun 29 12:46:58 UTC 2015 - jreidinger@suse.com
+
+- set only proper boot boot flags, otherwise it can confuse some
```
lslezak: Duplicated `boot` ?

jreidinger: yep, thanks for catch

lslezak: The term `proper boot flags` is a bit vague, could you be more specific if possible?

mvidner: Exactly. I see that the bug report does not help much because you wrote it yourself :) but please try a user's perspective: if the old code failed, the circumstances and symptoms would be such and such. Probably include "Distinguish between MSDOS and GPT tables when setting boot flags" ?

mvidner:
The code looks good (but maybe I change my mind after you fix the description ;-) )

OTOH, Rubocop has no feelings: "Class definition is too long. [179/176]"

lslezak:
LGTM

PR #246, opened 2015-07-07
Mbr update s390

jreidinger:


test/mbr_update_test.rb
```diff
@@ -28,6 +31,16 @@
       )
     end
 
+    context "on s390" do
+      before do
+        allow(Yast::Arch).to receive(:architecture).and_return("s390_64")
+      end
+
+      it "does nothing except returning true" do
+        expect(subject.run).to eq true
```
ancorgs: You are checking that it returns true, but not that it's doing nothing.

jreidinger: well, it is hard to test that nothing happen, but in general without mocks it will fail if it try to do anything, especially without root permissions.

ancorgs: What about a
```
it "does not do not create any backup on s390"
```
inside
```
context "BootCommon.backup_mbr config is set"
```

And a
```
it "does not write generic mbr anywhere on s390"
```
inside
```
context "when generic mbr is enabled"
```
?

jreidinger: and also add it to when partition activated? this mean, that for any new feature I need to verify that it really do nothing, that is a bit overkill for me.

arvidjaar: > it "does not do not create any backup on s390"

Should it not be
```
it "does not create any backup on s390"
```
??

ancorgs: @arvidjaar Yes. Copy&paste error.

@jreidinger Yes, it's kind of annoying. But otherwise you are actually not checking that s390 is actually skipped. Maybe there is a better way, but the current one (just relying on the absence of mocks for the testsuite to crash) doesn't look correct to me.

mvidner: > this mean, that for any new feature I need to verify that it really do nothing, that is a bit overkill for me.

Yes, and that will remind you of the "# FIXME: make it single responsibility class" ;-)

Since the tests already do contain many cases of `expect(...).to_not receive(...)` I think it is fair to add such expectations for the s390 case. Whether they are bundled inside a single "architecture" context or dispersed among the "responsibility" contexts, I don't care much.

jreidinger: so you are not against using single architecture context? Otherwise I do not mind to change it.

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jul  7 12:20:21 UTC 2015 - jreidinger@suse.com
+
+- skip mbr update on s390 (bnc#937015)
```
mvidner: "MBR"

test/mbr_update_test.rb
```diff
@@ -28,6 +31,21 @@
       )
     end
 
+    context "on s390" do
+      before do
+        allow(Yast::Arch).to receive(:architecture).and_return("s390_64")
```
mvidner: Actually here we're depending on an implementation detail of the `Arch` class. Instead:
```ruby
allow(Yast::Arch).to receive(:s390).and_return(true)
```

and `and_return(false)` for the default case. Then the default case is depending on *our* implementation details which is better.

jreidinger: ok, it is fair. Above I use architecture, because I have to mock other cases like `Arch.ppc` in one step, so for generic mock it is needed, for this one, it is not needed, so I modify it.

test/mbr_update_test.rb
```diff
@@ -28,6 +31,21 @@
...
+        allow(Yast::Arch).to receive(:s390).and_return(true)
+      end
+
+      it "does nothing except returning true" do
+        expect(Yast::WFM).to_not receive(:Execute)
+        expect(Yast::SCR).to_not receive(:Execute)
+        expect_any_instance_of(::Bootloader::BootRecordBackup).to_not(
+          receive(:write)
+        )
+        expect(subject.run).to eq true
```
ancorgs: I'm still not 100% satisfied, since there are other non-excluded reasons for WFM and SCR to not be  called (like ```Yast::BootCommon.globals["generic_mbr"]``` being "false" or `Yast::BootCommon.backup_mbr` being false), but I can live with it. So let's move forward.

ancorgs:
As said, I consider the current version to be good a enough for a LGTM, once the comments from @mvidner has been addressed.

test/mbr_update_test.rb
```diff
@@ -20,6 +20,9 @@
         dev + num.to_s
       end
 
+      # by default common architecture"
+      allow(Yast::Arch).to receive(:architecture).and_return("x86_64")
```
mvidner: So this is staying because another method is querying `Arch.ppc`, and more methods do query `Arch.architecture` anyway. In the end I acknowledge that [`Arch.architecture`](http://www.rubydoc.info/github/yast/yast-yast2/Yast/ArchClass#architecture-instance_method) is public API after all, with boolean shortcuts. OK.

mvidner:
LGTM.

test/mbr_update_test.rb
jreidinger: yeah, it also helps with other modules that can be called here to have unified env.

jreidinger:
@mvidner @ancorgs thanks for review

PR #250, opened 2015-08-05
always run mkinitrd at the end of S/390 installation (bsc#933177)

jsrain:


jsrain:
Fix tested by Ihno

jsrain:
The fix is S/390-specific and requires specific set-up, therefore hard to be tested automatically.

kobliha:
Let's have this as a hotfix and then PLS reassign to @jreidinger  for re-evaluation. I think that Initrd.changed could be set in cio_ignore (considering the comment), but maybe there's a reason to make it $this way.

LGTM

PR #251, opened 2015-08-14
as SCR hasn't been setup yet, use some trickery to read boot config (…

wfeldt:
…bsc #940486)

The real problem is that SCR isn't pointing at /mnt yet. IMO is should, but that would be a huge change
and there are probably very good reasons against it.

So, in view of the fact that the perl-Bootloader interface will soon be gone anyway, we take a shortcut
here and just link the files into the inst-sys (the target is already mounted at /mnt).

After reading the config, adjust the product name and we're done.

### DIFF
--- a/src/lib/bootloader/proposal_client.rb
+++ b/src/lib/bootloader/proposal_client.rb
@@ -123,8 +123,8 @@ module Bootloader
           # SCR isn't pointing to /mnt yet but we'd really like to read
           # the config files - so we're cheating a bit.
           WFM.Execute(Path.new(".local.bash"),
-            "ln -s /mnt/boot/grub2 /boot; " +
-            "ln -s /mnt/etc/default/grub{,_installdevice} /etc/default; " +
+            "ln -s /mnt/boot/grub2 /boot; " \
+            "ln -s /mnt/etc/default/grub{,_installdevice} /etc/default; " \
             "ln -s /mnt/etc/sysconfig/bootloader /etc/sysconfig"
           )
           Bootloader.blRead(true, true)
--- a/src/lib/bootloader/proposal_client.rb
+++ b/src/lib/bootloader/proposal_client.rb
@@ -122,22 +122,22 @@ module Bootloader
         if !Yast::BootCommon.was_read || force_reset
           # SCR isn't pointing to /mnt yet but we'd really like to read
           # the config files - so we're cheating a bit.
-          WFM.Execute(Path.new(".local.bash"),
+          Yast::WFM.Execute(Path.new(".local.bash"),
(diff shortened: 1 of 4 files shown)


wfeldt:
Also, I've no idea if the Yast:: prefix is really necessary or not. Seems not...

ancorgs:
The Yast prefix should be needed, you are not inside the Yast namespace there.

src/lib/bootloader/proposal_client.rb
```diff
@@ -119,8 +120,19 @@ def propose_for_update(force_reset)
...
+            "ln -s /mnt/boot/grub2 /boot; " +
+            "ln -s /mnt/etc/default/grub{,_installdevice} /etc/default; " +
+            "ln -s /mnt/etc/sysconfig/bootloader /etc/sysconfig"
+          )
+          Bootloader.blRead(true, true)
+          BootCommon.was_read = true
+          # update the product name
+          prod = Product.short_name
+          BootCommon.globals["distributor"] = prod
+          log.info "grub2 menu entry = #{prod}"
```
ancorgs: Is ok to leave everything mounted at the end?

src/lib/bootloader/proposal_client.rb
```diff
@@ -119,8 +120,19 @@ def propose_for_update(force_reset)
       if ["grub2", "grub2-efi"].include? old_bootloader
         log.info "update of grub2, do not repropose"
         if !Yast::BootCommon.was_read || force_reset
-          Yast::Bootloader.blRead(true, true)
-          Yast::BootCommon.was_read = true
+          # SCR isn't pointing to /mnt yet but we'd really like to read
+          # the config files - so we're cheating a bit.
+          WFM.Execute(Path.new(".local.bash"),
+            "ln -s /mnt/boot/grub2 /boot; " +
```
ancorgs: As rubocop would say, use `\` instead of `+`

wfeldt:
But it works without; is that bad?

src/lib/bootloader/proposal_client.rb
wfeldt: yes

ancorgs:
In plain Ruby (using good old `require` and sane namespaces) the namespace is needed. If it's working is due to some magic of Yast.import. I would sleep better if the namespaces are used, like in good ruby. Maybe some day we will "fix" import :-)

ancorgs:
Moreover, it's quite ugly to mix prefixed and prefix-less calls in the same method.

wfeldt:
The existing code sometimes used a prefix, sometimes not. So I was wondering what to do...

lslezak:
LGTM

PR #252, opened 2015-08-17
Fix parted prompt

jreidinger:


kobliha:
Although I prefer --script instead of -s because it tells everyone more than just the short one, it's LGTM :)

PR #254, opened 2015-08-24
Multivalued kernel arguments

imobach:
This PR adds support to have multivalued kernel arguments, like: `crashkernel=72M,low crashkernel=384M,high`.

It needs a lot of discussion because, as it's implemented, it could break the API: [getKernelParamFromLine](https://github.com/yast/yast-bootloader/blob/53bc37a0979671c63f2266a842431cec72d5b7b3/src/include/bootloader/routines/misc.rb#L147) could return an string or an array (if param has multiple values). On the other side, [setKernelParamToLine](https://github.com/yast/yast-bootloader/blob/53bc37a0979671c63f2266a842431cec72d5b7b3/src/include/bootloader/routines/misc.rb#L170) could receive also an array of values.

Also [modify_kernel_param](https://github.com/yast/yast-bootloader/blob/86ddff831cf2083527bcb2bdf2d730e55d40f22e/src/modules/Bootloader.rb#L480) and [kernel_params](https://github.com/yast/yast-bootloader/blob/86ddff831cf2083527bcb2bdf2d730e55d40f22e/src/modules/Bootloader.rb#L438) are also modified to support multivalued kernel parameters.


src/include/bootloader/routines/misc.rb
```diff
@@ -139,35 +139,51 @@ def GetBootloaderDevices
...
     def getKernelParamFromLine(line, key)
       # FIXME: this doesn't work with quotes and spaces
       res = "false"
       # we can get nil if params is not yet proposed, so return not there (bnc#902397)
       return res unless line
       params = line.split(" ").reject(&:empty?)
-      params.each do |p|
-        l = p.split("=")
-        res = l[1] || "true" if l[0] == key
+      values = params.map { |p| kernel_param_parts(p) }.select { |p| p[0] == key }.map(&:last)
```
jreidinger: hmm, this is quite confusing, as it is quite hidden that kernel_param_parts remap to "true" if value is not there. So prefer more old code, where it is more obvious. Change also have one nasty side effect. For flags like "verbose" it do not make sense to report it multipletime so for line like `verbose crashkernel=bla,bla verbose 3` it should return only "verbose" => "true" not ["true", "true"] as it do not make sense.

src/include/bootloader/routines/misc.rb
```diff
@@ -182,25 +198,28 @@ def setKernelParamToLine(line, key, value)
...
         elsif occurences[k] == 1 # last parameter with given key
           done = true
-          if value == "true"
+          if values == "true"
             res << key
-          elsif value != "false"
-            res << Builtins.sformat("%1=%2", key, value)
+          elsif values != "false"
+            Array(values).each do |v|
+              res << Builtins.sformat("%1=%2", key, v)
```
jreidinger: and now you hit your problem. With `verbose crashkernel=bla,bla verbose` read and write, then result will be `verbose=true crashkernel=bla,bla verbose=true` which is wrong.

src/modules/Bootloader.rb
```diff
@@ -447,11 +449,14 @@ def kernel_param(flavor, key)
       raise ArgumentError, "Unknown flavor #{flavor}" unless kernel_line_key
 
       line = BootCommon.globals[kernel_line_key]
-      ret = BootCommon.getKernelParamFromLine(line, key)
+      values = BootCommon.getKernelParamFromLine(line, key)
 
       # map old api response to new one
-      api_mapping = { "true" => :present, "false" => :missing }
-      api_mapping[ret] || ret
+      if values.kind_of?(Array) # more than one value
```
jreidinger: I prefer to use `is_a?` as it allows to also handle smarted arrays

jreidinger: ah, it is alias...hmm, rubocop do not unify it?

src/modules/Bootloader.rb
```diff
@@ -447,11 +449,14 @@ def kernel_param(flavor, key)
...
+        values
```
jreidinger: even this have problem with multi "true" value

src/include/bootloader/routines/misc.rb
imobach: For me it makes sense to have an small method that knows how to split a kernel parameter between key and value. But if `getKernelParamFromLine` is the only place to use it, it's fine to move that logic back.

I didn't though about a duplicated parameter, but a call to `uniq` will fix the issue.

src/modules/Bootloader.rb
imobach: I need to fix some rubocop issues and, yes, `is_a?` is one of them :)

test/bootloader_test.rb
```diff
@@ -150,4 +158,177 @@
...
+        expect(Yast::BootCommon.changed).to eq(false)
+      end
+    end
+
+    context "when no target is specified" do
+      it "uses :common by default" do
+        subject.modify_kernel_params(params)
+
+        expect(Yast::BootCommon.globals)
+          .to eq(kernel_line(:common) => append, "__modified" => "1")
```
jreidinger: this is quite fragile, as it can change what is in start in globals, better check directly kernel like key like 
```
expect(Yast::BootCommon.globals[kernel_line(:common)]).to eq append
```

test/bootloader_test.rb
```diff
@@ -150,4 +158,177 @@
...
+
+      it "adds the parameter as 'vgamode' to the global scope and does not mark BootCommon as 'modified'" do
+        subject.modify_kernel_params(:common, params)
+        expect(Yast::BootCommon.globals)
+          .to eq("vgamode" => "80")
+        expect(Yast::BootCommon.changed).to eq(false)
+      end
+    end
+
+    context "when parameter is 'root' (unsupported)" do
```
jreidinger: root parameter is supported, just grub2 do not allow modification as it is generated on its own. So changing it have no effect.

imobach: Ok, I'll choose a better wording for that.

jreidinger:
LGTM

imobach:
@jreidinger Thanks. I'll wait to merge until @ancorgs and me integrate changes from yast-kdump to drive some tests.

BTW, should I update in any way publish information about changed functions? I mean [this](https://github.com/yast/yast-bootloader/blob/53bc37a0979671c63f2266a842431cec72d5b7b3/src/modules/BootCommon.rb#L558).

jreidinger:
yes, I am not sure if there is any perl user, but to be on safe side, we should change it to e.g. `any(string, string)` for get...so replace all string which can return now array with `any`

imobach:
@jreidinger Yes, it sounds reasonable. And does exist any way to specify that two different types are supported? I'm thinking about `setKernelToLine` which could receive an string or an array as third parameter. I've been searching for some examples in our code a I didn't see anything like that.

jreidinger:
@imobach no, simply use `any` which stands for more possible types.

imobach:
@jreidinger Ok, done. Thanks again.

imobach:
I've rebased the commits, bumped the version and updated the changes file.

imobach:
We've make some tests and it's working.

jreidinger:
LGTM

PR #255, opened 2015-08-24
Fix removing pwd

jreidinger:


ancorgs:
LGTM. Certainly I'd have never expected a constant called YAST_BASH_PATH to be local instead of target and `bash_output` instead of `bash`.

wfeldt:
ok to me, too

jreidinger:
@ancorgs I agree that it is not nice. It is caused by not so nice SCR API. For me much better will be something like `Script.run "bla bla bla", local: true` for local and by default use target.

PR #258, opened 2015-09-04
Bsc 943749

wfeldt:


src/modules/BootGRUB2.rb
```diff
@@ -152,6 +152,9 @@ def Propose
         partition = prep_partitions.first
         if partition
           BootCommon.globals["boot_custom"] = partition
+          BootCommon.globals["boot_boot"] = "false"
+          BootCommon.globals["boot_root"] = "false"
+          BootCommon.globals["boot_mbr"] = "false"
```
jreidinger: it should be by default false. Do you have idea what propose it to true? It should not happen

wfeldt:
autoyast import into y2bl is imo not correct: if you use an empty autoyast file, you get a different proposal than with a default install.

E.g. on sle12-ga, with autoyast grub2 is installed into mbr, without autoyast it's installed into partition.

With sp1 the situation has improved: it proposes grub2 to install into a partition. But then it does it all _again_ and this time it goes into mbr. This is because the proposing fails and mbr is the default fallback. The first patch fixes this (the device map was missing).

The second patch is really for ppc and ensures we only install into the custom partition (prep). Else it will try root/boot as well and fails.

It might be correct to always skip the alternative location when boot_custom is active. But I'm not sure. I first thought about enforcing this in pbl but then this will go away soon, so there's no point.

src/modules/BootGRUB2.rb
wfeldt: it proposes boot_root and looks quite logical... :-)

wfeldt: stage1:::propose_boot_location can only come up with  mbr, boot or root or not?

jreidinger: thats right. I am considering to change it, so stage1 propose boot location can propose custom on ppc. I already have similar bug report and it looks like better solution for me, as then you can replace proposal in BootGRUB2 with simple stage1 propose.

wfeldt: ok, then go ahead

btw, my test completed and everything worked but the prep partition wasn't tagged as active; so it's still not correct...

wfeldt: There's still generic_mbr...

jreidinger: it is also handled in stage1 proposal...so for ppc I suggest to use custom prep partition, activate flag and do not set generic_mbr there. It at least unify behavior and similar stuff for s390. Reason why it is separated now, is that stage1 originally in SLE12 GA is used only on x86 architecture, so I refactoring it also this way, but it is more correct to do it properly.

jreidinger:
supperseeded by https://github.com/yast/yast-bootloader/pull/259

PR #259, opened 2015-09-07
Bsc 943749 2

jreidinger:


wfeldt:
A test on ppc64le worked. So, looks fine to me.

andreas-schwab:
Details: uninitialized constant Bootloader::Stage1::Arch
Caller:  /usr/share/YaST2/lib/bootloader/stage1.rb:36:in `propose'

jreidinger:
@andreas-schwab I get where is wrong, but it is also exception raising. I expect we are talking about arm, right? I am not sure which proposal is used in past for arch? same as for x86? or do we use something special when proposing stage1 location?

andreas-schwab:
There was no architecture dependent proposal before.  But aarch64 is always using EFI.

jreidinger:
@andreas-schwab it is very strange, as EFI do not use stage1 at all. Can you please attach logs to bug report? It is not clear for me how non-EFI stage1 proposal can appear on EFI only architecture.

PR #262, opened 2015-09-11
Workaround for bnc#945479 in SLE12

ancorgs:
We should probably fix perl-bootloader, but this is waaay less risky.

I tested manually, but I failed to generate a sane unit test involving perl-bootloader. A full integration test would be needed... too much. :(


lslezak:
LGTM

PR #264, opened 2015-09-15
Fix aarch64

jreidinger:


wfeldt:
looks ok

src/include/bootloader/routines/lilolike.rb
```diff
@@ -67,7 +67,8 @@ def FindMBRDisk
     # FIXME: replace with grub_ConfigureLocation() when lilo et al. have
     # changed to stop using selected_location and loader_device.
     def ConfigureLocation
-      ::Bootloader::Stage1.new.propose.to_s
+      # configure stage1 only on non-EFI systems
```
kobliha: Also PLS reference bug Nr if possible.

jreidinger: well, it is general statement as EFI do not have stage1. So bug number do not help here, as bug do not mention this problem at all. In general stage1 is only when bootloader need to have specific boot code somewhere, which is not EFI case which works differently

PR #265, opened 2015-09-16
Fix raid

jreidinger:


src/lib/bootloader/udev_mapping.rb
```diff
@@ -37,9 +37,20 @@ def to_hash
     # @param dev [String] device udev or kernel one like /dev/disks/by-id/blabla
```
aschnell: it's "/dev/disk/..."

jreidinger: thanks, will fix it...code below is correct

test/udev_mapping_test.rb
```diff
@@ -17,7 +18,27 @@
...
 
-    it "return argument for non-udev mapped device names" do
+    it "returns mapped raid name for partitioned devices" do
+      allow(Yast::Storage).to receive(:GetContVolInfo) do |_dev, info|
+        info.value["vdevice"] = "/dev/md126p1"
+        info.value["cdevice"] = ""
+        true
+      end
+
+      expect(subject.to_kernel_device("/dev/md/crazy_name")).to eq "/dev/md126p1"
```
aschnell: This test looks strange to me since the return value of GetContVolInfo is independent of its input.

jreidinger: OK, I add verification of input to call

src/lib/bootloader/udev_mapping.rb
```diff
@@ -37,9 +37,20 @@ def to_hash
...
     # @raise when device have udev format but do not exists
     def to_kernel_device(dev)
```
aschnell: The documentation of the function should be updated since /dev/md/r0d2n0s128 is not a udev name (I assume with udev names you mean anything in /dev/disk/by-...

jreidinger: what is proper googlable term that can be used for that kind of device names?

aschnell:
Overall the file looks confusing to me. E.g. having a function all_devices and a variable @all_devices looks bad to me. It's also not documented what all_devices is and contains.

jreidinger:
@aschnell OK, I improve documentation for all_devices method. In general having @all_devices as variable and all_devices reader is common practice. I do not use `attr_reader` because it I want lazy loading and also cache validation here, so I have own defined reader.

src/lib/bootloader/udev_mapping.rb
```diff
@@ -34,17 +34,28 @@ def to_hash
...
+    # @param dev [String] device udev or kernel one like /dev/disk/by-id/blabla
     # @raise when device have udev format but do not exists
     def to_kernel_device(dev)
-      return dev if dev !~ /^\/dev\/disk\/by-/
-
-      all_devices[dev] or raise "Unknown udev device #{dev}"
+      # for non-udev devices try to see specific raid names (bnc#944041)
+      if dev =~ /^\/dev\/disk\/by-/
+        all_devices[dev] or raise "Unknown udev device #{dev}"
+      else
```
aschnell: The special case for devices starting with /dev/disk/by-... should not be needed since GetContVolInfo should also find those. But surely that is not tested.

jreidinger: idea of whole this class is to have quick cache for udev mapped devices. It is relict from past that speed up this translation for system with incredible huge amount of disks. So any call outside just slow down its performance.

aschnell:
LGTM

jreidinger:
@aschnell thanks

PR #266, opened 2015-09-24
Activate fix

jreidinger:


src/modules/BootSupportCheck.rb
```diff
@@ -245,6 +241,21 @@ def check_zipl_part
...
+      return true if BootCommon.globals["activate"] == "true" || Yast::Storage.GetBootPartition(Yast::BootCommon.mbrDisk)
+
+      AddNewProblem(_("Activate flag is not set by installer. If it is not set at all, some BIOSes could refuse to boot."))
+      false
+    end
+
+    def check_mbr
+      return true if BootCommon.globals["generic_mbr"] == "true" || BootCommon.globals["boot_mbr"] == "true"
+
+      AddNewProblem(_("Installer do not modify disk MBR. Unless it contain already some boot code it, disk refuse to boot."))
```
ancorgs: Very hard to read and get the meaning. I guess you meant something like "The installer will not modify the MBR of the disk. Unless it already contains some boot code, the disk will not be able to boot".

src/modules/BootSupportCheck.rb
```diff
@@ -256,8 +267,10 @@ def GRUB
     def GRUB2
       ret = GRUB()
       # ensure that s390 have ext* partition for booting (bnc#873951)
-      ret &&= check_zipl_part if Arch.s390
-      ret &&= check_gpt_reserved_partition if Arch.x86_64
+      ret = check_zipl_part && ret if Arch.s390
```
ancorgs: Precedence can be very tricky here. I would use a more explicit format so it's understandable at a first sight for average Joe.

ancorgs:
LGTM

mvidner:
@jreidinger It is likely this change that makes http://river.suse.de/job/autoyast-integration-test fail. I have just seen it in a local run of the test

PR #267, opened 2015-09-25
Fix ppc raid

jreidinger:


k0da:
That fixes issue on PowerPC, but what about Z ?

jreidinger:
@k0da Z do not use boot flags and whole concept on Z is different. see https://github.com/yast/yast-bootloader/blob/fix_ppc_raid/src/lib/bootloader/mbr_update.rb#L23

ancorgs:
LGTM

PR #269, opened 2015-09-28
fix device map handling if there's no 'hd0' entry in it (bsc#947730)

wfeldt:


shundhammer:
LGTM

aschnell:
Changelog entry with bug number is missing.

jreidinger:
added changelog entry and merged

PR #270, opened 2015-10-01
Fix autoyast install

jreidinger:


src/include/bootloader/routines/lilolike.rb
```diff
@@ -29,7 +29,6 @@ def initialize_bootloader_routines_lilolike(include_target)
       Yast.import "Storage"
 
       Yast.include include_target, "bootloader/routines/i386.rb"
-      Yast.include include_target, "bootloader/grub2/misc.rb"
```
schubi2: Now I have understood how it should look like :-)

schubi2:
LGTM

jreidinger:
thanks

PR #272, opened 2015-10-16
Fixed validation of AutoYaST profile (bsc#805275).

mvidner:
Note that this is a quick and dirty way that makes one particular profile pass.
The actual element names are defined in [perl-Bootloader](https://github.com/openSUSE/perl-bootloader/blob/85a38144be290430c9e7885f2c7a592cbb787f5c/src/Core/GRUB2.pm#L766)


jreidinger:
Do we really return all of it? I think some elements we do not allow to configure at all in yast2-bootloader.

imobach:
@mvidner @jreidinger Here you are and example of `yast2 clone_system` on SLE 12 (SP0). It looks like we export a lot of elements.

```xml
  <bootloader>
    <global>
      <append> resume=/dev/disk/by-uuid/3963f96a-449a-43e8-bbef-e32ccdcb9226 splash=silent quiet showopts crashkernel=64M-:32M</append>
      <append_failsafe>showopts apm=off noresume edd=off powersaved=off nohz=off highres=off processor.max_cstate=1 nomodeset x11failsafe crashkernel=64M-:32M</append_failsafe>
      <cryptodisk config:type="integer">0</cryptodisk>
      <default>SLES12</default>
      <distributor>SLES12</distributor>
      <gfxbackground>/boot/grub2/themes/SLE/background.png</gfxbackground>
      <gfxmode>auto</gfxmode>
      <gfxtheme>/boot/grub2/themes/SLE/theme.txt</gfxtheme>
      <hiddenmenu>false</hiddenmenu>
      <os_prober>false</os_prober>
      <suse_btrfs>true</suse_btrfs>
      <terminal>gfxterm</terminal>
      <timeout config:type="integer">8</timeout>
      <xen_append>crashkernel=64M-:32M</xen_append>
    </global>
    <loader_type>grub2</loader_type>
    <sections config:type="list">
      <section>
        <append>${extra_cmdline}  resume=/dev/disk/by-uuid/3963f96a-449a-43e8-bbef-e32ccdcb9226 splash=silent quiet showopts crashkernel=64M-:32M</append>
        <image>/boot/vmlinuz-3.12.48-52.27-default</image>
        <menuentry>SLES12</menuentry>
        <name>SLES12</name>
        <root>UUID=7586b5a8-5c00-4338-aeb0-2157bc327914</root>
        <type>image</type>
        <usage>linux</usage>
      </section>
    </sections>
  </bootloader>
```

imobach:
And it looks like are the elements added by @mvidner in the PR are present.

jreidinger:
@imobach hmm, exporting sections are wrong. It is ignored as grub2 generate it itself. And few others are also generated without user, so the are useless.( like suse_btrfs, gfxtheme, cryptodisk )

imobach:
@jreidinger I see, so it's not only fixing the schema. On the other hand, AFAIK `gfxtheme` is also shown in the yast2-bootloader UI.

jreidinger:
OK, so in general I think we can allow such elements, but somewhere note, that it do nothing. In future I plan to fix it to not export it at all as it now export all options read by perl-bootloader, but not all of it is allowed to change nor having any effect.

jreidinger:
LGTM

imobach:
Thanks! Finally merging...

PR #273, opened 2015-10-19
Skip grub2 update

jreidinger:
new translatable string is not problem as it is already used at different location.

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Oct 19 16:10:34 UTC 2015 - jreidinger@suse.com
+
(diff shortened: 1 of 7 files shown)


wfeldt:
Would it make sense to also check for boot_common in
https://github.com/yast/yast-bootloader/blob/master/src/modules/BootGRUB2.rb#L207
?

wfeldt:
I meant boot_custom

jreidinger:
@wfeldt yes, it make sense to fix also this, will add it.

jreidinger:
@wfeldt done

wfeldt:
logic ok with be, but you'll need to make rubocup happy

wfeldt:
ok now

PR #275, opened 2015-11-04
Fix AutoYaST schema

imobach:
Related to [bsc#805275](https://bugzilla.suse.com/show_bug.cgi?id=805275).


src/config/bootloader.rnc
```diff
@@ -54,19 +54,25 @@ bl_global =
...
 gfxmenu = element gfxmenu { text }
+boot_efilabel = element boot_efilabel { text }
 
 # really "true"|"false", without config:type=boolean
 bl_activate = element activate { "true" | "false" }
 boot_root = element boot_root { "true" | "false" }
 boot_boot = element boot_boot { "true" | "false" }
 boot_mbr = element boot_mbr { "true" | "false" }
+prompt = element prompt { "true" | "false" }
+secure_boot = element secure-boot { "true" | "label" }
```
mvidner: Really true/**label**? That sounds really suspicious and should get a comment if it is indeed correct.

mvidner: Also, it is the only element with a `-` dash instead of an `_` underscore.

imobach: Oh, I forgot to submit the latest version (I detected that pitfall when I tried it).

imobach: Yes, it looks suspicious, but I found a reference here: https://github.com/yast/yast-bootloader/blob/Code-11-SP1/src/elilo/helps.ycp#L57

mvidner:
FYI, when I started to fix  [bsc#805275](https://bugzilla.suse.com/show_bug.cgi?id=805275) in #272, I targeted SLE-12-GA, because that's what the L3 bug was reported against. There the backend is different and the schema is different too.

imobach:
As a general issue, must I submit this change against a `Code-11-SP3`? I've used `Code-11-SP1` because the SP3 branch does not exist and  it's broken also for SLE11 SP1.

imobach:
@mvidner Yes, I've only focused in SLE11 (I've seen that you have done some fixes to SLE 12).

mvidner:
According to https://www.suse.com/lifecycle/ , SLE-11-SP3 is supported until 2016-01-31
But it seems @jreidinger has used the SP1 branch for SP4-only  fixes already? Josef, can you clarify?

jreidinger:
@mvidner yes, this branch is used for SP4 also...it is cuased due to maintenace policy change during SLE11 lifecycle, when it should be unified and when not, it is not changed, so in the end all happen in this one branch.

mvidner:
LGTM

PR #278, opened 2015-11-17
Fixed crash in bootloader proposal if previous installation was on software RAID (bsc#955216)

shundhammer:
Make sure to return a value from the loop body even if prematurely continuing with the next loop iteration; otherwise the accumulator variable ("res") will be nil for the next iteration, causing a crash.


shundhammer:
Closed to avoid duplication - will merge the SP1 branch instead

PR #280, opened 2015-11-18
Drop usage of perl-Bootloader

jreidinger:
proof of concept. Not yet with required level of test coverage and documentation. Mainly to get some quick feedback


src/lib/bootloader/grub2base.rb
```diff
@@ -110,6 +109,10 @@ def Propose
         end
       end
 
+      propose_os_probing
+      propose_terminal
+      propose_timout
```
imobach: Just a typo here (`propose_timeout`).

src/lib/bootloader/serial_console.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def self.load_from_kernel_args(kernel_params)
+      console_parameter = kernel_params.parameter("console")
+      return nil unless console_parameter
+
+      console_parameter = [console_parameter] unless console_parameter.is_a? ::Array
+      serial_console = console_parameter.find { |p| p =~ /ttyS/ || p =~ /ttyAMA/ }
+      return nil unless serial_console
+
+      console_regexp = /(ttyS|ttyAMA)[[:alpha:]]+([[:digit:]]*),?([[:digit:]]*)([noe]*)([[:digit:]]*)/
```
imobach: I'd use a constant here.

src/lib/bootloader/serial_console.rb
```diff
@@ -0,0 +1,70 @@
...
+    }
+    SPEED_DEFAULT = 9600
+    PARITY_DEFAULT = "no"
+    WORD_DEFAULT = ""
+
+    def self.load_from_kernel_args(kernel_params)
+      console_parameter = kernel_params.parameter("console")
+      return nil unless console_parameter
+
+      console_parameter = [console_parameter] unless console_parameter.is_a? ::Array
```
imobach: You could use `Array()` here.

src/lib/bootloader/serial_console.rb
kobliha: And that constant should have

1. Description - what is is good for
2. Example - what it matches :)

src/lib/bootloader/serial_console.rb
```diff
@@ -0,0 +1,70 @@
...
+      speed = SPEED_DEFAULT if speed.empty?
+      parity = serial_console[console_regexp, 4]
+      parity = PARITY_DEFAULT if parity.empty?
+      parity = PARITY_MAP[parity]
+      word = serial_console[console_regexp, 5]
+
+      new(unit, speed, parity, word)
+    end
+
+    def self.load_from_console_args(console_args)
```
imobach: Ok, I know that documentation is missing yet, but some explanation about `console_args` will help :)

imobach:
In general, looks good. Anyway, I miss some documentation to make this pre-review process easier :)

src/lib/bootloader/grub2base.rb
jreidinger: thanks for catch. I really indicate need of tests :)

src/lib/bootloader/serial_console.rb
jreidinger: well, it is obvious from code below, where I use names for matched pairs, for there is never enough good examples :)

src/lib/bootloader/serial_console.rb
jreidinger: I know about it, but some people find it tricky and with worse readability. If only experience rubist read code, I can use it :)

src/lib/bootloader/serial_console.rb
jreidinger: it is actually exactly arguments for serial console that grub needs to set it up. Reason why it need to be twice is to setup:
1) serial console for grub
2) serial console for kernel in kernel args

src/lib/bootloader/serial_console.rb
imobach: I tend to use it but I understand that it could be tricky. So I'm happy as it is anyway :)

jreidinger:
travis failed due to not yet published config_files gem.

ancorgs:
Just one note about a removed comment. Other than that, LGTM.

src/lib/bootloader/grub2.rb
```diff
@@ -0,0 +1,174 @@
...
+
+Yast.import "Arch"
+Yast.import "BootStorage"
+Yast.import "Storage"
+Yast.import "BootCommon"
+Yast.import "HTML"
+
+module Bootloader
+  # Represents non-EFI variant of GRUB2
+  class Grub2 < GRUB2Base
```
mvidner: This is inconsistent. We should use only one of `Grub`, `GRUB` in class names.

jreidinger: I know, currently it is WIP. I plan to rename GRUB2Base to Grub2Base

src/lib/bootloader/grub2.rb
```diff
@@ -0,0 +1,174 @@
...
+    # @return [Boolean] true on success
+    def write
+      # TODO: device map write
+      # TODO: install_dev write
+
+      # something with PMBR needed
+      # TODO: own class handling PBMR
+      if BootCommon.pmbr_action
+        boot_devices = BootCommon.GetBootloaderDevices
+        boot_discs = boot_devices.map { |d| Storage.GetDisk(Storage.GetTargetMap, d) }
```
mvidner: Inconsistent spelling **disk** vs **disc**. Use **disk**: https://en.wiktionary.org/wiki/disk#Usage_notes

src/lib/bootloader/none_bootloader.rb
```diff
@@ -0,0 +1,19 @@
+require "yast"
+
+Yast.import "HTML"
+
+module Bootloader
+  # Represents when bootloader want user manage itself
```
mvidner: > The user wants to manage the bootloader themself.

src/lib/bootloader/none_bootloader.rb
```diff
@@ -0,0 +1,19 @@
...
+  class NoneBootloader < BootloaderBase
+    def summary
```
mvidner: This should be declared in BootloaderBase together with a note about the non-obvious fact  that the return type is an array.

jreidinger: the best way would be abstract method, as it do not make sense in base as it would be empty. But I can add it there.

src/lib/bootloader/grub2base.rb
```diff
@@ -1,51 +1,59 @@
...
+module Bootloader
   # Common base for GRUB2 specialized classes
-  class GRUB2Base < Module
+  class GRUB2Base < BootloaderBase
+    include Yast::Logger
+    include Yast::I18n
+
     # @!attribute password
     #    @return [::Bootloader::GRUB2Pwd] stored password configuration object
     attr_reader :password
```
mvidner: For whatever reason my yardoc-0.8.7.6 (and also [rubydoc.org](http://www.rubydoc.info/github/yast/yast-bootloader/master/Yast/GRUB2Base#password-instance_method)) ignores markup like this. I got it to render by removing the `@!attribute` line and outdenting the `@return` line.

jreidinger: will check it. I have probably different version

src/lib/bootloader/grub2base.rb
```diff
@@ -1,51 +1,59 @@
...
+Yast.import "Product"
+Yast.import "ProductFeatures"
+Yast.import "Stage"
+Yast.import "Storage"
+Yast.import "StorageDevices"
+
+module Bootloader
   # Common base for GRUB2 specialized classes
-  class GRUB2Base < Module
+  class GRUB2Base < BootloaderBase
```
mvidner: Unify GRUB and Grub. Perhaps add a `GRUB2Base = Grub2Base` with a `# @deprecated` tag?

src/lib/bootloader/grub_install.rb
```diff
@@ -0,0 +1,66 @@
...
+  # Wraps grub install script for easier usage.
+  class GrubInstall
+    def initialize(efi: false, secure_boot: false)
+      @efi = efi
+      @secure_boot = secure_boot
+
+      raise "cannot have secure boot without efi" if secure_boot && !efi
+    end
+
+    def execute(devices: nil)
```
mvidner: > \# @param [Array\<String>,nil] devices identified by eg. "/dev/hda7", "/dev/ttyS0", "(hd8)", "/dev/by-zodiac/sagittarius" ?

src/lib/bootloader/sections.rb
```diff
@@ -0,0 +1,39 @@
+require "yast"
+require "yast2/execute"
+
+module Bootloader
+  # Represents available sections and handling of default boot entry
+  class Sections
```
mvidner: Funny, I would expect an array of Section, not a Sections class.

src/lib/bootloader/sections.rb
```diff
@@ -0,0 +1,39 @@
...
+    # @param [CFA::Grub2::GrubCfg] grub_cfg - loaded parsed grub cfg tree
+    def initialize(grub_cfg)
+      @data = grub_cfg.sections
+    end
+
+    # Gets all available sections
+    def all
```
mvidner: I guess it is an Array. Or a Hash? Of String? or of what?

src/lib/bootloader/sections.rb
```diff
@@ -0,0 +1,39 @@
+require "yast"
+require "yast2/execute"
+
+module Bootloader
+  # Represents available sections and handling of default boot entry
+  class Sections
+    # @param [CFA::Grub2::GrubCfg] grub_cfg - loaded parsed grub cfg tree
+    def initialize(grub_cfg)
+      @data = grub_cfg.sections
```
mvidner: I think it is good style to declare `@default = nil` here. `ruby -w` will warn otherwise.

src/lib/bootloader/serial_console.rb
```diff
@@ -0,0 +1,101 @@
+require "yast"
+
+Yast.import "Arch"
+
+module Bootloader
+  # Represents parameters for console. Its main intention is easy parsing serial
+  # console parameters parameters for grub or kernel and generate it to keep it
```
mvidner: 2 "parameters"

src/lib/bootloader/serial_console.rb
```diff
@@ -0,0 +1,101 @@
...
+  # in sync.
+  class SerialConsole
+    PARITY_MAP = {
```
mvidner: > \# This class uses the values of this map, which are also used by Grub. The kernel uses the keys.

src/lib/bootloader/serial_console.rb
```diff
@@ -0,0 +1,101 @@
...
+    #
+    # For details see https://en.wikipedia.org/wiki/Serial_port
+    # @example serial console param ( on kernel cmdline "console=<example>" )
+    #    "ttyS0,9600n8"
+    # @example also partial specification works
+    #    "ttyAMA1"
+    KERNEL_PARAM_REGEXP = /(ttyS|ttyAMA)([[:digit:]]*),?([[:digit:]]*)([noe]*)([[:digit:]]*)/
+
+    # Loads serial console configuration from parameters passed to kernel
+    # @param [ConfigFiles::Grub2::Default::KernelParams] kernel_params to read
```
mvidner: I think it is a needless dependency on the KernelParams class. The other method simply gets a string, so I would expect this one to get a string too.

src/lib/bootloader/serial_console.rb
```diff
@@ -0,0 +1,101 @@
...
+  # Represents parameters for console. Its main intention is easy parsing serial
+  # console parameters parameters for grub or kernel and generate it to keep it
+  # in sync.
+  class SerialConsole
+    PARITY_MAP = {
+      "n" => "no",
+      "o" => "odd",
+      "e" => "even"
+    }
+    SPEED_DEFAULT = 9600
```
mvidner: For consistency: SPEED_DEFAULT = "9600", **or** make WORD_DEFAULT a \[Integer, nil]

src/lib/bootloader/grub2base.rb
```diff
@@ -1,51 +1,59 @@
...
-      Yast.import "Stage"
-      Yast.import "Storage"
-      Yast.import "StorageDevices"
-
-      # includes
-      # for simplified widgets than other
-      Yast.include self, "bootloader/grub2/dialogs.rb"
-
       @password = ::Bootloader::GRUB2Pwd.new
+      @grub_default = ::ConfigFiles::Grub2::Default.new
```
mvidner: It seems that ConfigFiles was renamed to CFA (in <https://github.com/config-files-api>), right?

jreidinger: yes, not updated yet. will do in near future, at first I would like to change design, then continue with automatic and manual tests that will discover it

src/lib/bootloader/grub2base.rb
```diff
@@ -53,110 +61,108 @@ def pmbr_setup(action, *devices)
...
+      @sections = ::Bootloader::Sections.new(grub_cfg)
     end
 
-    # Update read settings to new version of configuration files
-    def Update
-      Read(true, true)
+    def write
+      super
+      grub_default.save
+      # TODO: call grub_install
```
mvidner: But master branch does not call it at all, why?

jreidinger: because perl-Bootloader call it. Goal of this branch is to remove pbl dependency.

src/include/bootloader/routines/common_helps.rb
```diff
@@ -25,7 +25,7 @@ def initialize_bootloader_routines_common_helps(_include_target)
         )
       }
 
-      @common_descriptions = { "timeout" => _("&Timeout in Seconds") }
+      @common_descriptions = { "timeout" =>  }
```
mvidner: Nice try, but Travis never sleeps :)

jreidinger: thanks for catch, I actually plan to remove whole file. I just removing part by part, to not forget anything

jreidinger:
just for interest codeclimate comparison shows significant improvement https://codeclimate.com/github/yast/yast-bootloader/compare/config_files_proto

jreidinger:
JFYI final code climate rating change 3.3 GPA(+1.57) https://codeclimate.com/github/yast/yast-bootloader/compare/config_files_proto

Rakefile
```diff
@@ -4,4 +4,7 @@ Yast::Tasks.configuration do |conf|
   # lets ignore license check for now
   conf.skip_license_check << /.*/
   conf.install_locations["doc/autodocs"] = conf.install_doc_dir
+
+  conf.obs_project = "home:cwh:branches:YaST:Head"
```
lslezak: NTH: This is just a temporary OBS project isn't it? Then if would be nice to have a `FIXME` comment there...

jreidinger: ah, good catch, it should be removed from master, thanks for noticing it

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Apr 12 13:29:54 UTC 2016 - jreidinger@suse.com
+
+- do not propose nor running grub2-install on bare metal ppc
```
lslezak: NP: Inconsistencies found:

- capitalization, the item below starts with `Do not ...` and it should be `nor run`
- the changes below use `POWER`, not `ppc` (I'm not sure which one should be preferred)

lslezak:
LGTM (after a quick scan of the big and incomplete diff)

jreidinger:
thanks

PR #283, opened 2015-11-20
Fix AutoYaST schema in SLE 12 SP1

imobach:
Fixes [bsc#954412](https://bugzilla.suse.com/show_bug.cgi?id=954412) for SLE 12 SP1. BTW, `SLE-12-GA` branch is merged into `sle-12-sp1-after_release`.


jreidinger:
In general I am a bit confused by name as it is nit fix for SP1, but for TW and SP2. Is target correct?

imobach:
@jreidinger Yes, it's for every target starting at SP1. I guess I should also submit a PR for master. Isn't it?

imobach:
Thinking it twice, for `master` maybe it's better to do the PR when `sle-12-sp1-after_release` branch is merged in `SLE-12-SP1`.

jreidinger:
@imobach well, not necessary, we just need to make sure master include everything in SP1. Usual workflow is fix in the oldest branch and then merging it up, so all fixes is included in newer ones.

imobach:
@jreidinger Ok. So from my POV I should keep creating PRs with those fixes against `sle-12-sp1-after_release`. After SP1 is release, we'll merge those `after_release` branches into `SLE-12-SP1` and, from there, we'll merge them to `master`.

Do you agree?

jreidinger:
@imobach yes, fine for me. I just write goal of this rule. There are many ways how to achieve it :)

imobach:
@jreidinger Ok, thanks for clarification. Reviewing the conversation, I've realized that you are right and the target was wrong (I must be aslept). In other PR I've done for this matter, the target is `sle-12-sp1-after_release` (https://github.com/yast/yast-autoinstallation/pull/165) :)

PR #288, opened 2016-01-19
Add nounzip to schema

imobach:
It fixes [bsc#962620](https://bugzilla.suse.com/show_bug.cgi?id=962620). The `nounzip` element is introduced by perl-Bootloader.


package/yast2-bootloader.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jan 19 17:12:10 UTC 2016 - igonzalezsosa@suse.com
+
+- Add nounzip elemento to AutoYaST schema (bsc#962620)
```
kobliha: `elemento` did not validate :)

src/config/bootloader.rnc
```diff
@@ -104,7 +104,8 @@ section =
     element remap { "true" | "false" }? &
     element configfile {text}? &
     element image   {text}? &
-    element description {text}?
+    element description {text}? &
+    element nounzip {text}?
```
kobliha: Was this actually already dropped from master? I could not find it: https://github.com/search?q=nounzip+user%3Ayast&type=Code

imobach: I think we should port this change to master according to https://github.com/openSUSE/perl-bootloader/search?utf8=%E2%9C%93&q=nounzip

kobliha:
LGTM

imobach:
Thanks!

src/config/bootloader.rnc
jreidinger: no, as it is grub1 only parameter, grub2 do not allow to set it

imobach: @jreidinger Thanks for clarification.

PR #289, opened 2016-02-25
Adjust default kernel flags for aarch64

dirkmueller:
Should be more x86_64 like (quiet is needed to pass
openQA tests)


jreidinger:
LGTM

PR #291, opened 2016-03-02
Add support to install missing packages on running system

jreidinger:


test/bootloader_base_test.rb
```diff
@@ -14,6 +19,14 @@
 
       subject.write
     end
+
+    context "Mode.normal is set" do
+      it "install packages listed required by bootloader from packages method" do
```
lslezak: Hardly understandable to me, probably simple `installs packages required by bootloader` would be enough.

jreidinger: ok, make sense

lslezak:
LGTM

jreidinger:
thanks

PR #292, opened 2016-03-03
Jreidinger config files proto

jreidinger:
decrease of test coverage is expected as tests are filtered out now

### DIFF
--- a/test/grub2_test.rb
+++ b/test/grub2_test.rb
@@ -1,0 +1,159 @@
+require_relative "test_helper"
+
+require "bootloader/grub2"
+
+describe Bootloader::Grub2 do
+  before do
+    allow(::CFA::Grub2::Default).to receive(:new).and_return(double("GrubDefault").as_null_object)
+    allow(::CFA::Grub2::GrubCfg).to receive(:new).and_return(double("GrubCfg").as_null_object)
+    allow(Bootloader::Sections).to receive(:new).and_return(double("Sections").as_null_object)
+    allow(Yast::BootStorage).to receive(:available_swap_partitions).and_return([])
+    allow(Yast::BootStorage).to receive(:gpt_boot_disk?).and_return(false)
+  end
+
+  describe "#read" do
+    before do
+      allow(Yast::BootStorage).to receive(:device_map).and_return(double(empty?: false))
+    end
+
+    it "proposes device map if it is empty" do
+      device_map = double(empty?: true)
+      expect(device_map).to receive(:propose)
+      allow(Yast::BootStorage).to receive(:device_map).and_return(device_map)
+
+      subject.read
+    end
+
+    it "reads bootloader stage1 location" do
+      stage1 = double(Bootloader::Stage1)
+      expect(stage1).to receive(:read)
+      allow(Bootloader::Stage1).to receive(:new).and_return(stage1)
+
+      subject.read
+    end
+  end
+
+  describe "write" do
+    before do
+      stage1 = double(Bootloader::Stage1, model: double(devices: [], generic_mbr?: false), write: nil)
+      allow(Bootloader::Stage1).to receive(:new).and_return(stage1)
+      allow(Bootloader::MBRUpdate).to receive(:new).and_return(double(run: nil))
+    end
+
+    it "writes stage1 location" do
+      stage1 = double(Bootloader::Stage1, model: double(devices: [], generic_mbr?: false))
+      expect(stage1).to receive(:write)
+      allow(Bootloader::Stage1).to receive(:new).and_return(stage1)
+
+      subject.write
+    end
+
+    it "changes pmbr flag as specified in pmbr_action for all boot devices with gpt label" do
+      stage1 = double(Bootloader::Stage1, model: double(devices: ["/dev/sda", "/dev/sdb1"], generic_mbr?: false), write: nil)
+      allow(Bootloader::Stage1).to receive(:new).and_return(stage1)
+
+      allow(Yast::Storage).to receive(:GetDisk) do |m, dev|
+        case dev
+        when "/dev/sda" then { "device" => "/dev/sda", "label" => "msdos" }
+        when "/dev/sdb1" then { "device" => "/dev/sdb", "label" => "gpt" }
+        else raise "unknown device #{dev}"
+        end
+      end
+
+      expect(subject).to receive(:pmbr_setup).with("/dev/sdb")
+
+      subject.write
+    end
+
+    it "runs grub2-install for all configured stage1 locations" do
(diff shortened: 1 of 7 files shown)


ancorgs:
Rubocop got offended and the tests are failing at travis

test/grub2_test.rb
```diff
@@ -0,0 +1,159 @@
...
+    allow(::CFA::Grub2::Default).to receive(:new).and_return(double("GrubDefault").as_null_object)
+    allow(::CFA::Grub2::GrubCfg).to receive(:new).and_return(double("GrubCfg").as_null_object)
+    allow(Bootloader::Sections).to receive(:new).and_return(double("Sections").as_null_object)
+    allow(Yast::BootStorage).to receive(:available_swap_partitions).and_return([])
+    allow(Yast::BootStorage).to receive(:gpt_boot_disk?).and_return(false)
+  end
+
+  describe "#read" do
+    before do
+      allow(Yast::BootStorage).to receive(:device_map).and_return(double(empty?: false))
```
ancorgs: I find quite confusing to find this here and then overridden in one of the only 2 `it` blocks that we have.

test/grub2_test.rb
```diff
@@ -0,0 +1,159 @@
...
+      grub2_install = double(Bootloader::GrubInstall)
+      expect(grub2_install).to receive(:execute).with(devices: ["/dev/sda", "/dev/sdb1"])
+      expect(Bootloader::GrubInstall).to receive(:new).with(efi: false).and_return(grub2_install)
+
+      subject.write
+    end
+
+    context "on s390" do
+      before do
+        allow(Yast::Arch).to receive(:architecture).and_return("s390_64")
```
ancorgs: Hmmm, so we don't mock the `Yast::Arch` for the default case. That could be fragile.

test/grub2_test.rb
```diff
@@ -0,0 +1,159 @@
...
+
+    it "proposes stage1" do
+      stage1 = double
+      expect(stage1).to receive(:propose)
+      allow(Bootloader::Stage1).to receive(:new).and_return(stage1)
+
+      subject.propose
+    end
+
+    it "propose to remove pmbr flag if boot disk is with gpt label" do
```
ancorgs: What about testing the complementary scenario?

test/grub2_test.rb
```diff
@@ -0,0 +1,159 @@
...
+      allow(Yast::BootStorage).to receive(:gpt_boot_disk?).and_return(true)
+
+      subject.propose
+
+      expect(subject.pmbr_action).to eq :remove
+    end
+  end
+
+  describe "#name" do
+    it "returns \"grub2\"" do
```
ancorgs: Does this deserve a test?

test/grub2_test.rb
```diff
@@ -0,0 +1,159 @@
...
+      expect(subject.name).to eq "grub2"
+    end
+  end
+
+  describe "#packages" do
+    it "return list containing grub2 package" do
+      expect(subject.packages).to include("grub2")
+    end
+
+    it "returns list containing syslinux package if generic_mbr is used" do
```
ancorgs: Again, no test for the complementary scenario

test/grub2_test.rb
```diff
@@ -0,0 +1,159 @@
...
+
+  describe "#summary" do
+    before do
+      allow(Bootloader::UdevMapping).to receive(:to_kernel_device) { |d| d }
+      allow(Yast::BootStorage).to receive(:DisksOrder)
+        .and_return(Bootloader::DeviceMap.new("/dev/sda" => "hd0"))
+    end
+
+    it "contain line saying that bootloader type is GRUB2" do
+      expect(subject.summary).to include("Boot Loader Type: GRUB2")
```
ancorgs: Just asking, is this summary i18nzed?

test/grub_install_test.rb
```diff
@@ -0,0 +1,132 @@
+#! /usr/bin/env rspec --format doc
+
+require_relative "./test_helper"
+
+require "bootloader/grub_install"
+
+describe Bootloader::GrubInstall do
+  describe "#execute" do
+    def stub_arch(arch)
```
ancorgs: I know you hate nesting. But this really looks like a mission for `let`

ancorgs:
Just some minor notes. In general looks quite ok... although travis disagrees.

test/grub2_test.rb
jreidinger: well, I want to document API where name is used as bootloader ID.

test/grub2_test.rb
jreidinger: well, if I consider it as specification, then I would like to keep it simple and specify what I did in certain condition. I worry it can quickly turn into not doing this and this and this. So for me from documentation POV it make sense to keep it reasonable short, for testing purpose I agree that we should check also reverse case. There is third view, how danger is opposite action. If we install unneeded package, it is low risk. Above if we propose to remove flag, then it also do nothing as parted just say that non gpt do not have it.

test/grub2_test.rb
jreidinger: yes

test/grub_install_test.rb
jreidinger: Actually I have version with let and I do not like much output. I think nesting make sense where there is at least two stements in context. Something like
```
on s390
  it propose this
  it propose that
  it do that
```
for simple stement in this case it create something like
```
on i386
  use i386-efi
on x86_64
  use x86_64-efi
...
```
where I like more single like statement as it is more like mapping from arch to grub2 target.

test/grub2_test.rb
jreidinger: this is generic mock that prevent to do actions ( as for reading device map or proposing it you have to have complete storage data to get mount points, etc. ) and in one of those two block I test that it really want to do such action.

test/grub2_test.rb
jreidinger: default case is generic that have to happen on all architectures. GRUB2 rules on all architectures :)

test/grub2_test.rb
jreidinger: see below

test/grub2_test.rb
ancorgs: I can live with the current state, so I would give a LGTM if you insist, but for every situation in which we have to decide about testing POV vs documentation POV I would always favor the first one.

We use RSpec primary for having unit tests and secondary for documentation purposes. Even of the output gets longer I prefer to always have the complementary scenarios. On the other hand longer does not necessarily means worst.

test/grub_install_test.rb
ancorgs: Ok. Not that important, actually.

ancorgs:
LGTM

PR #294, opened 2016-03-09
Fix reqires in spec

cwh42:


jreidinger:
thanks. I do mainly manual testing without rpm. LGTM

PR #295, opened 2016-03-09
fix warnings libyui reported and listing available sections

jreidinger:


test/grub2_widgets_test.rb
```diff
@@ -617,8 +617,8 @@ def stub_widget_value(id, value)
     expect(bootloader.sections.default).to eq "openSUSE"
   end
 
-  it "lists all available sections" do
-    expect(subject.list).to eq([
+  it "enlists all available sections" do
```
ancorgs: I found the absence of `describe "#list"` a little bit strange, but it's consistent with the rest of this rspec, so...

ancorgs:
lgtm

test/grub2_widgets_test.rb
jreidinger: reason is that I more describe widget and not its methods as its methods are not directly called. It is called by CWM, so I describe behavior of widget.

PR #298, opened 2016-03-11
Include quiet in cmdline for POWER

k0da:

### DIFF
--- a/test/boot_arch_test.rb
+++ b/test/boot_arch_test.rb
@@ -128,7 +128,7 @@ describe Yast::BootArch do
       it "does not add parameters from boot command line"
     end
 
-    context "on other archs" do
+    context "on POWER archs" do
       before do
         stub_arch("ppc64")
       end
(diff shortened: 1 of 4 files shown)


test/boot_arch_test.rb
```diff
@@ -138,7 +138,11 @@ def stub_arch(arch)
         # just to test that it do not add product features
         allow(Yast::ProductFeatures).to receive(:GetStringFeature).and_return("console=ttyS1")
 
-        expect(subject.DefaultKernelParams("/dev/sda2")).to eq "console=ttyS0 splash=verbose"
+        expect(subject.DefaultKernelParams("/dev/sda2")).to eq "console=ttyS0 resume=/dev/sda2 console=ttyS1 splash=silent quiet showopts"
+      end
+
+      it "adds splash=silent quit showopts parameters" do
+        expect(subject.DefaultKernelParams("/dev/sda2")).to include(" splash=silent quiet showopts")
```
jreidinger: this is now wrong as it is not other archs context, but ppc arch context. Otherwise it looks good.

k0da:
updated

jreidinger:
thanks for update

PR #299, opened 2016-03-22
Jreidinger config files proto

jreidinger:


src/lib/bootloader/grub2.rb
```diff
@@ -61,6 +61,16 @@ def propose
       # TODO: propose device map
     end
 
+    def merge(other)
```
mvidner: Why does grub2base call `super ` but this one doesn't?

jreidinger: bug, thanks

src/lib/bootloader/finish_client.rb
```diff
@@ -46,48 +42,36 @@ def write
...
-          log.info "Installation started with kexec_reboot set 0"
-        end
+      # we do not manage bootloader, so relax :)
+      return true if bl_current.name == "none"
+
+      # read one from system, so we do not overwrite changes done in rpm post install scripts
+      Bootloader::BootloaderFactory.clear_cache
+      system = Bootloader::BootloaderFactory.system
+      system.read
+      system.merge(bl_current)
```
mvidner: So, `merge` is consistent with `Hash#merge!` because in case of conflict, `other` wins. That's good, but the resulting "bl_current overrides system" contradicts the comment above ("so we do not overwrite changes done in rpm scripts"), doesn’t' it?

jreidinger: actually there is two cases what rpm usually did:
1) rpm did branding which we do not manage at all, so bl_current do not define at all, so keep system one ( old way is to replace complete file, so it is lost )
2) kernel parameters, which is in merge appended and not replaced, so it survive

src/lib/bootloader/grub2base.rb
```diff
@@ -125,6 +135,39 @@ def disable_serial_console
...
+
+      # string attributes
+      [:serial_console, :terminal, :timeout, :hidden_timeout, :distributor,
+       :gfx_mode, :theme].each do |attr|
+        default.send((attr.to_s + "="), other.send(attr)) if other.send(attr)
+      end
+
+      # boolean attributes
+      [:os_prober].each do |attr|
+        default.send(attr).value = other.send(attr).enabled? if other.send(attr).defined?
```
mvidner: I wondered whether `defined?` was a mistake until I realized that `default` and `other` were [CFA::BooleanValue](http://www.rubydoc.info/github/config-files-api/config_files_api/master/CFA/BooleanValue). Mention that in a comment?

jreidinger: I can add it. Do you think CFA API is confusing?

mvidner: Er, I mean, that the attr (os_prober) was a BooleanValue, of course

jreidinger: ok, comment added and I remember that I forgot one attribute.

mvidner:
Please do test `merge` also for subclasses, it looks like an easy place to hide bugs in.

jreidinger:
@mvidner plan is to first test this proof of concept with merge method if it works and then I will write all tests for it. Automatic testing makes sense for me only when I proof that it works in reality.

src/lib/bootloader/grub2base.rb
mvidner: @jreidinger I think the CFA API is OK, but the problem with Ruby is that it does not tell the reader what API to expect. Like in the Yast domain, is that variable a string, a primitive value parsed from the string, or a class wrapping that value?

mvidner:
Fair enough, it is too expensive to test while you're still experimenting. Just remember to test once the design has settled down.

jreidinger:
OK, so lets merge it. I add tasks to PBI about adding test to merge method.

PR #300, opened 2016-03-23
Set gfxterm to console on POWER

k0da:

### DIFF
--- a/src/lib/bootloader/grub2base.rb
+++ b/src/lib/bootloader/grub2base.rb
@@ -64,7 +64,10 @@ module Yast
         "default"   => "0",
         "vgamode"   => "",
         "gfxmode"   => "auto",
-        "terminal"  => Arch.s390 ? "console" : "gfxterm",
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/grub2base.rb
```diff
@@ -64,7 +64,7 @@ def StandardGlobals
         "default"   => "0",
         "vgamode"   => "",
         "gfxmode"   => "auto",
-        "terminal"  => Arch.s390 ? "console" : "gfxterm",
+        "terminal"  => (Arch.s390 || Arch.ppc) ? "console" : "gfxterm",
```
jreidinger: do we want to suppress video unconditionally for all ppc? and if so, please add comment also to code with bug number and explaning it is just workaround for slow module loading on LPAR, so when module loading is fixed, we can remove this workaround.

k0da:
Updated

jreidinger:
thanks

PR #301, opened 2016-03-24
Jreidinger config files proto

jreidinger:


src/lib/bootloader/bootloader_base.rb
```diff
@@ -61,9 +62,9 @@ def packages
     end
 
     # done in common write but also in installation pre write as kernel update need it
```
imobach: `pre_write` is self-explanatory, but maybe adding the usual `@param` documentation would be nice.

src/lib/bootloader/device_map.rb
```diff
@@ -2,73 +2,76 @@
...
-      remapped = @mapping.map do |k, v|
-        [UdevMapping.to_kernel_device(k), v]
-      end
+    def disks_order
+      grub_devices = @model.grub_devices
+      grub_devices.select! { |d| d.start_with?("hd") }
+      grub_devices.sort_by! { |dev| dev[2..-1].to_i }
 
-      Hash[remapped]
+      grub_devices.map { |grub_dev| @model.system_device_for(grub_dev) }
```
imobach: np: I would use a single-letter name for `dev` and `grub_dev` variables (according to openSUSE Ruby style guide). Moreover, it would be consistent with the `#select!` block.

jreidinger: ok, make sense, it was more lines, but I reduce it.

src/lib/bootloader/bootloader_base.rb
jreidinger: agreed, will add it.

src/lib/bootloader/device_map.rb
```diff
@@ -84,17 +87,22 @@ def propose
...
       end
 
-      log.info "device map before reduction #{@mapping}"
-      @mapping.select! do |_k, v|
-        v[2..-1].to_i < BIOS_LIMIT
+      grub_devices = @model.grub_devices
+      grub_devices.select! { |d| d.start_with?("hd") }
+      grub_devices.sort_by(&:to_i)
+
+      other_devices = size - grub_devices.size
```
imobach: What about `other_devices_size`? `grub_devices` is a "collection" and to be consistent, I expected `other_devices` to be another collection.

jreidinger: ok, make sense

src/lib/bootloader/grub2.rb
```diff
@@ -30,14 +32,30 @@ def initialize
...
+        log.info "grub_installdevice does not exist. Using empty one."
+        @stage1 = Stage1.new
+      end
+
+      begin
+        # device map is needed only for legacy boot on intel
+        @device_map.read if Yast::Arch.x86_64 || Yast::Arch.i386
+      rescue Errno::ENOENT
+        # grub_installdevice is not part of grub2 rpm, so it doesn't need to exist.
+        # In such case ignore exception and fresh empty @stage1
```
imobach: Copy & pasted from prior rescue block. Please, update.

src/lib/bootloader/grub2.rb
```diff
@@ -30,14 +32,30 @@ def initialize
...
+        log.info "grub2/device.map does not exist. Using empty one."
+        @device_map = DeviceMap.new
+      end
     end
 
     # Write bootloader settings to disk
     # @return [Boolean] true on success
     def write
-      # TODO: device map write
+      @device_map.write if Yast::Arch.x86_64 || Yast::Arch.i386
```
imobach: I would use `device_map` method as you do in line L78.

src/lib/bootloader/grub2.rb
```diff
@@ -121,6 +151,14 @@ def packages
 
   private
 
+    def replace_with(other)
+      stage1.clear_devices
```
imobach: Would it make sense to have some method `#Stage1#replace` which implements this logic?

src/lib/bootloader/grub2.rb
```diff
@@ -132,13 +170,12 @@ def gpt_disks_devices
     def disk_order_summary
       return "" if Yast::Arch.s390
 
-      order = Yast::BootStorage.DisksOrder
-      return "" if order.size < 2
+      return "" if @device_map.size < 2
```
imobach: `device_map` instead of `@device_map`?

src/lib/bootloader/grub2.rb
```diff
@@ -132,13 +170,12 @@ def gpt_disks_devices
...
 
       Yast::Builtins.sformat(
         # part of summary, %1 is a list of hard disks device names
         _("Order of Hard Disks: %1"),
-        order.join(", ")
+        @device_map.disks_order(", ")
```
imobach: Same as above.

src/lib/bootloader/grub2base.rb
```diff
@@ -151,14 +152,23 @@ def merge_grub_default(other)
...
+
+      log.info "after merge default #{default.inspect}"
+    end
+
+    def merge_attributes(default, other)
       # string attributes
       [:serial_console, :terminal, :timeout, :hidden_timeout, :distributor,
-       :gfx_mode, :theme].each do |attr|
+       :gfxmode, :theme].each do |attr|
         default.send((attr.to_s + "="), other.send(attr)) if other.send(attr)
```
imobach: np: Maybe you could use a local variable to avoid calling `#send` twice in the same line.

src/lib/bootloader/grub_install.rb
```diff
@@ -25,10 +25,10 @@ def execute(devices: nil, secure_boot: false)
 
       if devices
         devices.each do |dev|
-          Yast::Execute.locally(cmd + [dev])
+          Yast::Execute.on_target(cmd + [dev])
```
imobach: This reminds me that I would start using `Yast::Execute` instead of the old `SCR::Execute` :)

imobach:
Just some minor notes. It's a nice improvement also in terms of encapsulation and that kind of stuff.

src/lib/bootloader/grub2.rb
jreidinger: well, I am not sure. Originally I plan to have it in CFA, but then I found that there are more strategies for merge. So I move it to given bootloader itself to drive its merge strategy.

imobach:
LGTM

jreidinger:
thanks

imobach:
It still LGTM :) It looks like the problem in Travis has nothing to do with our code, right?

jreidinger:
no, one of travis that failed is again problem with cached gettext on some travis nodes.

src/lib/bootloader/device_map.rb
```diff
@@ -84,17 +87,22 @@ def propose
...
         log.info "device map not need to be reduced"
         return false
       end
 
-      log.info "device map before reduction #{@mapping}"
-      @mapping.select! do |_k, v|
-        v[2..-1].to_i < BIOS_LIMIT
+      grub_devices = @model.grub_devices
+      grub_devices.select! { |d| d.start_with?("hd") }
+      grub_devices.sort_by(&:to_i)
```
teclator: It's the second time i see this block, maybe could be moved to a method, and also what is the difference here with line 62?

jreidinger: good note, in fact, there is a bug as to_i with "hd" return false...I can use here disks_order method.

PR #302, opened 2016-03-29
dry code

jreidinger:


teclator:
LGTM

src/lib/bootloader/device_map.rb
```diff
@@ -57,11 +57,7 @@ def contain_disk?(disk)
     end
 
     def disks_order
-      grub_devices = @model.grub_devices
-      grub_devices.select! { |d| d.start_with?("hd") }
-      grub_devices.sort_by! { |dev| dev[2..-1].to_i }
-
-      grub_devices.map { |d| @model.system_device_for(d) }
+      sorted_disks.map { |d| @model.system_device_for(d) }
     end
```
teclator: NP: Maybe i would name it ordered_disks as we used sorted_disks method, but i'm not the best to recommend method names :)

src/lib/bootloader/device_map.rb
```diff
@@ -92,9 +94,7 @@ def reduce_to_bios_limit
         return false
       end
 
-      grub_devices = @model.grub_devices
-      grub_devices.select! { |d| d.start_with?("hd") }
-      grub_devices.sort_by(&:to_i)
```
mvidner: @jreidinger @teclator This is actually a bug fix, because
```rb
(irb)> %w(hd3 hd0 hd1 hd2).sort_by(&:to_i)
=> ["hd3", "hd0", "hd1", "hd2"]
```

But since this is the cfa branch, nobody has probably seen the bug anyway.

src/lib/bootloader/device_map.rb
jreidinger: it is even more tricky, as it is sorted grub devices, but then it is too long method for me...and it is not just grub devices, but grub devices representing disks ( we filter out all non hd devices ). So I expect that good name is very hard to find.

src/lib/bootloader/device_map.rb
teclator: @mvidner yeah, it was the reason of the previous comment but also it is not using "!" so grub_devices is not affected after sort_by and you are not assigned to any variable.

jreidinger: yes, @teclator already discover it. And test suite do not catch it as it checks only number if it is properly reduced to 8 disks and not if correct ones are removed. ( I know it is discutable, but for me having 8 with wrong order is not so bad as having more then 8 causing grub2 to refuse work ).

PR #304, opened 2016-04-01
Jreidinger config files proto

jreidinger:


src/lib/bootloader/proposal_client.rb
```diff
@@ -46,12 +47,18 @@ def make_proposal(attrs)
 
       if Yast::Mode.update
         return { "raw_proposal" => [_("do not change")] } unless propose_for_update(force_reset)
+      elsif Yast::Bootloader.proposed_cfg_changed
+        # do nothing as user already modify it
       else
         # in installation always propose missing stuff
```
lslezak: Is the comment still valid? The `propose` method call has been replaced...

jreidinger: it is two different propose. First propose is to `proposed` correct bootloader type ( now replaced by current, which in installation call propose if not already set ) and then is called real `propose` method, which propose configuration for given bootloader type ( as e.g. propose for grub2 and grub2-efi is different ).

lslezak: OK, could you add this info as comments into the code? It would help reading the code...

jreidinger: added

lslezak:
LGTM, thanks!

PR #305, opened 2016-04-04
propose and use suse_btrfs specific config

jreidinger:


ancorgs:
Using refinements looks a little bit weird, but I can understand the motivation of keeping CFA clean. So LWBGTM (Looks Weird But Good To Me) :wink:

PR #307, opened 2016-04-06
Jreidinger config files proto

jreidinger:


src/lib/bootloader/sections.rb
```diff
@@ -22,10 +22,10 @@ def default
 
       return @default = "" if Yast::Stage.initial
 
-      saved = Yast::Execute.on_target("/usr/bin/grub2-editenv", "list", stdout: :capture)
+      saved = Yast::Execute.on_target("/usr/bin/grub2-editenv", "list", stdout: :capture) || ""
```
mvidner: Side note about `Yast::Execute.on_target`:
I have a nice little app to help me look up YaST API, http://mvidner-vm3.suse.cz:4567/?q=on_target
which helps me find http://www.rubydoc.info/github/yast/yast-yast2/Yast/Execute.on_target
but it does not say what the return value is and the link to Cheetah.run is not rendered. (we should quote the full URL, http://www.rubydoc.info/github/openSUSE/cheetah/Cheetah.run )
... and now I see that Cheetah does not mention the falsey return value either)

jreidinger: it is not false, it is nil...and maybe it can return nil only in testsuite?

test/grub2_efi_test.rb
```diff
@@ -4,7 +4,9 @@
 
 describe Bootloader::Grub2EFI do
   before do
-    allow(::CFA::Grub2::Default).to receive(:new).and_return(double("GrubDefault").as_null_object)
+    default = double("GrubDefault").as_null_object
+    allow(default).to receive(:timeout) # collision between timeout method of double and attr
```
mvidner: Huh? There is  CFA::Grub::Default#timeout and which other timeout?

jreidinger: just try it. I expect it is method of double object, so as_null_object do not respond as it is already defined. It report error (expected 1..2 arguments when removed )

jreidinger: what is interesting that `default.timeout` work as expected but `default.send(:timeout)` raise exception, so I expect problem with some private method.

jreidinger: another possible fix is to use `public_send` instead of `send` which is probably better solution from my POV

test/grub2base_test.rb
```diff
@@ -348,4 +348,75 @@
...
+      other.grub_default.os_prober.disable
+
+      subject.merge(other)
+
+      expect(subject.grub_default.default).to eq "saved"
+      expect(subject.grub_default.terminal).to eq :gfxterm
+      expect(subject.grub_default.os_prober).to be_disabled
+    end
+
+    it "for kernel line place subject params and then merged ones" do
```
mvidner: What if one says `debug=true` and the other says `debug=false`?

jreidinger: on kernel precedence is for later, that is why I use it this way. Do you have idea for better merge strategy? Problem is that rpm in post install script can add something to kernel command line and then we have trouble if we lost it.

mvidner: I don't care which way it is resolved, but it should be specified, as I expect it to be a common case.

mvidner:
Thank you, LGTM.

PR #308, opened 2016-04-06
fix inverted logic for pmbr flag causing not working uefi boot

jreidinger:
Manual testing showed that I accidentally reverted logic and after reread of mentioned bug report I fix proposal to correct values.

### DIFF
--- a/src/lib/bootloader/grub2.rb
+++ b/src/lib/bootloader/grub2.rb
@@ -73,7 +73,7 @@ module Bootloader
 
       stage1.propose
       # for GPT add protective MBR flag otherwise some systems won't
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/grub2.rb
```diff
@@ -72,9 +72,9 @@ def propose
       super
 
       stage1.propose
-      # for GPT remove protective MBR flag otherwise some systems won't
+      # for GPT add protective MBR flag otherwise some systems won't
       # boot, safer option for legacy booting
```
mvidner: OK, please mention  (bnc#872054) also here.

mvidner:
LGTM.

PR #309, opened 2016-04-07
Jreidinger config files proto

jreidinger:


src/lib/bootloader/auto_client.rb
```diff
@@ -74,7 +76,16 @@ def change
     #
     # return map or list
     def export
-      Export2AI(Yast::Bootloader.Export)
+      # it is needed to have information about storage configuration to understand current config
+      Yast::BootStorage.detect_disks
+
+      config = BootloaderFactory.current
+      config.read if !config.read? && !config.proposed?
```
imobach: Maybe it's just a matter of taste, but I would write:
```
config.read unless config.read? || config.proposed?
```

jreidinger: well, it is my internal rule to use `unless` only if there is single parameter, this way it is more like `if !(config.read? || config.proposed?)` is need at least for me more brain capacity to read it :)

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -0,0 +1,90 @@
...
+      STAGE1_MAPPING = {
+        "activate"      => :activate?,
+        "generic_mbr"   => :generic_mbr?,
+        "boot_root"     => :root_partition?,
+        "boot_boot"     => :boot_partition?,
+        "boot_mbr"      => :mbr?,
+        "boot_extended" => :extended_partition?
+      }
+      def export_stage1(res, stage1)
+        res["global"] ||= {}
```
imobach: Side effects are not nice from my POV. I would prefer other approach here. Or, at least, document it please.

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      DEFAULT_KERNEL_PARAMS_MAPPING = {
+        "append"            => :kernel_params,
+        "xen_append"        => :xen_kernel_params,
+        "xen_kernel_append" => :xen_hypervisor_params
+      }
+      def export_default(res, default)
+        DEFAULT_BOOLEAN_MAPPING.each do |key, method|
+          val = default.public_send(method)
+          res["global"][key] = val.enabled? ? "true" : "false" if val.defined?
```
imobach: I was about to say that you could define some kind of helper method to convert from boolean to string, but it looks like this is the only place where is needed.

src/lib/bootloader/autoyast_converter.rb
jreidinger: what side effect you mean? goal of this method is to fill res with information from stage1. I do not see any side-effect

src/lib/bootloader/autoyast_converter.rb
jreidinger: yes, used only here and to be more funny it is string and not boolean xml attribute due to old p-bl map which use strings.

imobach:
Assuming that documentation and test cases will be added later, LGTM.

src/lib/bootloader/auto_client.rb
imobach: Hehe, no problem, as I said is just a matter of taste :)

src/lib/bootloader/autoyast_converter.rb
imobach: Well, you're modifying the argument you receive (`res`) and that's a side efect. `export_stage1` could return a new hash and the caller could be the responsible for merging. But, as I have said, I'm happy as long as it's documented.

imobach:
Looks better, thanks. LGTM.

PR #312, opened 2016-04-12
Jreidinger config files proto

jreidinger:


mchf:
It LGTM from my top level point of view if there is no need of ".changes"

jreidinger:
@mchf no, it is just merge to topic branch. I will add changes when merge to master going to happen

imobach:
It also LGTM :)

PR #313, opened 2016-04-12
fix writting btrfs on btrfs as grub2-install need some data to be alr…

jreidinger:
…eady written into /etc/default/grub


schubi2:
LGTM

PR #317, opened 2016-04-13
clean TODOs and implement back some Bootloader API calls

jreidinger:


src/lib/bootloader/auto_client.rb
```diff
@@ -26,19 +26,10 @@ def run
...
-      proposed_configuration = BootloaderFactory.bootloader_by_name(imported_configuration.name)
-      proposed_configuration.propose
-
-      proposed_configuration.merge(imported_configuration)
-      BootloaderFactory.current = proposed_configuration
+      Yast::Bootloader.Import(data)
 
       Yast::PackagesProposal.AddResolvables("yast2-bootloader",
-        :package, proposed_configuration.packages)
+        :package, BootloaderFactory.current..packages)
```
imobach: It looks like you have "two points" instead of only one.

jreidinger: ah, you are right, typo

src/lib/bootloader/grub2base.rb
```diff
@@ -153,17 +153,21 @@ def merge_password(other)
...
 
       log.info "before merge default #{default.inspect}"
       log.info "before merge other #{other.inspect}"
 
-      # TODO: other kernel flavors probably will be also needed
-      unless other.kernel_params.serialize.empty?
-        new_kernel_params = default.kernel_params.serialize + " " + other.kernel_params.serialize
-        default.kernel_params.replace(new_kernel_params)
+      KERNEL_FLAVORS_METHODS.each do |method|
+        next if  other.public_send(method).empty?
```
imobach: np: you have two spaces instead of just one :)

src/modules/Bootloader.rb
```diff
@@ -201,9 +168,16 @@ def Read
     def Reset
       return if Mode.autoinst
       log.info "Reseting configuration"
```
imobach: Rese*tt*ing

imobach:
If Travis is ok, LGTM.

src/lib/bootloader/grub2base.rb
jreidinger: good catch ;)

PR #318, opened 2016-04-13
[blog] Better pwd

jreidinger:
Bootloader
--------------

Password protection widget in bootloader do not change for a long time from GRUB1 ages. But GRUB2 allows more fine tunning of password. In fact it contain whole user management with multiple users that can have different permissions and different passwords. For YaST we would like to keep things simplier, so we support only password for "root" user. It confuse users, so we decide to improve UX and change labeling in widget to make it more obvious for anyone using password protection. Beside it also some typos are fixed.

Before change it looks like on image below
how it looks before

and after change it adds more explanation to password specification. Of course also help is improved to reflect such change.
how it looks now

### DIFF
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -308,7 +308,7 @@ module Bootloader
             HBox(
               HSpacing(2),
               # text entry
-              Password(Id(:pw1), Opt(:hstretch), _("&Password for GRUB2 user 'root'")),
+              Password(Id(:pw1), Opt(:hstretch), _("&Password for GRUB2 User 'root'")),
               # text entry
               HSpacing(2),
               Password(Id(:pw2), Opt(:hstretch), _("Re&type Password")),
@@ -386,9 +386,9 @@ module Bootloader
           "booting any entry is not restricted but modifying entries requires " \
           "the password (which is the way GRUB 1 behaved).<br>" \
           "YaST will only accept the password if you repeat it in " \
-          "<b>Retype Password</b>. Password is specified for user 'root'. YaST2" \
-          "bootloader currently do not support other users and if needed to use," \
-          " then it have to be specified manually.</p>"
+          "<b>Retype Password</b>. The password applies to the GRUB2 user 'root' " \
+          "which is distinct from the Linux 'root'. YaST currently does not support" \
+          "other GRUB2 users. If you need them, use a separate GRUB2 script.</p>"
       )
     end
   end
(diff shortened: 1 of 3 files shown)


mvidner:
Before I forget: the capitalization in the tab headers is wrong, should be "Boot Code Options" and "Kernel Parameters"

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -386,7 +386,9 @@ def help
           "booting any entry is not restricted but modifying entries requires " \
           "the password (which is the way GRUB 1 behaved).<br>" \
           "YaST will only accept the password if you repeat it in " \
-          "<b>Retype Password</b>.</p>"
+          "<b>Retype Password</b>. Password is specified for user 'root'. YaST2" \
+          "bootloader currently do not support other users and if needed to use," \
+          " then it have to be specified manually.</p>"
```
mvidner: > The password applies to the GRUB2 user 'root' which is distinct  from the Linux 'root'. YaST currently does not support other GRUB2 users. If you need them, use a separate GRUB2 script.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -308,7 +308,7 @@ def contents
             HBox(
               HSpacing(2),
               # text entry
-              Password(Id(:pw1), Opt(:hstretch), _("&Password")),
+              Password(Id(:pw1), Opt(:hstretch), _("&Password for GRUB2 user 'root'")),
```
mvidner: capitalize User

mvidner:
LGTM

jreidinger:
Ken approves it, I just would like to wait after Alpha3 as we would like to get new bootloader there.

kobliha:
Great stuff :) I especially appreciate the @lslezak-Style PR :)

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -308,7 +308,7 @@ def contents
             HBox(
               HSpacing(2),
               # text entry
-              Password(Id(:pw1), Opt(:hstretch), _("&Password")),
+              Password(Id(:pw1), Opt(:hstretch), _("&Password for GRUB2 User 'root'")),
```
kobliha: NTH: you might also change the comment for translators saying something like: $this and $that labels have to both fit into XX characters (XX < 80) as they will appear next to each other on the screen. If really needed, you might split it to two lines.

jreidinger: OK, will add it. I think something like keep it short is enough as problem is only in ncurses in small window

PR #319, opened 2016-04-14
fixed module Bootloader

schubi2:


teclator:
LGTM but if it failed could be nice to have a test for Export method.

schubi2:
As far I see they are on Josefs todo list :-)

jreidinger:
in fact I have on list autoyast converter, but export import method maybe also make sense.

PR #321, opened 2016-04-18
Fix serial console

jreidinger:


coolo:
the code looks good - your 'chagnes' typo not :)

jreidinger:
@coolo thanks

PR #323, opened 2016-04-21
Fix lvm boot

jreidinger:


package/yast2-bootloader.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Apr 20 15:18:46 UTC 2016 - jreidinger@suse.com
+
+- fix installing grub2 to underlaying devices (bnc#976315)
```
imobach: "underlying" or "underlaying"? I'm not sure at all.

jreidinger:
```
underlay
verb
gerund or present participle: underlaying
ʌndəˈleɪ/
place something under (something else), especially to support or raise it.
"the green fields are underlaid with limestone"
```

versus 
```
underlying
ʌndəˈlʌɪɪŋ/Submit
present participle of underlie.
synonyms:	latent, lurking, repressed, suppressed, unrevealed, undisclosed, unexpressed, undivulged, concealed, hidden, veiled, masked, shrouded, under wraps
"she spoke airily despite an underlying feeling of irritation"
```

so I do not lie, but lay it :)

src/lib/bootloader/stage1.rb
```diff
@@ -144,33 +140,29 @@ def available_locations
...
       end
 
-      log.info "md_disks: #{disks.inspect} laying on #{partitions.inspect}"
+      log.info "Boot partition info #{part.inspect}"
 
-      disks
+      # cannot install stage one to xfs as it doesn't have reserved space (bnc#884255)
+      return false if part["used_fs"] == :xfs
+
+      true
```
imobach: I would write something like (more concise from my POV):

```ruby
if part
  log.info "Boot partition info #{part.inspect}"
  part["used_fs"] != :xfs
else
  log.error "cannot find partition #{partition}"
  false
end
```

src/lib/bootloader/stage1.rb
```diff
@@ -351,12 +342,7 @@ def assign_bootloader_device(selected_location)
     # BootPartitionDevice, or the devices from which the soft-RAID device for
     # "/boot" is built)
     def underlying_boot_partition_devices
```
imobach: `underlaying` or `underlying`? :)

src/modules/BootStorage.rb
```diff
@@ -114,174 +116,51 @@ def extended_partition_for(device)
...
-
-        if p["raid_type"] == "raid1"
-          p_devices = p["devices"] || []
-          if p_devices.size == 2 # TODO: why only 2? it do not make sense
-            ret = checkDifferentDisks(p_devices)
-          else
-            log.info "Device: #{device} doesn't contain 2 partitions: #{p_devices}"
+    # returns device where dev physically lives, so where can be bootloader installed
+    # it is main entry point when real stage 1 device is needed to get
+    def underlaying_devices(dev)
```
imobach: Please, add some documentation about params and return value.

test/boot_storage_test.rb
```diff
@@ -4,20 +4,68 @@
...
+      # clear cache
+      subject.instance_variable_set(:@underlaying_devices_cache, {})
+      allow(subject).to receive(:underlaying_devices).and_call_original
+
+      # nasty hack to allow call of uninitialized libstorage as we do not want
+      # to overmock Yast::Storage.GetDiskPartitionTg call
+      Yast::Storage.instance_variable_set(:@sint, double(getPartitionPrefix: "").as_null_object)
+    end
+
+    it "returns for physical device itself in array" do
```
imobach: `for`?

test/stage1_test.rb
```diff
@@ -66,4 +72,69 @@
...
+    it "returns false if boot partition fs is xfs" do
+      target_map_stub("storage_xfs.yaml")
+
+      allow(Yast::BootStorage).to receive(:mbr_disk).and_return("/dev/vda")
+      allow(Yast::BootStorage).to receive(:BootPartitionDevice).and_return("/dev/vda1")
+
+      expect(subject.can_use_boot?).to eq false
+    end
+
+    it "returns true otherwise" do
```
imobach: np: I prefer to write the whole condition: "returns tru if boot partition is not xfs". It looks better, for example, if only this test fails (it shows the whole description at the end).

src/lib/bootloader/stage1.rb
jreidinger: well, there can was more conditions in past, that is reason why I wrote it this way and I expect it will be extended in future.

test/stage1_test.rb
jreidinger: in fact it was list of conditions where it returns false and this was the last one to specifi rest of states

imobach:
LGTM

mvidner:
We missed that `Yast::BootStorage.Md2Partitions` is still used in [src/lib/bootloader/mbr_update.rb:146:](https://github.com/yast/yast-bootloader/blob/ac8b6620b1953b79d6b9fb51be1f22315bef8e00/src/lib/bootloader/mbr_update.rb#L146)  
```rb
    md = Yast::BootStorage.Md2Partitions(md_device)
```
and it was now hit by a customer in https://bugzilla.suse.com/show_bug.cgi?id=995627

PR #324, opened 2016-04-22
A3 fixes

jreidinger:


test/stage1_test.rb
```diff
@@ -51,17 +51,42 @@
...
+        allow(Yast::Arch).to receive(:architecture).and_return("ppc64")
 
-      subject.propose
+        object_double(
+          "Yast::BootStorage",
+          prep_partitions: ["/dev/sda1", "/dev/sdb1", "/dev/sdc1"],
+          detect_disks:    nil,
+          disk_with_boot_partition: "/dev/sdb"
+        ).as_stubbed_const
+
```
imobach: Interesting :)

test/test_helper.rb
```diff
@@ -68,6 +68,11 @@ def mock_disk_partition
 
 # stub udev mapping everywhere
 RSpec.configure do |config|
+  Yast.import "Storage"
+  Yast.import "BootStorage"
+  Yast.import "Bootloader"
+  require "bootloader/udev_mapping"
```
imobach: Is this needed for all tests?

imobach:
LGTM

test/test_helper.rb
jreidinger: as it is below mocked, we need it to have such constant defined

PR #325, opened 2016-04-22
fix device name by-path handling (bsc #976228)

wfeldt:
The relevant keys in Storage::GetTargetMap() are named 'udev_path', not 'path'.


jreidinger:
thanks for fix...it used to work on past or never worked?

jreidinger:
LGTM

wfeldt:
ISTR that the target map entry names changed a bit at some point in time

aschnell:
That change must have been a looong time ago.

wfeldt:
Hm, git blame says the code was done in one go. So it looks just like a plain bug.

wfeldt:
... and never worked

PR #327, opened 2016-04-27
adapted to autoyast config module

schubi2:


src/lib/bootloader/udev_mapping.rb
```diff
@@ -23,6 +23,9 @@ class << self
     # @param dev [String] device udev, mdadm or kernel name like /dev/disk/by-id/blabla
     # @raise when device have udev format but do not exists
     def to_kernel_device(dev)
+      # AutoYaST configuration mode. There is no access to the system
+      return nil if Yast::Mode.config
```
imobach: We should document the `@return` value (I know it wasn't documented before, but it would be nice).

src/modules/Bootloader.rb
```diff
@@ -90,16 +90,18 @@ def Export
     # @param [Hash] data map of bootloader settings
     # @return [Boolean] true on success
     def Import(data)
-      Yast::BootStorage.detect_disks
+      # AutoYaST configuration mode. There is no access to the system
+      Yast::BootStorage.detect_disks unless Mode.config
```
imobach: Does `detect_disks` returns a boolean?

schubi2: The return value is quite undefined and will be be evaluated by any call. The docu says:
 # Sets properly boot, root and mbr disk.

src/lib/bootloader/udev_mapping.rb
```diff
@@ -22,6 +22,7 @@ class << self
     # Converts full udev name to kernel device ( disk or partition )
     # @param dev [String] device udev, mdadm or kernel name like /dev/disk/by-id/blabla
     # @raise when device have udev format but do not exists
+    # @return kernel device[String] or nil (AutoYaST configuration)
```
imobach: `@return [String,nil] kernel device or nil when running AutoYaST configuration`.

src/lib/bootloader/udev_mapping.rb
```diff
@@ -49,6 +50,7 @@ def to_kernel_device(dev)
     # option or kernel device if such udev device do not exists
     # @param dev [String] device udev or kernel one like /dev/disk/by-id/blabla
     # @raise when device have udev format but do not exists
+    # @return udev name[String]
```
imobach: `@return [String] udev name`

imobach:
LGTM :)

PR #328, opened 2016-05-02
Removed empty reject

schubi2:


package/yast2-bootloader.changes
```diff
@@ -1,10 +1,18 @@
 -------------------------------------------------------------------
+Mon May  2 10:24:17 CEST 2016 - schubi@suse.de
+
+- Fixed nil in custom_devices. [[]].compact! is returning nil.
```
mvidner: I thought the array contained Strings (or nil), so the `[[]]` is misleading, isn't it?

mvidner: Anyway, the changes file is user oriented so we should not fill it with implementation details (they can go to comments or commit messages). I would say
> Fixed an "undefined method reject! ... in custom_devices" (bnc#977945)

schubi2: Hm, as far I see it is returning also array in array too. Otherwise I cannot reproduce this error that custom_devices is nil. Could it be that we have different data types in this array and have a complete different bug here which has been found by the "reject" call ? Will check again......

mvidner:
LGTM

PR #329, opened 2016-05-04
trusted boot

mvidner:
https://trello.com/c/VlEwlLX4
https://fate.suse.com/316553

### DIFF
--- a/src/lib/bootloader/autoyast_converter.rb
+++ b/src/lib/bootloader/autoyast_converter.rb
@@ -29,6 +29,7 @@ module Bootloader
         # let it be empty if not defined to keep code simplier as effect is same
         data["global"] ||= {}
 
+        import_grub2(data, bootloader)
         import_stage1(data, bootloader)
         import_default(data, bootloader.grub_default)
         # TODO: import Initrd
@@ -48,6 +49,7 @@ module Bootloader
 
         res["global"] = {}
         global = res["global"]
+        export_grub2(global, config) if config.name == "grub2"
         export_stage1(global, config.stage1) if config.respond_to?(:stage1)
         export_default(global, config.grub_default)
 
@@ -55,6 +57,17 @@ module Bootloader
       end
 
     private
+
+      def import_grub2(data, bootloader)
+        return unless bootloader.name == "grub2"
+
+        GRUB2_BOOLEAN_MAPPING.each do |key, method|
+          val = data["global"][key]
+          next unless val
+
+          bootloader.public_send(:"#{method}=", val == "true")
+        end
+      end
 
       def import_default(data, default)
         DEFAULT_BOOLEAN_MAPPING.each do |key, method|
@@ -178,6 +191,17 @@ module Bootloader
         end
 
         res["boot_custom"] = stage1.custom_devices.join(",") unless stage1.custom_devices.empty?
+      end
+
+      # only for grub2, not for others
+      GRUB2_BOOLEAN_MAPPING = {
+        "trusted_grub" => :trusted_boot
+      }
+      def export_grub2(res, bootloader)
+        GRUB2_BOOLEAN_MAPPING.each do |key, method|
+          val = bootloader.public_send(method)
+          res[key] = val ? "true" : "false" unless val.nil?
+        end
       end
 
       DEFAULT_BOOLEAN_MAPPING = {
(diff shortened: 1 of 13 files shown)


src/lib/bootloader/grub2.rb
```diff
@@ -79,12 +83,14 @@ def propose
       # boot, safer option for legacy booting (bnc#872054)
       self.pmbr_action = :add if Yast::BootStorage.gpt_boot_disk?
       device_map.propose if Yast::Arch.x86_64 || Yast::Arch.i386
+      @trusted_boot = true # FIXME: another default value
```
jreidinger: I really do not think it is good idea to propose trusted boot to true by default

src/lib/bootloader/grub2.rb
```diff
@@ -149,9 +159,21 @@ def packages
...
+      if Yast::Arch.x86_64
+        # FIXME: both? arch dependency?
+        res << "trustedgrub2" << "trustedgrub2-i386-pc" if @trusted_boot
+      end
+
       res
     end
 
+    # FIXME: refactor with injection like super(prewrite: prewrite, sysconfig = ...)
+    # overwrite BootloaderBase version to save trusted boot
```
jreidinger: I think it is fine this way, grub2efi also use it this way.

src/lib/bootloader/grub2.rb
```diff
@@ -149,9 +159,21 @@ def packages
...
+    def write_sysconfig(prewrite: false)
+      sysconfig = Bootloader::Sysconfig.new(bootloader: name, trusted_boot: @trusted_boot)
```
jreidinger: this is not so nice as it write this sysconfig value even on non x86 architectures, so if someone check that file, he is confused

src/lib/bootloader/sysconfig.rb
```diff
@@ -61,6 +67,18 @@ def pre_write
...
+      # FIXME: the default?
+      trusted_boot: "\n" \
+        "## Path:\tSystem/Bootloader\n" \
+        "## Description:\tBootloader configuration\n" \
+        "## Type:\tyesno\n" \
+        "## Default:\t\"no\"\n" \
+        "#\n" \
+        "# Enable Trusted Boot support\n" \
+        "# FIXME here we explain the benefits and distinction from Secure Boot\n" \
+        "# Only available for legacy (non-UEFI) boot.\n" \
```
jreidinger: and only on certain architectures. Ideally I prefer to write it only where it make sense.

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -180,6 +193,17 @@ def export_stage1(res, stage1)
         res["boot_custom"] = stage1.custom_devices.join(",") unless stage1.custom_devices.empty?
       end
 
+      # only for grub2, not for others
+      GRUB2_BOOLEAN_MAPPING = {
+        "trusted_grub" => :trusted_boot
+      }
```
jreidinger: I prefer to have constant definition near others and ideally before its first usage (in this case in import)

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -787,6 +816,10 @@ def secure_boot_widget?
       (Yast::Arch.x86_64 || Yast::Arch.i386) && grub2.name == "grub2-efi"
     end
 
+    def trusted_boot_widget?
+      (Yast::Arch.x86_64 || Yast::Arch.i386) && grub2.name != "grub2-efi"
```
jreidinger: I prefer to use `grub2.name != "grub2"` otherwise here can be confusion that it is shown by `"none"` bootloader

src/lib/bootloader/stage1.rb
```diff
@@ -159,6 +159,31 @@ def can_use_boot?
...
+      else
+        clear_devices
+        other.devices.each { |d| add_udev_device(d) }
+
+        self.activate    = other.activate?
+        self.generic_mbr = other.generic_mbr?
+      end
+
+      log.info "stage1 after merge #{inspect}"
+    end
```
jreidinger: :+1:

src/lib/bootloader/autoyast_converter.rb
mvidner: Yes, but it is consistent with other constants around, so we agreed to postpone this.

jreidinger:
LGTM

PR #330, opened 2016-05-05
Only use secure boot on x86_64 (bsc#978157)

agraf:


agraf:
The failing tests look like broken tests to me. Do we set Arch.x86_64 inside the tests?

agraf:
Ok, I updated the tests to properly verify that secure boot gets enabled on x86_64 and disabled on aarch64.

I also added a patch to fix (parts of) bug 978593.

wfeldt:
lgtm

PR #332, opened 2016-05-10
changes

jreidinger:


ancorgs:
LGTM. Not bad for a sick guy :wink:

wfeldt:
lgtm

mvidner:
@jreidinger Thank you! But I have some questions :)
What about [line 345](https://github.com/yast/yast-bootloader/blob/6b17ac1971a5ca3504ae5d5881a0cdae21faa4bc/src/lib/bootloader/stage1.rb#L345)?
Also, is it intended to have both `extended_partition?` and `extended_partition`? How should I choose which one to use?

jreidinger:
@mvidner `extended_partition?` method is public available one which specify if stage1 bootloader code is installed in extended partition. `extended_partition` on other hand is just shortcut to value of such extended partition.
and good catch with line 345, it have same typo. I will fix it.

PR #333, opened 2016-05-11
fix extended typo (thanks mvidner)

jreidinger:


mvidner:
Code looks good, but we still have a hole in code coverage: 
https://coveralls.io/builds/6134237/source?filename=src%2Flib%2Fbootloader%2Fstage1.rb#L345
should we fix it right now?

jreidinger:
@mvidner I need specific target map that have /boot on logical partition to test it for real.

mvidner:
I think I have such a machine at home, running 13.2. Will it work if I run `Yast::Storage.GetTargetMap` and dump that to YAML?

jreidinger:
@mvidner yes, then I will add it, but I prefer to release fix soon and then add test when you dump it

mvidner:
Sounds good!
LGTM.

mvidner:
(This is Pull Řequest numbeř Třista třicet tři! We Řule!)

PR #338, opened 2016-05-20
fix device name by-path handling (bsc#976228)

wfeldt:
The relevant keys in Storage::GetTargetMap() are named 'udev_path', not 'path'.

This is the same as https://github.com/yast/yast-bootloader/pull/325.


jreidinger:
LGTM

PR #340, opened 2016-05-25
Handle boot of mixed LVM with combination of partition-less and partition based volumes

jreidinger:


src/modules/BootStorage.rb
```diff
@@ -139,22 +139,27 @@ def underlaying_devices(dev)
...
         elsif disk["type"] == :CT_LVM
           # not happy with this usage of || but target map do not need to have it defined
-          res = (disk["devices"] || []) + (disk["devices_add"] || [])
+          res = devices_on(disk)
+          # skip lvm on partiotionless disks as it cannot be used, see bnc#980529
+          res.delete_if { |d| tm[d] }
           res.map! { |r| Yast::Storage.GetDiskPartition(r)["disk"] }
         end
-      else
+      else # partition
```
kobliha: `partition` what?

src/modules/BootStorage.rb
```diff
@@ -166,6 +171,11 @@ def underlaying_devices(dev)
       res
     end
 
+    # @private
+    def devices_on(dev)
```
kobliha: Missing docu: what it does? returns list of partitions on device? or devices in LVM or ... ?

src/modules/BootStorage.rb
```diff
@@ -139,22 +139,27 @@ def underlaying_devices(dev)
...
         part = Yast::Storage.GetPartition(tm, dev)
         if part["type"] == :lvm
           lvm_dev = Yast::Storage.GetDisk(tm, disk_data["disk"])
-          res = (lvm_dev["devices"] || []) + (lvm_dev["devices_add"] || [])
+          # skip lvm on partiotionless disks as it cannot be used, see bnc#980529
+          res = devices_on(lvm_dev)
+          res.delete_if { |d| tm[d] }
         elsif part["type"] == :sw_raid
-          res = (part["devices"] || []) + (part["devices_add"] || [])
+          res = devices_on(part)
```
kobliha: looks strange: `devices_on(part)` but the method is `devices_on(dev)`

kobliha:
Please add to SUPPORTED_SCENARIOS.md

kobliha:
The code needs refactoring. It's not mandatory now, but definitely nice to have :)

src/modules/BootStorage.rb
kobliha: Additionally, please, move comments for if/else `above` the if/else so they start at the same level (indentation) of code

jreidinger:
updated, refactorings done. code climate see it as good change https://codeclimate.com/github/yast/yast-bootloader/compare/whole_disk_lvm#ratings also rubocop config updated to be more strict again

src/lib/bootloader/stage1.rb
```diff
@@ -196,203 +218,249 @@ def available_partitions(res)
...
-        self.activate = need_activate
-        self.generic_mbr = false
-      else
-        # if not installing to MBR, always activate (so the generic MBR will
-        # boot Linux)
-        self.activate = true
-        self.generic_mbr = true
-      end
+  # Represents object that can set passed stage1 to proposed values
+  class Stage1Proposal
```
mvidner: Please move this nontrivial class to its own file.

jreidinger: On other hand it is specialized class, that should be used only from this class. That is reason why I still keep using #proposal method in Stage1 to hide this one. And it is highly coupled to Stage1 class. It do not make sense to have it without Stage1.

src/lib/bootloader/grub_install.rb
```diff
@@ -64,30 +71,48 @@ def no_device_install?
...
+                  when "ppc", "ppc64"       then ppc_target
+                  when "s390_32", "s390_64" then s390_target
+                  when "aarch64"            then aarch64_target
                   else
                     raise "unsupported architecture '#{Yast::Arch.architecture}'"
                   end
     end
+
+    def i386_target
+      if @efi
```
imobach: np: you could use `@efi ? "i386-efi" : "i386-pc"` which looks more concise. But I don't have a strong opinion on that.

jreidinger: yes, looks better

src/lib/bootloader/grub_install.rb
```diff
@@ -14,11 +14,24 @@ def execute(devices: [], secure_boot: false, trusted_boot: false)
       raise "cannot have secure boot without efi" if secure_boot && !@efi
```
imobach: np: I would prefer having an `attr_reader :efi` or a `efi?` method instead of use the instance variable.

src/lib/bootloader/grub_install.rb
jreidinger: another option I have in mind is to use constants for EFI and NON_EFI targets and do hash lookup.

src/lib/bootloader/stage1.rb
```diff
@@ -38,17 +40,14 @@ def write
 
     def include?(dev)
```
imobach: Missing method documentation

src/lib/bootloader/stage1.rb
```diff
@@ -38,17 +40,14 @@ def write
...
       kernel_dev = Bootloader::UdevMapping.to_kernel_device(dev)
```
imobach: Is this line still needed?

src/lib/bootloader/stage1.rb
```diff
@@ -103,26 +130,13 @@ def custom_devices
...
-      when /ppc/
-        propose_ppc
-      when /s390/
-        propose_s390
-      else
-        raise "unsuported architecture #{Yast::Arch.architecture}"
-      end
-
-      log.info "proposed stage1 configuratopn #{inspect}"
+      Stage1Proposal.propose(self)
```
imobach: Much better :)

src/lib/bootloader/grub_install.rb
jreidinger: done

src/lib/bootloader/stage1.rb
jreidinger: will add it

src/lib/bootloader/stage1.rb
```diff
@@ -196,203 +218,249 @@ def available_partitions(res)
...
+      else
+        raise "Unknown value to select bootloader device #{selected_location.inspect}"
       end
+    end
 
-      same_disk_part = partitions.find do |part|
-        disk = Yast::Storage.GetDiskPartition(part)["disk"]
-        Yast::BootStorage.disk_with_boot_partition == disk
+    # x86_64 specific stage1 proposal
+    class X64 < Stage1Proposal
```
imobach: I'm not against nesting classes (although I prefer a module). But I would prefer them to be in a separated file. If we add new architectures (I know, it's unlikely to happen) this class will keep growing.

src/lib/bootloader/stage1.rb
```diff
@@ -196,203 +218,249 @@ def available_partitions(res)
...
+          # If there is an active partition, then we do not need to activate
+          # one (otherwise we do).
+          # Reason: if we use our own MBR code, we do not rely on the activate
+          # flag in the partition table to boot Linux. Thus, the activated
+          # partition can remain activated, which causes less problems with
+          # other installed OSes like Windows (older versions assign the C:
+          # drive letter to the activated partition).
+          used_disks = ::Bootloader::Stage1Device.new(Yast::BootStorage.mbr_disk).real_devices
+          need_activate = used_disks.any? { |d| Yast::Storage.GetBootPartition(d).empty? }
+          stage1.activate = need_activate
```
imobach: Why not just `stage1.activate = used_disks...`?

src/lib/bootloader/stage1.rb
jreidinger: so you want nested class in own file? so structure like lib/bootloader/stage1_proposal/x64.rb?

src/lib/bootloader/stage1.rb
jreidinger: you are right, relict of past where it was more complex

src/lib/bootloader/stage1_proposal.rb
```diff
@@ -0,0 +1,256 @@
...
+module Bootloader
+  # Represents object that can set passed stage1 to proposed values.
+  # It is highly coupled with Stage1 class and it is recommended to use
+  # {Stage1#proposal} instead of direct usage of this class.
+  class Stage1Proposal
+    include Yast::Logger
+
+    # @param [Bootloader::Stage1] stage1 where write proposal
+    def self.propose(stage1)
+      proposal = case Yast::Arch.architecture
```
imobach: np: If you prefer, you could indent this way:

```ruby
proposal =
  case Yast::Arch.architecture
  when "i386", "x86_64"
     X64
  # and so on
```

src/lib/bootloader/stage1_proposal.rb
```diff
@@ -0,0 +1,256 @@
...
+    # ppc64le specific stage1 proposal
+    class PPC < Stage1Proposal
+      def propose
+        partition = proposed_prep_partition
+        if partition
+          assign_bootloader_device([:custom, partition])
+
+          stage1.activate = true
+          stage1.generic_mbr = false
+        # handle diskless setup, in such case do not write boot code anywhere
```
imobach: Comment indentation looks wrong.

imobach:
The code looks better, thank you :) The only thing I'm missing is documentation on some methods. On the other hand, I would wait for @mvidner's review because I'm not a bootloader expert and sure I've overlooked some things.

src/lib/bootloader/stage1_proposal.rb
jreidinger: actually it is correct as it is comment for below else branch and not for above lines

src/lib/bootloader/stage1_proposal.rb
jreidinger: well, when I look at code from longer distance, it is not so bad, it is very visible that this block of code is together. Another option is to use constant same as in grub_install

mvidner:
LGTM

imobach:
LGTM

PR #342, opened 2016-05-30
Fix stage1

jreidinger:


mvidner:
I'm trying to verify that there was a code coverage hole before which is now covered, but I cannot see it :( Which method was in error? When you're referring to "found by openQA" it would be helpful to paste a link to the failing test.

mvidner:
OK I can see it now, "NoMethodError: undefined method 'separated_boot?' for #<Bootloader::Stage1:0x0000000445c1e0>", it even shows in the rspec run but it is marked as OK with `pending`.

jreidinger:
@mvidner yep, problem with pending, that I cannot specify that pending method expect given failure and not different. openQA failure https://openqa.opensuse.org/tests/203062/modules/yast2_bootloader/steps/4

mvidner:
LGTM

PR #346, opened 2016-06-17
[Review] Request from 'schubi2' @ 'yast/yast-bootloader/review_160617_autoyast_conifuration_bnc_985007'

schubi2:
Please review the following changes:
  * 4f2fdbe packaging
  * f712419 checking if BootStorage has already been initialized

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Jun 17 16:10:23 CEST 2016 - schubi@suse.de
+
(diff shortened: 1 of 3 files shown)


jreidinger:
this is not correct fix, it is just hack. Proper solution is to properly init such variables, as mbr disk should be never empty and also extended partition if set. Also in my test it have very strange configuration like boot from mbr, root and also extended partition, which is bogus. So I prefer to at least init libstorage and use `BootStorage.init_disks`

schubi2:
OK, then lets close this hack.

PR #347, opened 2016-07-01
ruboOptimization

jreidinger:
Result of profiling. After this change run of `yast2 bootloader` which write configuration spend 75% of time in grub2-install, grub2-mkconfig, grub2-set-default and writing configuration. Another 4% used for storage detection and 4% in wait loop ( sorry, was slow :wink: ). Remaining 15% is UI drawing and ruby code itself.

### DIFF
--- a/src/lib/bootloader/sections.rb
+++ b/src/lib/bootloader/sections.rb
@@ -36,7 +36,7 @@ module Bootloader
       log.info "set new default to '#{value.inspect}'"
 
       # empty value mean no default specified
-      raise "Unknown value #{value.inspect}" if !all.include?(value) && !value.empty?
+      raise "Unknown value #{value.inspect}" if !all.empty? && !all.include?(value) && !value.empty?
 
       @default = value
     end
--- a/src/modules/BootStorage.rb
+++ b/src/modules/BootStorage.rb
@@ -176,7 +176,8 @@ module Yast
     def detect_disks
       # The AutoYaST config mode does access to the system.
       # bnc#942360
-      return :ok if Mode.config
+      return if Mode.config
+      return unless @RootPartitionDevice.empty? # quit if already detected
 
       mp = Storage.GetMountPoints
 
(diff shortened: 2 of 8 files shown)


src/lib/bootloader/sections.rb
```diff
@@ -36,7 +36,7 @@ def default=(value)
       log.info "set new default to '#{value.inspect}'"
 
       # empty value mean no default specified
-      raise "Unknown value #{value.inspect}" if !all.include?(value) && !value.empty?
+      raise "Unknown value #{value.inspect}" if !all.empty? && !all.include?(value) && !value.empty?
```
imobachgs: np: In this case, I would apply the De Morgan's law:

`unless all.empty? || all.include?(value) || value.empty?`

But just a matter of preference.

src/modules/BootStorage.rb
```diff
@@ -176,7 +176,8 @@ def bootloader_installable?
     def detect_disks
       # The AutoYaST config mode does access to the system.
       # bnc#942360
-      return :ok if Mode.config
+      return if Mode.config
```
imobachgs: If you're not returning `:ok` anymore, maybe you could write:

```ruby
return if Mode.config || !@RootPartitionDevice.empty?
```

imobachgs:
Just two minor comments. Otherwise, LGTM.

src/lib/bootloader/sections.rb
jreidinger: well, I do not like using unless if it is more then one condition as it is harder to see what it actually is as using unless is same as `if !(all.empty? || all.include?(value) || value.empty?)` which at least for me need some brain processing.

src/modules/BootStorage.rb
jreidinger: well, for me it is better to not merge non-related stuff together. First return is if Mode Config is used and needs to be skipped. The second return is caching.
I think it make sense to put thinks in one condition if it is related like `return if !@RootPartitionDevice && !@RootPartitionDevice.empty?` or `return if !@BootPartitionDevice.empty? || !@RootPartitionDevice.empty?` as both is related to caching.

PR #349, opened 2016-07-10
Add initial support for EFI GRUB on 32-bit ARM

afaerber:
With LOADER_TYPE manually set to "grub2-efi" it was erroring out in
grub_install.rb with "unsupported combination of architecture arm and
enabled EFI".

This lets `yast bootloader` save without errors.

Signed-off-by: Andreas Färber <afaerber@suse.de>

### DIFF
--- a/src/lib/bootloader/grub2efi.rb
+++ b/src/lib/bootloader/grub2efi.rb
@@ -97,6 +97,8 @@ module Bootloader
       when "x86_64"
         res << "grub2-x86_64-efi"
         res << "shim" << "mokutil" if @secure_boot
+      when "arm"
+        res << "grub2-arm-efi"
       when "aarch64"
         res << "grub2-arm64-efi"
       else
--- a/src/lib/bootloader/grub_install.rb
+++ b/src/lib/bootloader/grub_install.rb
@@ -83,6 +83,7 @@ module Bootloader
     EFI_TARGETS = {
       "i386"    => "i386-efi",
       "x86_64"  => "x86_64-efi",
+      "arm"     => "arm-efi",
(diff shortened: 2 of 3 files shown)


jreidinger:
Do we support 32 bit arm? Currently I am aware only about aarch64

afaerber:
https://events.opensuse.org/conference/oSC16/program/proposal/920

jreidinger:
@afaerber I was there and also on @agraf presentation about UEFI there, but one think is announcement on conference and second is official start of supporting it. @jsrain how do you see it?

jsrain:
@jreidinger: I was not at the conference and am not aware of any statement regarding 32-bit ARM support. On the other hand, this patch is harmless for the other cases, does not make the code more unreadable, therefore I don't object aginst including the code if there are intentions for future support.

afaerber:
This is a code contribution fixing actual errors, not a support request. Once there is Arch.arm available I can also more easily patch the suggest parts as well.

jreidinger:
OK, so lets merge it. @afaerber just expect if I get bug report for this part of code, I will contact you :wink: 

Thanks for contribution

afaerber:
@jsrain: The ARMv7 port exists since 2011, much longer than AArch64. What's new is the growing existence of 32-bit UEFI implementations via U-Boot beyond ARM's Tianocore for its own boards. I would've expected a "Thank you for your contribution" and not discussions about whether or not that code needs to be supported...

jreidinger:
@afaerber well, we have some bad experience from past, where we support bazilions of power pc subarchitectures and when bugs appear, we have to deal with them, so we are a bit defensive. It is not anything against you.

PR #351, opened 2016-07-13
Fix default in submenu

jreidinger:


jreidinger:
depends on https://github.com/config-files-api/config_files_api_grub2/pull/12

imobachgs:
LGTM

imobachgs:
Well, is version bump needed?

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jul 13 08:49:50 UTC 2016 - jreidinger@suse.com
+
+- fix writting default boot entry when it is located in grub2
```
imobachgs: writting -> writing

imobachgs:
LGTM

package/yast2-bootloader.spec
```diff
@@ -41,7 +41,7 @@ Requires:       yast2-packager >= 2.17.24
 Requires:       yast2-pkg-bindings >= 2.17.25
 Requires:       yast2-storage >= 2.18.18
 # GRUB_DEFAULT attribute
```
mvidner: remove the comment or update it to # filter out unbootable menuentries

src/lib/bootloader/sections.rb
```diff
@@ -44,7 +42,31 @@ def default=(value)
...
     def write
       return if default.empty?
-      Yast::Execute.on_target("/usr/sbin/grub2-set-default", default)
+
+      Yast::Execute.on_target("/usr/sbin/grub2-set-default", title_to_path(default))
+    end
+
+  private
+
+    def read_default
```
mvidner: I still think you should document methods no matter whether they are private or not.
```
# @return path of the default entry
# @return title of the default entry
```

src/lib/bootloader/sections.rb
```diff
@@ -44,7 +42,31 @@ def default=(value)
...
+      # Execute.on_target can return nil if call failed. It shows users error popup, but bootloader
```
mvidner: Is the comment relevant? Do you mean "it would show a popup for an  internal error nil.lines.grep"?
Do you want to distinguish "program failed" from "output does not contain saved_entry"?

src/lib/bootloader/sections.rb
mvidner: maybe `read_default_path`

src/lib/bootloader/sections.rb
```diff
@@ -22,12 +23,9 @@ def default
...
       return @default = "" if Yast::Stage.initial
 
-      # Execute.on_target can return nil if call failed. It shows users error popup, but bootloader
-      # can continue with empty default section
-      saved = Yast::Execute.on_target("/usr/bin/grub2-editenv", "list", stdout: :capture) || ""
-      saved_line = saved.lines.grep(/saved_entry=/).first
+      default_path = read_default
 
-      @default = saved_line ? saved_line[/saved_entry=(.*)$/, 1] : all.first
+      @default = default_path ? path_to_title(default_path) : all.first
```
mvidner: comment `attr :all` with `# titles of all sections`
to distinguish from `# paths of all sections` ?

mvidner: Will this crash if `all` is empty?

src/lib/bootloader/sections.rb
jreidinger: I mean that Execute.on_target show error popup and return nil. And in such case it use default `""` which is later properly handled as use first available entry.

src/lib/bootloader/sections.rb
jreidinger: well, in such case @default will be nil, which maybe can crash, but as this can be runned only in runtime, I do not expect that it can be empty, otherwise it have problem to boot at all.

PR #352, opened 2016-07-13
Review fixes

jreidinger:


src/lib/bootloader/sections.rb
```diff
@@ -48,7 +48,9 @@ def write
 
   private
 
-    def read_default
+    # @return [String, nil] return default boot path as string or nil if not set or something goes wrong
+    # @note shows error popup if calling grub2-editenv failed
```
mvidner: ah, this note should in fact be added also to the docs of `on_target` and `locally` https://github.com/yast/yast-yast2/blob/master/library/system/src/lib/yast2/execute.rb#L41

src/lib/bootloader/sysconfig.rb
```diff
@@ -120,14 +120,12 @@ def ensure_file_exists_in_target
...
-        File.join(destdir, "/etc/sysconfig")
-                       )
+        File.join(destdir, "/etc/sysconfig"))
       Yast::WFM.Execute(Yast::Path.new(".local.bash"),
-        "touch #{destdir}/etc/sysconfig/bootloader"
-                       )
+        "touch #{destdir}/etc/sysconfig/bootloader")
     end
 
-    def temporary_target_agent(&block)
```
lslezak: Well, this is is a bit controversial change. Rubocop says that `yield` is faster, but for me the `&block` style is better readable.

You can see just from the method header that it takes a block. After the change this is hidden in the method body  and without any documentation comment it's even worse... :worried: 

Here I'd prefer a different default for YaST. What do you think about it?

src/modules/BootSupportCheck.rb
```diff
@@ -185,12 +185,12 @@ def check_zipl_part
       boot_part = Storage.GetEntryForMountpoint("/boot") if boot_part.empty?
       boot_part = Storage.GetEntryForMountpoint("/") if boot_part.empty?
 
-      if [:ext2, :ext3, :ext4].include? boot_part["used_fs"]
-        return true
-      else
+      if ![:ext2, :ext3, :ext4].include? boot_part["used_fs"]
```
lslezak: NP: Here I'd fix it the other way (less indentation, no negation in the condition):

```ruby
return true if [:ext2, :ext3, :ext4].include? boot_part["used_fs"]

add_new_problem(_("Missing ext partition for booting. Cannot install boot code."))
false
```

mvidner:
`#default=` does check `all.empty` but `#default` does not, we should unify it.

test/grub2_widgets_test.rb
```diff
@@ -623,9 +623,9 @@ def stub_widget_value(id, value)
 
   it "enlists all available sections" do
     expect(subject.items).to eq([
```
lslezak: I think if you move the opening `[` on a separate line then Rubocop does not indent that much...

test/grub2pwd_test.rb
```diff
@@ -17,22 +17,22 @@ def mock_file_presence(exists)
       .and_return(exists)
   end
 
-  ENCRYPTED_PASSWORD = "grub.pbkdf2.sha512.10000.774E325959D6D7BCFB7384A0245674D83D0D540A89C02FEA81E35489F8DE7ADFD93988190AD9857A0FFF363825DDF97C8F4E658D8CC49FC4A22C053B08AB3EFE.6FB19FF26FD03D85C40A33D8BA7C04E72EDE3DD5D7080C177553A4FED370F71C579AF0B15B3B93ECECEA355469A4B6D0560BFB53ED35DDA0B80F5363BFBD54E4"
+  ENCRYPTED_PASSWORD = "grub.pbkdf2.sha512.10000.774E325959D6D7BCFB7384A0245674D83D0D540A89C02FEA81E35489F8DE7ADFD93988190AD9857A0FFF363825DDF97C8F4E658D8CC49FC4A22C053B08AB3EFE.6FB19FF26FD03D85C40A33D8BA7C04E72EDE3DD5D7080C177553A4FED370F71C579AF0B15B3B93ECECEA355469A4B6D0560BFB53ED35DDA0B80F5363BFBD54E4".freeze
```
lslezak: Ha! What a nice constant! :wink:

src/lib/bootloader/sections.rb
jreidinger: ah, you are right, there was old documention which mention that it raises Cheetah::ExecutionFailed, which is no longer true

src/lib/bootloader/sysconfig.rb
jreidinger: I agree, I think it is not so big slowdown and readability is really better

mvidner: I think the speedup is negligible, I mean we are not even iterating over a container. Scrap this kind of speedups.

lslezak:
Beside that block parameter change, which we need to  agree on, LGTM.

src/modules/BootSupportCheck.rb
jreidinger: I also considering it, but on other hand, for me it is more natural work-flow in check method, to verify that no problem happend and if nothing found, then return true.

test/grub2pwd_test.rb
jreidinger: real grub output for test :)

test/grub2_widgets_test.rb
jreidinger: it is right

jreidinger:
> #default= does check all.empty but #default does not, we should unify it.

@mvidner it have different consequences, if in `#default=` it is empty, then it will raise exception, which is not case for `#default` but fine for me to unify.

.rubocop.yml
```diff
@@ -120,3 +120,11 @@ Style/MethodName:
...
   Include:
     - 'src/lib/**/*.rb' # force known variables only for new code
+
+# alias method is safer option
+Style/Alias:
+  EnforcedStyle: prefer_alias_method
+
+# performance gain is not so big and readability is better
+Performance/RedundantBlockCall:
+  Enabled: false
```
lslezak: I'd like to add this globally, see https://github.com/yast/yast-devtools/pull/107/files#diff-71e6a5f24eae84627de24615bb8d6545R218

So maybe we should discuss and merge https://github.com/yast/yast-devtools/pull/107 first...

src/lib/bootloader/sections.rb
```diff
@@ -17,18 +19,20 @@ def initialize(grub_cfg = nil)
       @all = @data.map { |e| e[:title] }
     end
 
-    # @return [String] name of default section
+    # @return [String] name of default boot title. It is not full path,
```
mvidner: "name of default boot title"? what about "title of default boot section"?

jreidinger: ok, sounds better

mvidner:
And now RuboCop finds new formatting problems :)

jreidinger:
@mvidner it is caused by enabled one of cop, that was not enabled before. fix is now pushed.

lslezak:
LGTM

PR #354, opened 2016-07-18
Fix tests

jreidinger:


test/grub2base_test.rb
```diff
@@ -225,7 +261,7 @@
 
         context "on other TERM" do
           before do
-            ENV["TERM"] = "xterm"
+            allow(ENV).to receive(:"[]").with("TERM").and_return("xterm")
```
mvidner: I was surprised to find out that `:[]` is valid syntax. Not sure what RuboCop likes.

mvidner: No complaints, good.

mvidner:
LGTM.

PR #357, opened 2016-08-26
warn user if enabling TPM when not available (bsc#994556)

jsrain:


lslezak:
LGTM, the question is how much difficult would be to add an unit test for this change?

jsrain:
Trusted boot widget has (unlike some others) no test coverage at the moment :-/

mvidner:
LGTM
(Jiri says he has tested it manually)

PR #358, opened 2016-08-30
Fix activating mdrad

jreidinger:
for initial review. I plan to do some manual testing if everything is fine.
Test case included.

### DIFF
--- a/src/lib/bootloader/mbr_update.rb
+++ b/src/lib/bootloader/mbr_update.rb
@@ -139,6 +139,10 @@ module Bootloader
       ret = [mbr_disk]
       # Add to disks only if part of raid on base devices lives on mbr_disk
       ret.concat(mbrs) if mbrs.include?(mbr_disk)
+      # get only real disks
+      ret = ret.each_with_object([]) do |disk, res|
+        res.concat(::Bootloader::Stage1Device.new(disk).real_devices)
+      end
 
       ret.uniq
     end
--- a/src/lib/bootloader/mbr_update.rb
+++ b/src/lib/bootloader/mbr_update.rb
@@ -1,6 +1,7 @@
 require "yast"
 
 require "bootloader/boot_record_backup"
+require "bootloader/stage1_device"
 require "yast2/execute"
 
 Yast.import "Arch"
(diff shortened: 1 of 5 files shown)


src/lib/bootloader/mbr_update.rb
```diff
@@ -181,20 +198,14 @@ def extended_partition_num(disk)
     #  containing device (eg. "/dev/hda4"), disk (eg. "/dev/hda") and
     #  partition number (eg. 4)
```
mvidner: Since we're fixing this method, let's make sure the API docs are OK:
```rb
# @return a Hash { "mbr" => String, "num" => Integer }
# containing device (eg. "/dev/hda4"), disk (eg. "/dev/hda") and
#  partition number (eg. 4)
```
But I do not understand the "device" part. Like, a *master* BR can be in the BR of a *partition*? (/dev/hda4)

Also, the method sometimes returns {}. Does that mean "no partition should be activated"?

src/lib/bootloader/mbr_update.rb
```diff
@@ -142,20 +143,36 @@ def disks_to_rewrite
...
+      lowest_id = MAX_BIOS_ID
+      md.real_devices.reduce do |res, device|
+        res ||= device # ensure at least one device is used, even if none have bios_id
+        bios_id_num = bios_id_for(device)
+        if lowest_id > bios_id_num
+          lowest_id = bios_id_num
+          next device
+        else
+          res
+        end
```
mvidner: Could this be written more clearly as `md.real_devices.sort_by {|d| bios_id_for(d)}.first` ?

jreidinger: yes, can be

src/lib/bootloader/mbr_update.rb
jreidinger: empty one can be returned only for very strange storage setup like only swap partitions on disk or ones that cannot be activated, so empty one means "activate nothing".

I do not get your question. It return device for mbr ( disk device ) and number of its partition (giving together PBR - partition boot record ) which is needed for parted, as parted need disk and partition number.

src/lib/bootloader/mbr_update.rb
mvidner: Ah, `min_by` is even better http://ruby-doc.org/core-2.2.0/Enumerable.html#method-i-min_by

mvidner:
LGTM

jreidinger:
thanks, will do manual testing and then merge if everything go right

mvidner:
LGTM

jreidinger:
manual testing looks good so far ( 1 commit added to fix another issue found during testing ). I will just test also lvm installation.

jreidinger:
all tests works fine, so merging

PR #359, opened 2016-08-30
Import device map

jreidinger:


src/lib/bootloader/device_map.rb
```diff
@@ -33,6 +33,7 @@ def read
     end
 
     def write
+      log.info "writting device map: #{to_s}"
```
mvidner: only one T: writing

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Tue Aug 30 13:31:14 UTC 2016 - jreidinger@suse.com
+
+- import properly device map in autoyast profile (found during
+  debugging bnc#995627)
+- log device map entries before written down to allow easier
```
mvidner: > log device map entries before writing them, to allow easier

mvidner:
Please also add device map exporting to AutoyastConverter.export, or add a comment in there explaining why it is not needed.

mvidner:
LGTM

PR #360, opened 2016-09-30
Boot storage requirements

jreidinger:


doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,38 @@
+List of Information Needed About Storage Configuration
+======================================================
+
+Mount Points
+---------------
+
+- what partition is mounted to "/" ( so it can be queried )
+- what partition is mounted to "/boot" and if it is same as "/" ( so it can be queried )
+- what partition is mounted to "/boot/zipl" ( only on s390 to query )
+- what partition is mounted to "/boot/efi" ( only on efi based archs )
```
ancorgs: All this can be easily queried in several ways. For example:
```
devicegraph.filesystems.with(mountpoint: "/").first
devicegraph.filesystems.with(mountpoint: "/boot").first
```
and so on.

jreidinger: well, for me it is what I need to know, not how :) So it is up to storage guys if there provide method like `root_partition` or `GetMountPoints` ( like now) or `mountpoint: "/"`. What is important for me is to recognize that two mount point points to same device.

doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,38 @@
...
+Partitions
+---------------
+
+- is partition logical ( so MBR have to be installed to extended one )
+- extended partition for given disk ( needed only if /boot is on logical partition )
+- file system of partition ( as some fs have problem embedding stage1 code, so do not propose to install it there )
+- fsid of partition ( e.g. to detect bios_grub partition and others )
+- if partition is part of MD, DM or Linux RAID ( target map now returns it, needed to decide if partition can be used for stage1 code )
+- raid level ( needed to check if it is 1, where we support redundancy boot from )
+- if partition is freshly created or deleted ( well, for deleted is to not count partitions that will gone in installation, for creation it is used for prep partition chooser, as newly created prep is prefered as it will not overwrite existing code )
```
ancorgs: We will have the probed (original) devicegraph and the committed one. Knowing if a partition is new (added by us) can be done comparing both.

doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,38 @@
...
+- enlist all swap partitions with its size, and if swap is encrypted, then also detect it and use proper device ( allow user to choose any, but propose the biggest one )
+- for encrypted devices provide info it is encrypted and also both devices name ( needed to use proper devices names, needed mainly for encrypted /boot and swap as mentioned above )
+- is partition mounted by nfs ( when partition is on nfs, we have to skip writting bootloader code )
+- convert raid with alternative name to standard name ( see bnc#944041 )
+
+Disks
+-----
+
+- label of disk ( e.g. to install proper generic code or to respect some msdos PT limitation )
+- where virtual disks like LVM, RAID ones, etc. layes on physical devices ( in some cases it have to use physical devices to write boot code )
```
ancorgs: Is this line intentionally repeated here and in "partitions"?

doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,38 @@
...
+- enlist all partitions that are prep ones ( prep or gpt_prep ) ( needed to decide where to install stage1 on powerpc )
+- enlist all swap partitions with its size, and if swap is encrypted, then also detect it and use proper device ( allow user to choose any, but propose the biggest one )
+- for encrypted devices provide info it is encrypted and also both devices name ( needed to use proper devices names, needed mainly for encrypted /boot and swap as mentioned above )
+- is partition mounted by nfs ( when partition is on nfs, we have to skip writting bootloader code )
+- convert raid with alternative name to standard name ( see bnc#944041 )
+
+Disks
+-----
+
+- label of disk ( e.g. to install proper generic code or to respect some msdos PT limitation )
```
ancorgs: Not sure what you mean with "label of disk".

doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,38 @@
...
+- if partition is part of MD, DM or Linux RAID ( target map now returns it, needed to decide if partition can be used for stage1 code )
+- raid level ( needed to check if it is 1, where we support redundancy boot from )
+- if partition is freshly created or deleted ( well, for deleted is to not count partitions that will gone in installation, for creation it is used for prep partition chooser, as newly created prep is prefered as it will not overwrite existing code )
+- if partition have boot flag ( bootloader ensure at least one partition have it and if requested, then also mark partition as one for boot )
+- marking partition with boot flag ( currently done directly via parted, but in future it would be nice to have it in one location, marking partition means also ensuring that no other partition have it )
+- where virtual partitions like EVMS, LVM, RAID ones, etc. layes on physical devices ( in some cases it have to use physical devices to write boot code )
+- convert udev device to kernel name and also kernel name to udev name according to settings in storage ( mountby option ) ( needed for persistant storage names feature )
```
ancorgs: Pasting comment by @aschnell  in IRC:

> where do you get "alternative name" and "udev device" from? e.g. do the devices already exist in the system or might they be new?
> e.g. would you ask for the kernel name of /dev/disk/by-id/sdnfio-part8 if the partition does not exist on disk yet?

doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,38 @@
...
+- what partition is mounted to "/boot/efi" ( only on efi based archs )
+
+Partitions
+---------------
+
+- is partition logical ( so MBR have to be installed to extended one )
+- extended partition for given disk ( needed only if /boot is on logical partition )
+- file system of partition ( as some fs have problem embedding stage1 code, so do not propose to install it there )
+- fsid of partition ( e.g. to detect bios_grub partition and others )
+- if partition is part of MD, DM or Linux RAID ( target map now returns it, needed to decide if partition can be used for stage1 code )
```
ancorgs: Pasting comment from @aschnell  in IRC

> What is linux raid if not MD?

doc/boot_storage_needed_info.md
jreidinger: yes, as I to know physical part for both for different cases ( so virtual partition and also virtual disks )

doc/boot_storage_needed_info.md
jreidinger: my answer is still same. I do not know, but we query it in ancient code - https://github.com/yast/yast-bootloader/blob/master/src/modules/BootStorage.rb#L93 so it is question more for @aschnell what is difference between "Linux RAID" and "MD RAID" partition fstype

doc/boot_storage_needed_info.md
jreidinger: for udev, I usually get it from storage. Regarding alternative name, it was used by autoyast profile. That I will need check where it is actually translated. But I know that e.g. for uuid translation is done in bootloader_finish phase, when all mapping is available.

ancorgs:
The requirements look clear enough for me (that was the scope of the PR). But please wait for a final LGTM from @aschnell  since he knows the domain much better and can come up with better questions.

doc/boot_storage_needed_info.md
jsrain: AFAIK we support the BIOS RAID (the one you set up in BIOS, but hardware still shows individual disks), and such RAID is handled via device mapper (Arvin, correct me if I'm wrong)

aschnell:
BIOS RAIDs can also be handled by mdadm. It depends on the type on the BIOS RAID.

aschnell:
In general it looks OK although some requirements are vague. It should be possible to sort things out during implementation.

But I want the question from Steffen about the need of the BIOS ID clarified.

doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,38 @@
...
+- is partition mounted by nfs ( when partition is on nfs, we have to skip writting bootloader code )
+- convert raid with alternative name to standard name ( see bnc#944041 )
+
+Disks
+-----
+
+- label of disk ( e.g. to install proper generic code or to respect some msdos PT limitation )
+- where virtual disks like LVM, RAID ones, etc. layes on physical devices ( in some cases it have to use physical devices to write boot code )
+- convert udev device to kernel name and also kernel name to udev name according to settings in storage ( mountby option )
+- BIOS boot order ( needed mainly for non-EFI x86_64 to enlist disks in device map )
```
jreidinger: code that use it now is in https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/mbr_update.rb#L152

doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,38 @@
+List of Information Needed About Storage Configuration
+======================================================
+
+Mount Points
+---------------
+
+- what partition is mounted to "/" ( so it can be queried )
```
jreidinger: ok, I will replace it with generic device as I know that it can be whole disk. On other hand bootloader is not much interested in filesystem, but more what kind of device it is.

doc/boot_storage_needed_info.md
jreidinger: I agree that code in this part is quite unclear as it checks both, but when I check storage code, it cannot ever happen. So I will replace requirement by generic term, that we need info about any kind of raid.

doc/boot_storage_needed_info.md
jreidinger: OK, I add notes there about specific requirements.

doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,38 @@
...
+- marking partition with boot flag ( currently done directly via parted, but in future it would be nice to have it in one location, marking partition means also ensuring that no other partition have it )
+- where virtual partitions like EVMS, LVM, RAID ones, etc. layes on physical devices ( in some cases it have to use physical devices to write boot code )
+- convert udev device to kernel name and also kernel name to udev name according to settings in storage ( mountby option ) ( needed for persistant storage names feature )
+- enlist all partitions that are prep ones ( prep or gpt_prep ) ( needed to decide where to install stage1 on powerpc )
+- enlist all swap partitions with its size, and if swap is encrypted, then also detect it and use proper device ( allow user to choose any, but propose the biggest one )
+- for encrypted devices provide info it is encrypted and also both devices name ( needed to use proper devices names, needed mainly for encrypted /boot and swap as mentioned above )
+- is partition mounted by nfs ( when partition is on nfs, we have to skip writting bootloader code )
```
jreidinger: fixed

aschnell:
What about the legacy_boot flag on GPT?

doc/boot_storage_needed_info.md
jreidinger: we use it also to decide where write generic_mbr and I think that `dd if=.. of=/dev/dm-0` probably won't work. I will add note that if we change implementation, this requirement no longer is needed.

jreidinger:
@aschnell right, we also need it for legacy boot on GPT. I will add it.

doc/boot_storage_needed_info.md
```diff
@@ -0,0 +1,44 @@
...
+
+- is partition logical ( so MBR have to be installed to extended one )
+- extended partition for given disk ( needed only if /boot is on logical partition )
+- file system of partition ( as some fs have problem embedding stage1 code, so do not propose to install it there )
+- fsid of partition ( e.g. to detect `bios_grub` partition and others )
+- if partition is part of any kind of RAID ( target map now returns it, needed to decide if partition can be used for stage1 code )
+- raid level ( needed to check if it is 1, where we support redundancy boot from )
+- if partition is freshly created or deleted ( well, for deleted is to not count partitions that will gone in installation, for creation it is used for prep partition chooser, as newly created prep is prefered as it will not overwrite existing code )
+- if partition have boot flag or legacy_boot flag( bootloader ensure at least one partition have it and if requested, then also mark partition as one for boot )
+- marking partition with boot flag ( currently done directly via parted, but in future it would be nice to have it in one location, marking partition means also ensuring that no other partition have it )
```
aschnell: I assume legacy_boot flag is missing here.

jreidinger: yes, also here. I will add also there.

doc/boot_storage_needed_info.md
wfeldt: > we use it also to decide where write generic_mbr

Yes, we did. But if the new strategy is to modify only the device with boot/root this is no longer needed. You would put the mbr on the system disk.

I think we should talk with mchang here to be sure we're doing the right thing.

jreidinger: @wfeld I mean if /boot is on dm device, then dd to this device do not work, so we need to write to physical one. There we use one with lower bios ID, but in the end we can write it to all disk where lives this mirrored partition, so as I write to document if we change this implementation, we no longer needed it.

wfeldt: @jreidinger maybe deserves some tests.

jreidinger: @wfeldt yes, I think it can be part of adaptation to new storage, as it will also need code adaptation and lot of testing.

PR #361, opened 2016-09-30
Unsupported popup

jreidinger:


src/modules/Bootloader.rb
```diff
@@ -150,7 +152,21 @@ def Read
       Progress.NextStage
       return false if testAbort
 
-      ::Bootloader::BootloaderFactory.current.read
+      begin
+        ::Bootloader::BootloaderFactory.current.read
+      rescue ::Bootloader::UnsupportedBootloader => e
+        ret = Yast::Report.AnyQuestion(_("Unsupported Bootloader"),
+          _("System uses unsupported bootloader '%s'. Propose to use supported bootloader?") %
```
ancorgs: It would be "The system". But moreover, I think the phrasing is not crystal clear for the average user. Maybe something like: "Unsupported bootloader '%s' detected. Switch to a supported one?". Even better, since I assume we know the name of the proposed booloader: "Switch to '%s'?"

jreidinger: well, we do not know it in advance, that I would have to computate and still user can change it. It is mainly to open bootloader module with something reasonable and do not crash.

src/modules/Bootloader.rb
```diff
@@ -150,7 +152,20 @@ def Read
...
       return false if testAbort
 
-      ::Bootloader::BootloaderFactory.current.read
+      begin
+        ::Bootloader::BootloaderFactory.current.read
+      rescue ::Bootloader::UnsupportedBootloader => e
+        ret = Yast::Report.AnyQuestion(_("Unsupported Bootloader"),
+          _("Unsupported bootloader '%s' detected. Switch to a supported one?") %
+            e.bootloader_name,
+          _("Propose"),
```
ancorgs: Now this should be "Switch" or something similar, isn't it?

jreidinger: make sense

ancorgs:
LGTM

PR #362, opened 2016-10-04
Correct version number

ancorgs:
- Master should start using 3.2.x
 - 3.1.203.x should be reserved for critical fixes in SP2 before the final release.
 - 3.1.2XX shoud be reserved for fixes for Leap 42.2 before the release (also released as maint. update for SP2)

I need to correct this in order to be able to create the new branch and submit there the fix to https://bugzilla.suse.com/show_bug.cgi?id=1000629


jreidinger:
LGTM

PR #363, opened 2016-10-04
Rakefile. For the time being, submit fixes to Leap42.2

ancorgs:


jreidinger:
OK, LGTM, we just have to not forget to remove it after merge.

ancorgs:
`rake osc:sr` doesn't work. AFAICS that happens because the version of `rubygem-yast-rake` in Leap 42.2 doesn't contain the `leap_42_2` target. There are two possible solutions
 
1) Change this rakefile to use `obs_project`, `obs_sr_projects`, etc. directly instead of a target
2) Submit an up-to-date version of `rubygem-yast-rake` to Leap 42.2

I prefer 2 (since, as a bonus track, Leap 42.2 would be a better development environment for YaST). As a first step, I created this two submit request (I believe we should submit to Leap from YaST:openSUSE:42.2, not directly from YaST:Head).
https://build.opensuse.org/request/show/433207
https://build.opensuse.org/request/show/433206

@jreidinger (or any other, for the matter), if you agree please accept the submit requests (auto-acceptance is kind of ugly :wink:) and I will submit from there to Leap42.2 tomorrow morning (and will ping Ludwig to speed up acceptance). That should unblock the situation.

jreidinger:
@ancorgs accepted, please also submit rubygem-libyui-rake to Leap, so it will be consistent.

jreidinger:
@ancorgs and BTW I suggest to use always the latest package from YaST:Head or from rubygems.org

PR #366, opened 2016-10-06
Restore broken config

jreidinger:


src/lib/bootloader/exceptions.rb
```diff
@@ -7,4 +9,17 @@ def initialize(bootloader_name)
...
+  class BrokenConfiguration < RuntimeError
+    include Yast::I18n
+    attr_reader :reason
+
+    def initialize(msg)
+      @reason = msg
+      textdomain "bootloader"
+
+      super _("Found problem during reading bootloader configuration files. " \
+        "Please open bootloader module and fix it. Details: %s") % msg
```
ancorgs: That "during reading" sounds weird. What about "Error reading the bootloader configuration files"?

In the second sentence the meaning of "open bootloader module" could be obvious for us, but not so sure for all users. Maybe, "Please use the bootloader YaST module to fix it", or "...the bootloader module of YaST...", or "the YaST bootloader module..." or simply  "Please use YaST2 bootloader to fix it".

src/lib/bootloader/udev_mapping.rb
```diff
@@ -24,13 +27,21 @@ class << self
...
     def to_kernel_device(dev)
+      textdomain "bootloader"
       log.info "call to_kernel_device for #{dev}"
       raise "invalid device nil" unless dev
 
       # for non-udev devices try to see specific raid names (bnc#944041)
       if dev =~ /^\/dev\/disk\/by-/
-        # in mode config if not found, then return itself
-        all_devices[dev] or Yast::Mode.config ? dev : raise("Unknown udev device #{dev}")
+        # TRANSLATORS: error message, %s stands for problematic device.
```
ancorgs: This comment is missplaced

src/modules/Bootloader.rb
```diff
@@ -156,9 +156,20 @@ def Read
...
+          _("Use"),
+          _("Quit"),
+          :yes) # focus proposing new one
+        return false unless ret
+
+        ::Bootloader::BootloaderFactory.current = ::Bootloader::BootloaderFactory.proposed
+        ::Bootloader::BootloaderFactory.current.propose
+      rescue ::Bootloader::BrokenConfiguration => e
+        ret = Yast::Report.AnyQuestion(_("Broken Configuration"),
+          _("Bootloader configuration is broken (%s). Propose configuration from scratch?") %
```
ancorgs: Maybe we should be less assertive about current configuration. What about "YaST cannot process current bootloader configuration (%s). Propose new configuration from scratch?". Even if we can't parse it/understand it, maybe is not broken strictly speaking.

src/modules/Bootloader.rb
```diff
@@ -156,9 +156,20 @@ def Read
...
             e.bootloader_name,
-          _("Show"),
+          _("Use"),
+          _("Quit"),
+          :yes) # focus proposing new one
+        return false unless ret
+
+        ::Bootloader::BootloaderFactory.current = ::Bootloader::BootloaderFactory.proposed
+        ::Bootloader::BootloaderFactory.current.propose
+      rescue ::Bootloader::BrokenConfiguration => e
```
ancorgs: I don't have a full overview on how the pieces interact, but having this four lines here and also in bootloader/config_dialog.rb looks suspicious.

src/lib/bootloader/udev_mapping.rb
ancorgs: Is not misplaced anymore after the rubocop commit

src/modules/Bootloader.rb
jreidinger: it depends if configuration information is needed during read or later when trying to understand it ( like e.g. udev can be broken when read in various places, but if it is in stage1 location, then it is read as value, but translated from udev later when we try to interpret if it is e.g. MBR device or not ).

jreidinger: so catch is in two phases. The first in Read and second during showing configuration dialog.

ancorgs:
LGTM

Mirppc:
Just ran into a similar situation yesterday with a fresh install of 42.2 on my workstation after a drive exploded.

here is a screenshot.

http://i.imgur.com/wi1MhkA.png

jreidinger:
@Mirppc yes, it do not catch 42.2 deadline, so this improvement will be in 42.3.

Mirppc:
This is a big issue, how can i resolve it for 42.2?

jreidinger:
@Mirppc for 42.2 you need to fix you configuration. usually main problem is /etc/fstab and /etc/default/grub_installdevice and /boot/grub2/device.map where is invalid id specified.

Mirppc:
thanks for the response, i will look into everything except fstab as that already has been reviewed and cleared.  I also checked /etc/default/grub.conf.  I was more wondering if there is a repository i could add that would fix this.

jreidinger:
I worry it can be a bit tricky due to dependencies, but YaST:Head repository contain the latest YaST modules, including bootloader with fix.

Mirppc:
It seems to still spout the same error with the Yast:HEAD repo installed.  ;;;>_> Just wanted to give you a heads up, looks like i will need to dig into dealing with the bootloader device config file thingy.  I pray to the glorious green chameleon that i dont break anything @_@

PR #367, opened 2016-10-07
Ensure correct packages

jreidinger:
Users sometimes reported, then when they try to tune their package selection during installation sometimes happen, that then they cannot boot. Reason is that bootloader suggest package proposal what packages he need, but user can overwrite it. To prevent this situation and to increase awareness when such situation happen, bootloader now have feature, that when it is found that required package missing, it reports error and named it. It shoud be obvious from screenshot.
screenshot_2016-10-07_16-01-26

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Fri Oct  7 12:46:18 UTC 2016 - jreidinger@suse.com
 
-- Warn user if package needed for booting is deselected
-  (bsc#885496) 
+- Warn user if the packages needed for booting are deselected
+  (bsc#885496)
 - 3.2.3
 
 -------------------------------------------------------------------
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Oct  7 12:46:18 UTC 2016 - jreidinger@suse.com
+
+- Warn user if package needed for booting is deselected
+  (bsc#885496) 
+- 3.2.3
+
+-------------------------------------------------------------------
 Thu Oct  6 08:24:51 UTC 2016 - jreidinger@suse.com
 
 - allow user to repropose configuration if unknown udev link found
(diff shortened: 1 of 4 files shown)


lslezak:
The `make rubocop happy` commit again shows that [overcommit](http://blog.ladslezak.cz/2016/06/06/overcommit/) really would make sense for us...

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Oct  7 12:46:18 UTC 2016 - jreidinger@suse.com
+
+- Warn user if package needed for booting is deselected
```
lslezak: **the** package, or **the** package**s**...are... .

jreidinger:
@lslezak yeah, my memory is bad, I should find time to check it and configure it for me

lslezak:
Actually we should do it globally in my opnion.

BTW I have a script to install overcommit and create the default config, works recursively for complete Git checkout: https://github.com/lslezak/scripts/blob/master/git/install_overcommit.rb

src/lib/bootloader/proposal_client.rb
```diff
@@ -172,28 +173,48 @@ def construct_proposal_map
...
       end
 
-      ret
+      pkgs = current_bl.packages.map { |p| [p, Yast::Pkg.ResolvableProperties(p, :package, "")] }
+      log.info "packages info #{pkgs.inspect}"
+      pkgs.select! { |_n, p| unselected?(p) }
+      return if pkgs.empty?
+
+      ret["warning_level"] = :error
+      ret["warning"] = _("Packages required for booting is deselected (%s). " \
```
lslezak: If you look at the screenshot the wording seems a bit weird. "Packages required ... (grub2)" - there is  *Packages* but there is only *one* package showed.

Better use the plural form:
```ruby
# TRANSLATORS: %s are package names
n_("A package required for booting is deselected (%s). Please select it for installation again.",
  "Packages required for booting are deselected (%s). Please select them for installation again.",
  pkgs.size) % pkgs.map(&:first).join(", ")
```

This will look much better.

PR #368, opened 2016-10-13
Do not require syslinux on target 42.2

jreidinger:


ancorgs:
So you are always using the mbr.bin from the inst-sys now, isn't it? I mean, not doing what was suggested in the bug report (using the inst-sys option as a fallback).

jreidinger:
@ancorgs no, as it increase number of possible ways that need to be tested and can break. And if we have network and preferred way is using installer update, then it will work. I would like to keep it simple.

ancorgs:
ok. LGTM... if travis passes.

package/yast2-bootloader.spec
```diff
@@ -46,10 +46,6 @@ Requires:       rubygem(%rb_default_ruby_abi:cfa_grub2) >= 0.5.1
 # lenses are needed here
 Requires:       augeas-lenses
 
-%ifarch %ix86 x86_64
-Requires:       syslinux
-%endif
-
```
lnussel: I don't think the bug report requested this change

jreidinger: yes, but it is consistent with other yast modules optional dependencies install only when needed.

PR #371, opened 2016-10-27
fixed build with storage-ng

aschnell:
Yes, this is primitive but a first step to setup autosubmission.

For https://trello.com/c/NQ524dss/459-8-storageng-bootloader-proposal-efi.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,6 @@ script:
     - rake check:pot
     - rubocop
     - yardoc
-    - COVERAGE=1 rake test:unit
+    # - COVERAGE=1 rake test:unit
     - sudo rake install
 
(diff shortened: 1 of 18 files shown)


jreidinger:
fine for me if it fixes build for you. Maybe you should also modify somehow Rakefile to avoid accidental send to factory? Of course to make it work, it will need more changes to support at least minimal case.

aschnell:
Working on it. Jenkins setup will be next.

ancorgs:
@aschnell could you add a commit commenting this line in .travis.yml?
```
 - COVERAGE=1 rake test:unit
```
That should turn travis green

ancorgs:
Shouldn't we also add a changelog entry and bump version to 3.3.0?
We did it for the other branched packages to make everything more clear.

I'm fine with doing it in a separate pull request. If that's the case, LGTM.

ancorgs:
Thanks, LGTM

PR #372, opened 2016-10-28
- make simple EFI proposal work with storage-ng

aschnell:
For https://trello.com/c/NQ524dss/459-8-storageng-bootloader-proposal-efi.

### DIFF
--- a/src/modules/BootStorage.rb
+++ b/src/modules/BootStorage.rb
@@ -50,6 +50,10 @@ module Yast
       @md_physical_disks = []
     end
 
+    def staging
+      Y2Storage::StorageManager.instance.staging
+    end
+
     def gpt_boot_disk?
       require "bootloader/bootloader_factory"
       current_bl = ::Bootloader::BootloaderFactory.current
@@ -103,8 +107,6 @@ module Yast
 
     # Get extended partition for given partition or disk
     def extended_partition_for(device)
-      staging = Y2Storage::StorageManager.instance.staging
-
       blk_device = Storage::BlkDevice.find_by_name(staging, device)
       if Storage.partition?(blk_device)
         partition = Storage.to_partition(blk_device)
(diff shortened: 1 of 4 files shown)


src/modules/BootStorage.rb
```diff
@@ -101,29 +103,15 @@ def possible_locations_for_stage1
...
-        mp,
-        ["/boot", 2],
-        Ops.get_string(mp, ["/", 2], "")
-      )
-      log.info "Disk with boot partition: #{boot_disk}, using for MBR"
+      blk_device = Storage::BlkDevice.find_by_name(staging, device)
+      if Storage.partition?(blk_device)
+        partition = Storage.to_partition(blk_device)
+        return partition.parent.name if partition.type == Storage::PartitionType_LOGICAL
+      end
```
ancorgs: Just pointing a possible alternative using `Refinements::DevicesList`. No offense meant :wink:

```ruby
     def extended_partition_for(device)
        staging = Y2Storage::StorageManager.instance.staging
        logical_partition = staging.partitions.with(name: device).with(type: Storage::PartitionType_LOGICAL).first
        return nil unless logical_partition
        logical_partition.parent.name
    end
```

I don't have a strong preference. The advantage of "my" alternative is that you don't need casting, something that looks weird in Ruby.

src/modules/BootStorage.rb
```diff
@@ -167,10 +158,25 @@ def bootloader_installable?
...
 
       true
     end
 
+    # Find the blkdevice for the filesystem mounted at mountpoint. Returns nil
+    # if no filesystem is found or the filesystem has no blkdevice (e.g. NFS).
+    # TODO: function also defined in grub2efi.rb but simply require grub2efi
+    # does not work and debugging is impossible due to bsc#932331.
+    def find_blk_device_at_mountpoint(mountpoint)
+      staging = Y2Storage::StorageManager.instance.staging
```
ancorgs: We have this line in several places. What about?
```ruby
def staging
   Y2Storage::StorageManager.instance.staging
end
```

ancorgs:
LGTM

PR #374, opened 2016-11-01
Use proper locale

jreidinger:


package/yast2-bootloader.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Nov  1 11:49:05 UTC 2016 - jreidinger@suse.com
+
+- start using system locale when generating grub menu (bsc#998152)
```
mvidner: "Start using foo" is superfluous. "Use foo" is enough.

mvidner: When I read "Using the system locale" I understand that it is using `LANG` (as opposed to, say always English). But in fact the fix makes it use `RC_LANG`. So I'd make it clear with

> Use the system-wide locale (/etc/sysconfig/language:RC_LANG) when generating the GRUB menu (bsc#998152)

src/lib/bootloader/grub2base.rb
```diff
@@ -148,6 +150,22 @@ def disable_serial_console
 
   private
 
+    def system_locale
```
mvidner: For the same reason as in the .changes file, I'd name this `systemwide_locale`

src/lib/bootloader/language.rb
```diff
@@ -0,0 +1,19 @@
+require "cfa/base_model"
+require "cfa/augeas_parser"
+
+module Bootloader
+  # read only model to get system language from sysconfig
+  # Uses CFA base model with augeas parser
+  class Language < CFA::BaseModel
```
mvidner: I think `Bootloader::Language` is wrong. The class reads the systemwide language settings, there is nothing bootloader specific. We may not go as far as refactoring [Yast::Language](https://github.com/yast/yast-country/blob/e1c633c63eedc8894e0b79874dea9d2f9c54351b/language/src/modules/Language.rb#L553) but we should name this class right so that it can be simply moved to another package when it is needed there.

`Sysconfig::Language` ?

src/lib/bootloader/grub2base.rb
```diff
@@ -148,6 +150,27 @@ def disable_serial_console
...
+      begin
+        language = ::Bootloader::Language.new
+        language.load
+      rescue Errno::ENOENT
```
mvidner: I think this is the wrong place. CFA::BaseModel or CFA::AugeasParses should handle the exception (or document that the user must do it, which they don't).

src/lib/bootloader/grub2base.rb
jreidinger: ok

src/lib/bootloader/language.rb
jreidinger: this means, that bootloader package will contain sysconfig directory in lib. I am not sure if I want to have it in bootloader. BTW I will be happy if such functionality already exist in YaST in easy way how to get it, but I do not find it.

mvidner: Well it is a generic class, but we're not moving it to a different package yet, so I have reconsidered and the current namespace is fine. Can we add a comment like "If you like this class, please move it to a base package and rename to Sysconfig::Language" ?

jreidinger: yes, it make sense for me.

src/lib/bootloader/grub2base.rb
mvidner: We don't have to change CFA now, but I'd like this PR to quote the future documentation or the ideal code usage

jreidinger: change is in https://github.com/config-files-api/config_files_api/pull/12

when playing with it, I found that exist? is not enough, as there can be problem with permissions, temporary read error and possible other problems.

jreidinger:
thanks

PR #377, opened 2016-11-11
Fix mountby label

jreidinger:


test/udev_mapping_test.rb
```diff
@@ -100,6 +100,13 @@
       expect(subject.to_mountby_device("/dev/vda1")).to eq "/dev/vda1"
     end
 
+    it "returns kernel device if mount by is used by label" do
```
imobachgs: np: Maybe something like "returns kernel device if mounted by label" is more readable IMHO.

jreidinger:
replaced with https://github.com/yast/yast-bootloader/pull/383

PR #381, opened 2016-11-22
Fix kexec

jreidinger:


test/sections_test.rb
```diff
@@ -50,6 +51,25 @@
...
+
+    it "handle localized grub.cfg" do
+      data_path = File.expand_path("../data/grub.cfg", __FILE__)
+      file = CFA::MemoryFile.new(File.read(data_path))
+      grub_cfg = CFA::Grub2::GrubCfg.new(file_handler: file)
+      grub_cfg.load
+
+      sections = Bootloader::Sections.new(grub_cfg)
+
+      puts sections.all.inspect
```
ancorgs: leftover?

src/lib/bootloader/kexec.rb
```diff
@@ -99,8 +99,8 @@ def copy_kernel
       cmd = Yast::Builtins.sformat(
         "/bin/cp -L %1/%2 %1/%3 %4",
         Yast::Installation.destdir,
-        "vmlinuz",
-        "initrd",
+        "boot/vmlinuz",
+        "boot/initrd",
```
ancorgs: Doesn't this deserve a test? As far as I see, the new test is only about the localization fix and has nothing to do with the original bugfix (these two lines).

test/sections_test.rb
jreidinger: right, thanks for notice

src/lib/bootloader/kexec.rb
jreidinger: you are right, it already have test, just it do not check exact location, see - https://github.com/yast/yast-bootloader/blob/master/test/kexec_test.rb#L87
I am not sure if checking exact location helps there, as it is single shot fix and can change in future ( I expect that vmlinuz symlink is arch specific, so it can be more tricky and we can get some bug reports, if SLE start using this kexec ).

test/sections_test.rb
```diff
@@ -50,6 +51,25 @@
 
       expect(subject.default).to eq "windows"
     end
+
+    it "raises exception if section do not exists" do
+      expect { subject.default = "non-exist" }.to raise_error(RuntimeError)
+    end
+
+    it "handle localized grub.cfg" do
```
jreidinger: will fix it

ancorgs:
LGTM

PR #382, opened 2016-11-22
Fix UEFI for storage-ng

ancorgs:
This fixes several errors introduced while porting the code to storage-ng in the previous PBI (see descriptions of individual commits for details about those errors) https://trello.com/c/NQ524dss/459-8-storageng-bootloader-proposal-efi

With these fixes, the UEFI system works for me in qemu.


shundhammer:
LGTM

PR #384, opened 2016-11-22
storage-ng: make the minimal legacy scenario work

ancorgs:


shundhammer:
Tentative LGTM from me, but somebody with deeper knowledge about the bootloader part (jreidinger?) should have a look at that, too.

src/lib/bootloader/mbr_update.rb
```diff
@@ -166,21 +178,21 @@ def bios_id_for(device)
...
-      partitions = target_map.fetch(disk, {}).fetch("partitions", [])
+      return [] unless disk
+
       # do not select swap and do not select BIOS grub partition
       # as it clear its special flags (bnc#894040)
-      partitions.select do |p|
-        p["used_fs"] != :swap && p["fsid"] != Yast::Partitions.fsid_bios_grub
+      devicegraph.disks.with(name: disk.name).partitions.reject do |part|
+        [Storage::ID_SWAP, Storage::ID_GPT_BIOS].include?(part.id)
       end
```
jreidinger: this is not how it works before. Previously it was more reboust if disk is partitionaless or if disk is not passed. what happen here if disk is not found?

src/lib/bootloader/mbr_update.rb
```diff
@@ -166,21 +178,21 @@ def bios_id_for(device)
...
     end
 
-    def extended_partition_num(disk)
-      part = activatable_partitions(disk).find { |p| p["type"] == :extended }
+    def extended_partition(disk)
+      part = activatable_partitions(disk).find { |p| p.type == Storage::PartitionType_EXTENDED }
```
jreidinger: just API note, it would be nice if type is object and can have method like `p.type.extended?`

src/lib/bootloader/mbr_update.rb
```diff
@@ -166,21 +178,21 @@ def bios_id_for(device)
...
       return nil unless part
 
-      num = part["nr"]
-      log.info "Using extended partition #{num} instead"
-      num
+      log.info "Using extended partition #{part.number} instead"
```
jreidinger: I am fine if youy use part.inspect instead. Especially when you do not return number, but object.

src/lib/bootloader/mbr_update.rb
ancorgs: The current version:

* Disk is not passed => returns []
* Disk is not found => returns []
* Disk has no partitions => returns []
* Disk has partitions => returns all partitions except swaps and gpt_bios

Wasn't that the original behavior?

src/lib/bootloader/stage1.rb
```diff
@@ -158,20 +160,20 @@ def available_locations
     end
 
     def can_use_boot?
-      tm = Yast::Storage.GetTargetMap
       partition = Yast::BootStorage.BootPartitionDevice
 
-      part = Yast::Storage.GetPartition(tm, partition)
+      parts = partitions.with(name: partition)
```
jreidinger: well, maybe in the end `Yast::BootStorage.BootPartitionDevice` can hold object instead of string, just idea, even less then NP :)

src/lib/bootloader/stage1.rb
```diff
@@ -158,20 +160,20 @@ def available_locations
...
         log.error "cannot find partition #{partition}"
         return false
       end
 
-      log.info "Boot partition info #{part.inspect}"
+      log.info "Boot partition info #{parts.first.inspect}"
 
       # cannot install stage one to xfs as it doesn't have reserved space (bnc#884255)
-      return false if part["used_fs"] == :xfs
+      fs = parts.filesystems.first
```
jreidinger: why not parts.first.filesystem ? it logs parts.first above and use something different?

src/lib/bootloader/stage1.rb
ancorgs: Of course it should.... in the future. We are just making sure the current code keep working. In the future a bigger refactoring to take advantage of the new functionalities in the library would be nice. No more passing strings but objects.. But that's phase II. :wink:

src/lib/bootloader/stage1.rb
ancorgs: Because `parts.first` returns a `Storage::Partition` object. Objects coming directly from libstorage usually exhibit surprising behavior for Ruby developers. For example, if the partition has no filesystem, `parts.first.filesystem` raises an exception. That's why we have the interface based in device lists, which is more ruby-friendly.  `parts.filesystems.first` returns nil in the same scenario.

src/lib/bootloader/stage1_proposal.rb
```diff
@@ -146,24 +152,30 @@ def boot_with_btrfs?
...
+          @logical_boot = true if p.type == Storage::PartitionType_LOGICAL
         end
       end
 
+      def with_btrfs?(partition)
+        partition.filesystem.type == ::Storage::FsType_BTRFS
+      rescue Storage::WrongNumberOfChildren
+        # No filesystem in the partition
+        false
+      end
```
jreidinger: hmmm, previously it looks better for me, to be honest I still do not like new API for it.

src/lib/bootloader/mbr_update.rb
ancorgs: I tend to agree. But that would be out of the scope of this PBI.

src/lib/bootloader/mbr_update.rb
ancorgs: I will change that.

src/lib/bootloader/mbr_update.rb
jreidinger: so if disks is not found by given name it return [] and not nil, right? then it is fine.

src/lib/bootloader/stage1.rb
jreidinger: ok, so result of not so nice api of this filesystem call. OK. So just result of my below note.

src/lib/bootloader/mbr_update.rb
ancorgs: right

jreidinger:
LGTM

PR #386, opened 2016-11-23
Quick fix for unit tests: Move broken tests out of the way

shundhammer:
Since "rake test:unit" also recurses into subdirectories of test/, I moved the tests that are currently broken due to storage-ng changes out of the way.

In order to be able to run individual broken tests to fix them, I added a symlink to test_helper.rb in ../test .


ancorgs:
Instead of moving the files, it would be probably more "polite" to disable the tests by changing
```ruby
it "whatever" do
 blah
end
```
into
```ruby
xit "whatever" do
  blah
end
```
See https://relishapp.com/rspec/rspec-core/v/2-0/docs/pending/pending-examples#temporarily-pending-by-changing-%22it%22-to-%22xit%22

ancorgs:
I meant, using "xit" or "pending" or any of the 4 mechanisms described in the link

jreidinger:
another way is to not end file with `_test.rb` and instead you can use `_test_not_ready.rb`. see https://github.com/yast/yast-rake/blob/master/lib/tasks/test_unit.rake

ancorgs:
Not sure whether you want to merge already or whether you are planning to add some storage-ng mocking to this same PR.

In the former case, it LGTM as it is.

shundhammer:
This PR is obsoleted by https://github.com/yast/yast-bootloader/pull/387 . Closing.

PR #388, opened 2016-11-28
storage-ng based tests for Stage1Device

ancorgs:


src/lib/bootloader/stage1_device.rb
```diff
@@ -59,11 +59,23 @@ def underlaying_devices_for(dev)
 
     # get one level of underlaying devices, so no recursion deeper
     def underlaying_devices_one_level(dev)
-      vgs = devicegraph.volume_groups.with(vg_name: dev)
-      return usable_pvs(vgs).disks.to_a unless vgs.empty?
+      # FIXME: there is nothing like this right now in libstorage-ng
+      #  a_vg.name #=> "/dev/vgname"
+      # revisit when such thing exist
+      match = /^\/dev\/(\w*)$/.match(dev)
```
shundhammer: probably not a real-life problem, but it should be \w+ IMHO - you want at least one word character, zero would be wrong

src/lib/bootloader/stage1_device.rb
```diff
@@ -59,11 +59,23 @@ def underlaying_devices_for(dev)
...
+        vgs = devicegraph.volume_groups.with(vg_name: match[1])
+        return usable_pvs(vgs).disks.map(&:name) unless vgs.empty?
+      end
 
-      lvs = devicegraph.logical_volumes.with(lv_name: dev)
-      return usable_pvs(lvs.vgs).map { |pv| pv.blk_device.name } unless lvs.empty?
+      # FIXME: there is nothing like this right now in libstorage-ng
+      #  a_lv.name #=> "/dev/vgname/lvname"
+      # revisit when such thing exist
+      match = /^\/dev\/\w*\/(\w*)$/.match(dev)
```
shundhammer: Same as above: \w+ instead of \w*

shundhammer:
LGTM

PR #392, opened 2016-12-13
Adapt test to new libstorage-id partition ids

ancorgs:
See https://github.com/openSUSE/libstorage-ng/pull/174


cwh42:
If we don't do version updates in devel branches it LGTM.

PR #393, opened 2016-12-19
Improve ui serial console

jreidinger:
blog content
==========

As experience shown in past, writting parameters for serial console is not easy task for majority of users and anyone can make mistakes in it. The error message when wrong ones are passed is not much helpful we decide to improve it and show new nicer popup including example. We hope that you find useful this usability improvements. See in attached screenshot how it will look:
serial_console

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Dec 19 12:40:33 UTC 2016 - jreidinger@suse.com
+
+- provide more helpful error message when invalid arguments for
+  serial console is provided (bsc#1012383)
+- 3.2.13
+
+-------------------------------------------------------------------
 Mon Dec  5 13:58:33 UTC 2016 - jreidinger@suse.com
 
 - stop failing with new cfa ( caused by deleting nil, change
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -2,7 +2,7 @@
 Mon Dec 19 12:40:33 UTC 2016 - jreidinger@suse.com
 
 - provide more helpful error message when invalid arguments for
-  serial console is provided (bsc#1012383)
+  serial console are provided (bsc#1012383)
 - 3.2.13
 
 -------------------------------------------------------------------
--- a/src/lib/bootloader/exceptions.rb
+++ b/src/lib/bootloader/exceptions.rb
@@ -27,7 +27,7 @@ module Bootloader
   end
 
   # Represents error when serial console arguments are not valid
-  class InvalidSerialConsoleArguments < RuntimeError
+  class InvalidSerialConsoleArguments < BrokenConfiguration
     MESSAGE = "Invalid serial console arguments".freeze
     def initialize(msg = MESSAGE)
       super
--- a/src/lib/bootloader/exceptions.rb
+++ b/src/lib/bootloader/exceptions.rb
@@ -26,8 +26,9 @@ module Bootloader
     end
   end
 
+  # Represents error when serial console arguments are not valid
   class InvalidSerialConsoleArguments < RuntimeError
-    MESSAGE = "Invalid serial console arguments"
+    MESSAGE = "Invalid serial console arguments".freeze
     def initialize(msg = MESSAGE)
       super
     end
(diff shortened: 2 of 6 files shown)


src/lib/bootloader/grub2_widgets.rb
```diff
@@ -488,6 +489,17 @@ def validate
           Yast::UI.SetFocus(Id(:console_args))
           return false
         end
+        if ::Bootloader::SerialConsole.load_from_console_args(console_value).nil?
+          Yast::Report.Error(
+            _("To enable serial console you must provide the corresponding arguments.\n" \
+              "Minimal requirement is unit parameter.\nWhole format is " \
+              "\"serial --speed=<number> --unit=<number> --word=<number> --parity=<odd|even|no> " \
```
mvidner: If it's too wide for the source file, it will be too wide for the ncurses UI.

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Dec 19 12:40:33 UTC 2016 - jreidinger@suse.com
+
+- provide more helpful error message when invalid arguments for
+  serial console is provided (bsc#1012383)
```
mvidner: arguments ARE provided

mvidner:
Complaining in the blog post that the users do not read help texts, while in fact NOT having any help text regarding the correct format of the serial console arguments... could anger the users.

src/lib/bootloader/exceptions.rb
```diff
@@ -25,4 +25,12 @@ def initialize(msg)
         "Please use YaST2 bootloader to fix it. Details: %s") % msg
     end
   end
+
+  # Represents error when serial console arguments are not valid
+  class InvalidSerialConsoleArguments < RuntimeError
```
mvidner: Why make a new exception class when you only add code to raise `BrokenConfiguration`?

src/lib/bootloader/grub2_widgets.rb
mvidner: I made it a bit complicated to make it less likely that translations mess it up:
```rb
# Translators: NUM is an abbreviation for "number", to be substituted in a command
# like "serial --unit=NUM --speed=NUM --parity={odd|even|no} --word=NUM --stop=NUM"
# so do not use punctuation
n = _("NUM")
syntax = "serial --unit=#{n} --speed=#{n} --parity={odd|even|no} --word=#{n} --stop=#{n}" 
# Translators: do not translate "unit" 
msg = _("To enable the serial console you must provide the corresponding arguments.\n"
"The \"unit\" argument is required, the complete syntax is:\n%s") % syntax
```

mvidner:
Some more ideas:
- We use `SerialConsole.load_from_console_args`. Would it make sense to let the user enter it in the kernel format `ttyS0,9600n8` (which is presumably more known) and translate it with `SerialConsole.load_from_kernel_args` when writing to the file? 
- In the help text we mention the `--stop` argument but SerialConsole does not model that. Is it OK?

src/lib/bootloader/grub2_widgets.rb
jreidinger: ok, sounds good, so lets use it.

jreidinger:
regarding ideas

- it is possible to translate it in other order, but it have to be consulted with UX team as in kernel args are more arguments, so maybe more confusing?

- regarding stop argument, it can be useful in some case, but I think kernel arg do not have counterpart for it. So I am not sure if it make sense. for stop argument see http://www.tldp.org/HOWTO/Remote-Serial-Console-HOWTO/preparation-setspeed.html

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -460,6 +461,32 @@ def contents
...
+      # so do not use punctuation
+      n = _("NUM")
+      "serial --unit=#{n} --speed=#{n} --parity={odd|even|no} --word=#{n} --stop=#{n}"
+    end
+
+    def help
+      # Translators: do not translate the quoted parts like "unit"
+      _(
+        "When graphical console is used, it will allow to use beside other" \
+        " features also different resolution. Option 'auto' try to find " \
```
mvidner: Thanks for the help text!
I will help with fixing the grammar... *but* only after you do your part: https://en.wikipedia.org/wiki/English_verbs#Third_person_singular_present

jreidinger:
thanks

PR #395, opened 2016-12-20
storage-ng: fixed legacy proposal with GPT

ancorgs:
According to my manual tests, this should fix http://yast-openqa.suse.cz/tests/403#step/install_and_reboot/1

PR #396, opened 2016-12-21
Use #with_name_or_partition to DRY the code

ancorgs:
I guess the tests will fail until https://github.com/yast/yast-storage-ng/pull/137 gets merged and published as RPM.


ancorgs:
Travis shouldn't be green in its first attempt. Looks like we are skipping a test we should not. Let me check.

ancorgs:
Looks like I forgot to re-enable the test I fixed as part of #395, added a commit re-enabling it that should turn travis temporarily red.

PR #400, opened 2017-01-06
Fix getting default

jreidinger:


src/lib/bootloader/grub2base.rb
```diff
@@ -120,6 +120,9 @@ def propose
 
       propose_serial
 
+      # we proposing new config and we do not know sections yet, so propose empty
```
afaerber: "we are proposing a new config"

afaerber: "propose empty" reads weird, but I don't have a better suggestion

jreidinger: "maybe propose no default"?

src/lib/bootloader/grub2base.rb
```diff
@@ -120,6 +120,9 @@ def propose
...
+      @sections.default = ""
```
mvidner: Isn't this a bug in the constructor of Sections? IMHO this code should go to https://github.com/yast/yast-bootloader/blob/f519e4818cecaea9f25c52e63a41b7b0692be0b5/src/lib/bootloader/sections.rb#L20

jreidinger: well, sections by default after construction do lazy loading of settings. It is question of default if we should read it in constructor? but then we have to have some read method, which now do not exist.

src/lib/bootloader/grub2base.rb
afaerber: Sure. (any noun will do)

src/lib/bootloader/grub2base.rb
mvidner: I still can't quite put my finger on it. Here we're changing `sections.default ` from "unknown" to "empty" so we should do the same to `sections.all`. Either here or in the constructor.

mvidner: To put it in yet another way: I suspect that we do not have a consistent state, either before this assignment, or after it.

jreidinger:
@mvidner yes, you understand correctly. So lets merge it

PR #401, opened 2017-01-09
try tracking feature of simplecov

jreidinger:


mvidner:
This looks good but we should describe it better - link to https://github.com/colszowka/simplecov/pull/422 
Also I want to understand what happens with https://coveralls.io/builds/9582948/source?filename=src%2Flib%2Fbootloader%2Fauto_client.rb

jreidinger:
@mvidner it is fine for me to have link, but I do not like referencing pull request as it can change later. I think better to reference configuration where is documentation like https://github.com/colszowka/simplecov/blob/master/lib/simplecov/configuration.rb#L53

regarding auto_client I am not sure. I will try it locally what it will report.

jreidinger:
@mvidner OK, now I reproduce it locally. Reason is simple. If we require everything, then simplecov mark all lines that are run during require ( like class and method definitions ) as test covered ( as it will fail if there is a problem ). But now it is not required, so if there is some problem it won't be detected, so it is not covered. Does it make sense?

mvidner:
OK, so the auto client is simply not covered.

Linking to specific line numbers on a symbolic reference like `master` will also break. How about this for the commit message:

> Switched to Simplecov.track_files to know about files that contain only untested code
>
> http://www.rubydoc.info/gems/simplecov/SimpleCov/Configuration#track_files-instance_method
>
> Previously we implemented this ourselves by loading all files but that produced false negatives for code that called a method without requiring its file.

PR #403, opened 2017-01-10
Edit AY settings

schubi2:


src/lib/bootloader/device_map.rb
```diff
@@ -43,8 +43,9 @@ def disks_order
     # or to label (ufo_disk)
     # @return [Hash{String => String}] new device map
     def remapped_hash
-      if !Yast::Arch.ppc
-        return to_hash if Yast::Storage.GetDefaultMountBy == :label
+      if Yast::Mode.config || # AutoYaST configuration mode --> no Storage available
+        !Arch.ppc && Storage.GetDefaultMountBy == :label
```
jreidinger: here, it have to be `Yast::Arch`

jreidinger: and also Yast::Storage as device map do not live in yast namespace

jreidinger:
LGTM

PR #405, opened 2017-01-10
test forced encoding

jreidinger:


mvidner:
Why?

jreidinger:
@mvidner due to failure in jenkins where is LANG set to "C" - http://ci.opensuse.org/job/yast-bootloader-master/584/Changes

jreidinger:
problem is that LANG set default external encoding, which is used for reading this file. But this file is always UTF-8

mvidner:
I see. This is related to https://github.com/yast/yast-ruby-bindings/pull/173 : YaST assumes UTF-8 and breaks when  something deviates (the actual locale in 173, the declared locale here).
It is not specific to the tests, the production code would fail in the same way. But let's solve that together with https://fate.suse.com/321409 "Make YaST work in non-Unicode locales"

jreidinger:
thanks

PR #406, opened 2017-01-17
Implement simple_mode -> label_proposal (FATE#322328)

mvidner:
https://trello.com/c/XDgl2dj7/816-13-casp-new-all-in-one-dialog
https://github.com/yast/yast-yast2/pull/530

### DIFF
--- a/src/lib/bootloader/proposal_client.rb
+++ b/src/lib/bootloader/proposal_client.rb
@@ -153,7 +153,7 @@ module Bootloader
 
       ret["links"] = PROPOSAL_LINKS # use always possible links even if it maybe not used
       ret["raw_proposal"] = Yast::Bootloader.Summary
-      ret["label_proposal"] = Yast::Bootloader.Summary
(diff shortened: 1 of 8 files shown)


src/lib/bootloader/proposal_client.rb
```diff
@@ -153,6 +153,7 @@ def construct_proposal_map
 
       ret["links"] = PROPOSAL_LINKS # use always possible links even if it maybe not used
       ret["raw_proposal"] = Yast::Bootloader.Summary
+      ret["label_proposal"] = Yast::Bootloader.Summary
```
jreidinger: so no check for simple_mode param? BTW to make it more efficient you can use for benefit knowledge that never both values are modified, so it can be
```
ret["raw_proposal"] = Yast::Bootloader.Summary
ret["label_proposal"] = ret["raw_proposal"]
```

teclator:
LGTM

PR #410, opened 2017-02-16
storage-ng: fixes to work with LVM and encryption

ancorgs:
This should fix the current failures in openQA when calculating the bootloader proposal. Verified to work locally.

### DIFF
--- a/src/modules/BootStorage.rb
+++ b/src/modules/BootStorage.rb
@@ -165,14 +165,24 @@ module Yast
       true
     end
 
-    # Find the blkdevice for the filesystem mounted at mountpoint. Returns nil
-    # if no filesystem is found or the filesystem has no blkdevice (e.g. NFS).
+    # Find the partition or disk for the filesystem mounted at mountpoint.
+    #
+    # Returns nil if no filesystem is found or the filesystem has no blkdevice
+    # (e.g. NFS).
+    #
+    # If the filesystem is in a virtual device (like a LUKS or LVM volume), it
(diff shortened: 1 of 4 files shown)


src/modules/BootStorage.rb
```diff
@@ -165,14 +165,24 @@ def bootloader_installable?
...
+    # returns the (first) underlying partition or disk.
     def find_blk_device_at_mountpoint(mountpoint)
-      fses = Storage::Filesystem.find_by_mountpoint(staging, mountpoint)
+      fses = staging.filesystems.with_mountpoint(mountpoint)
       return nil if fses.empty?
-      return nil if fses[0].blk_devices.empty?
 
-      fses[0].blk_devices[0]
+      partitions = fses.partitions
+      partitions = fses.lvm_vgs.partitions if partitions.empty?
```
jreidinger: hmm, I expect that it always return partitions and not need this special lvm handling, if we in future support more such stuff, how complex can be this combination? like lvm, device_mapper, various raids and so on. We already in bootloader fighting with it that we have problem to get physical device for such virtual ones and I hope new storage abstract it.

ancorgs: Initially it was written this way because there is not 1:1 relationship between formatted LVs and partitions.

We have already realized that's not a strong enough reason and we plan to change it to work as you suggest. But I wanted to fix this first for openQA. This will be adapted as we change the DevicesList API to work as you expect (something we plan to do in the very short term).

PR #412, opened 2017-03-03
avoid call to old libstorage

wfeldt:
There's at least another Storage.InitLibstorage() in the code but it's not hit, so I leave it as is for now.

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Mar  3 15:33:11 CET 2017 - snwint@suse.de
+
+- storage-ng: avoid call to old libstorage
+- 3.3.5
+
+-------------------------------------------------------------------
 Thu Feb 16 14:47:03 UTC 2017 - ancor@suse.com
 
 - storage-ng: fixed proposal to work with LVM and encrypted LVM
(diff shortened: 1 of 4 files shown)


ancorgs:
Since that method is called only once, I would kill it completely, instead of redefining it to `true`.

wfeldt:
Problem is I'm not sure what the point of the function was anyway. With all those different mode variants there are so many workflows trough the code. :-/

ancorgs:
Well, if we want to keep the method, then I guess we should do something really equivalent, instead of just returning `true`. As I understand the original code, it's main purpose if to ensure libstorage is initialized, returning false if the initialization fails. So I guess we could do:
```ruby
def checkUsedStorage
  Y2Storage::StorageManager.instance
  true
end
```

The initialization of storage-ng never fails, so no need to for the second part of the check.

wfeldt:
Uhm, I'd rather not. Let's get the code simpler. from what I see in
https://bugzilla.suse.com/show_bug.cgi?id=419197 it was a rather strange thing anyway.

ancorgs:
What is simpler than just deleting the method the calls to it? :wink:

This is what the original code did (@jreidinger, please correct me if I'm wrong).
 * If libstorage was already initialized, return true
 * If libstorage was not initialized, do it (in readonly mode)
   * If the initialization was ok, return true
   * If the initialization failed, return false only in normal mode, true otherwise

I don't believe we need to explicitly initialize yast-storage-ng and, in case we do, the initialization never returns false, so.... kill that method!

ancorgs:
LGTM

PR #413, opened 2017-03-09
don't create grub_installdevice and let perl-Bootloader run grub2-install

wfeldt:
The current code uses a quite complex logic to determine the grub2
location. And the result is not what storage-ng expects (grub2 installed
into mbr).

This bypasses the yast-bootloader logic for now until something better
can be implemented.

### DIFF
--- a/src/lib/bootloader/grub2.rb
+++ b/src/lib/bootloader/grub2.rb
@@ -78,7 +78,11 @@ module Bootloader
       end
       # Do some mbr activations ( s390 do not have mbr nor boot flag on its disks )
       # powernv do not have prep partition, so we do not have any partition to activate (bnc#970582)
+# storage-ng
+# the default is not to install into mbr
+=begin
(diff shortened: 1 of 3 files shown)


jreidinger:
well, it is quite big change that need at first to be discussed (as I am not sure if always MBR is something users want and welcome). It also adds again dependency on perl-bootloader, which was dropped in past.

wfeldt:
This is just POC code and not intended to be merged (not even for storage-ng).
But I think it's worth pursuing this idea  - see bsc 994322.

grub2 into MBR is certainly not always the correct idea (say when the disk is shared with other
systems). But it's the best in most cases and storage-ng assumes it when making a proposal
(sanctioned from the grub2 side (mchang) and other architecture experts).

jreidinger:
well, logic on y2-bl is now improved a lot, so lets close this POC.

PR #414, opened 2017-03-20
Use shared RSpec examples for CWM widgets

mvidner:


mvidner:
Using https://github.com/yast/yast-yast2/pull/555

mvidner:
Ugh, in Travis I get a ExpiredTestDoubleError and I have no idea how that appears.

jreidinger:
@mvidner my idea is that widget need to return some value some mocking and it is missing there, and instead double from other test is used. Btw see my LabeledWidget shared example which looks similar to your attempt.

jreidinger:
yast2 already have rspec test, can you please rebase it? then I will merge it.

PR #417, opened 2017-03-28
Lvm part storage ng

jreidinger:


src/lib/bootloader/stage1.rb
```diff
@@ -175,6 +175,9 @@ def can_use_boot?
       fs = parts.filesystems.first
       return false if fs && fs.type == ::Storage::FsType_XFS
 
+      # LVM partition does not have reserved space for LVM
```
ancorgs: Didn't you mean "reserved space for the bootloader" or something like that?

src/lib/bootloader/stage1_proposal.rb
```diff
@@ -95,6 +95,14 @@ def propose
 
     private
 
+      def any_boot_flag_partition?(disk_name)
+        disks = devicegraph.disks.with(name: disk_name)
+        legacy_boot = disks.first.partition_table.partition_legacy_boot_flag_supported?
```
ancorgs: Are you sure the disk will always have a partition table at this point? If not this can blow away

src/lib/bootloader/stage1.rb
jreidinger: yes, it is reserved space for stage1.5

src/lib/bootloader/stage1_proposal.rb
jreidinger: well, if we use for booting any disk without partition table, then we have serious problem. But I can change it to false if  there is no boot_flag. But still if disk won't have partition table, then bootloader later blown up

src/modules/BootSupportCheck.rb
```diff
@@ -196,8 +198,14 @@ def check_zipl_part
...
 
     def check_activate_partition
-      # activate set or there is already activate flag
-      return true if stage1.activate? || Yast::Storage.GetBootPartition(Yast::BootStorage.mbr_disk)
+      # activate set
+      return true if stage1.activate?
+
+      # there is already activate flag
+      disks = staging.disks.with(name: Yast::BootStorage.mbr_disk)
+      legacy_boot = disks.first.partition_table.partition_legacy_boot_flag_supported?
```
ancorgs: Same that above. Are we sure we have a partition table?

src/modules/BootSupportCheck.rb
```diff
@@ -101,12 +104,11 @@ def correct_loader_type(lt)
...
       return true unless stage1.mbr?
 
-      devices = Storage.GetTargetMap
-      mbr_disk = Storage.GetDisk(devices, BootStorage.mbr_disk)
-      boot_device = Storage.GetPartition(devices, BootStorage.BootPartitionDevice)
-      return true if mbr_disk["label"] != "gpt"
-      return true if boot_device["used_fs"] != :btrfs
-      return true if mbr_disk["partitions"].any? { |p| p["fsid"] == Partitions.fsid_bios_grub }
+      disk = staging.disks.with(name: BootStorage.mbr_disk).first
+      boot_device = staging.disks.partitions.with(name: BootStorage.BootPartitionDevice).first
```
ancorgs: NP: you can do `staging.partitions` directly here. If you are not filtering by disk there is no need to pass through that collection.

ancorgs:
LGTM

PR #418, opened 2017-04-04
[POC] try run osc:build in travis

jreidinger:


lslezak:
## Disadvantages

Here are some reasons why I decided to **not** use `rake osc:build` at Travis in the Docker build as it has several disadvantages:

- It does not support code coverage using *coveralls* as the Git data like branch, commit, etc... are missing. That's because the package is built from tarball without the `.git` directory.
- The build takes much more time. The [Travis history](https://travis-ci.org/yast/yast-bootloader/builds) shows that the previous builds usually took less than 1:30 with `osc` build we are at ~6 minutes
- It downloads 222 packages - besides the slowdown mentioned above it raises the traffic for Travis (although I do not know whether this is a real problem or not)
- The additional checks are not run, e.g. Rubocop. The tricky part is that we do not want to run it in OBS for normal builds as it does not make sense there and it would just slowdown the build and add more build dependencies.

I tried to solve the coveralls and Rubocop issue in the past: see https://github.com/yast/yast-registration/compare/jenkins_coveralls_support?expand=1 and commented commands at the end of the build script at https://ci.opensuse.org/view/Yast/job/yast-registration-github-push/configure
- It passes the Git info via quite ugly RPM build parameters
- It uses `yast-rake-ci` gem which pulls in the Rubocop&Co. dependencies
- The `check:ci` task takes care of running Rubocop

But this does not look nice and is a bit tricky and complex :confused:.  The easier way is to run the tests twice, first in the normal Travis environment (with code coverage) and then as the part of the RPM package build (without code coverage). This is the current solution.

## Advantages?

However, the question is what are the advantages of using osc build?

- We already build the RPM packages at Travis, although using simpler `rpmbuild` command, but the packaging issues should be still caught
- The built packages are [installed/upgraded/uninstalled](https://github.com/yast/docker-yast-ruby/blob/master/yast-travis-ruby#L59) to verify the `%post`/`%pre`/etc... scripts
- We [run the osc `source_validator` script](https://github.com/yast/docker-yast-ruby/blob/master/yast-travis-ruby#L40) which is internally used by `osc`

Is there anything missing in the current approach? The `rake check:license` call is in my TODO list... ([PBI](https://trello.com/c/YLWtoKFR/915-todo-run-rake-check-license-at-travis))

jreidinger:
well, reason why I am trying this is that I want to have unified ENV. so if we use docker and rake osc for submitting, we should use same in travis.  we can run rubocop or rake test:unit after rake osc:build. My goal is just to test we can use it in docker.

jreidinger:
advatages is:

- uniform env where rpm build happen, no need for dependencies in image
- same as what we actually do and need when submitting packages
- new checks are automatic added like e.g. changelog checker, which we currently do not have or license check and maybe other checks
- we run same check as osc without need to mimic it.

lslezak:
I'm fine with using this at Jenkins, but I'm not sure if using the very same approach at Travis is good... :question:

jreidinger:
@lslezak well, in that case we will need two dockerfiles and separate env. Well, I try to write it down to PBI card. If we use same setup for both, then in the end only difference is to run more stuff in travis and to have different .oscrc in home.

jreidinger:
decision is to not run osc:build in travis

PR #419, opened 2017-04-06
add to travis checking of valid documentation

jreidinger:
See https://github.com/openSUSE/packaging_rake_tasks/pull/34.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,2 @@
 FROM yastdevel/ruby
 COPY . /usr/src/app
-
-# temporary to use the last packaging tasks
-# TODO: remove when it will be in openSUSE:Factory
-RUN gem install packaging_rake_tasks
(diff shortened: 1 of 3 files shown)


lslezak:
Just for the record: it uses this new functionality: https://github.com/openSUSE/packaging_rake_tasks/pull/34/

Dockerfile
```diff
@@ -1,3 +1,6 @@
 FROM yastdevel/ruby
 COPY . /usr/src/app
 
+# temporary to use the last packaging tasks
+# TODO: remove when it will be in openSUSE:Factory
```
lslezak: You do not have to wait for Factory, the Docker image installs the packages from YaST:Head: https://github.com/yast/docker-yast-ruby/blob/master/Dockerfile#L2 so the latest package should be installed soon... (You can force the Docker image to not wait for ~2hours auto rebuild).

jreidinger: good to know, thanks

PR #420, opened 2017-04-13
Trusted boot efi

jreidinger:


src/lib/bootloader/grub2efi.rb
```diff
@@ -109,7 +113,8 @@ def packages
 
     # overwrite BootloaderBase version to save secure boot
```
ancorgs: NP: slightly outdated comment?

test/grub2base_test.rb
```diff
@@ -27,6 +27,15 @@
...
       subject.read
     end
+
+    it "reads trusted boot configuration from sysconfig" do
+      mocked_sysconfig = ::Bootloader::Sysconfig.new(trusted_boot: true)
+      expect(::Bootloader::Sysconfig).to receive(:from_system).and_return(mocked_sysconfig)
+
+      subject.read
+
+      expect(subject.trusted_boot).to eq true
```
ancorgs: NP: strictly speaking, nothing ensures this `true` is coming from `mocked_sysconfig.trusted_boot`. I would like to see a complementary case with `false` to be sure.

test/grub_install_test.rb
```diff
@@ -93,6 +93,17 @@ def expect_grub2_install(target, device: nil, removable: false)
 
         subject.execute
       end
+
+      it "pass suse-enable-tpm option when trusted boot is requested" do
```
ancorgs: NP: passes

src/lib/bootloader/grub2efi.rb
jreidinger: in fact no, as difference between grub2 and grub2efi version is only secure boot, now when both have trusted boot option.

test/grub2base_test.rb
jreidinger: well, default are false. But I will add it.

PR #423, opened 2017-04-25
Boot flags ng

jreidinger:


src/lib/bootloader/grub2base.rb
```diff
@@ -64,14 +64,14 @@ def pmbr_setup(*devices)
...
+      when :remove then false
       else raise "invalid action #{action}"
       end
 
-      devices.each do |dev|
-        Yast::Execute.locally("parted", "-s", dev, "disk_set", "pmbr_boot", action_parted)
-      end
+      staging = Y2Storage::StorageManager.instance.y2storage_staging
+      disks = staging.disks.select { |d| devices.include?(d.name) }
+      disks.each { |d| d.pmbr_boot = action_parted }
```
ancorgs: This will not call parted immediately. You need to call `Y2Storage::StorageManager.commit` to commit to the disks whatever you have modified in `staging`.

src/lib/bootloader/mbr_update.rb
```diff
@@ -72,54 +72,26 @@ def install_generic_mbr
...
-    end
-
-    def can_activate_partition?(disk, num)
+    def can_activate_partition?(partition)
       # if primary partition on old DOS MBR table, GPT do not have such limit
-      gpt_disk = gpt?(disk)
+      gpt_disk = partition.disk.gpt?
 
-      !(Yast::Arch.ppc && gpt_disk) && (gpt_disk || num <= 4)
+      !(Yast::Arch.ppc && gpt_disk) && (gpt_disk || partition.number <= 4)
```
ancorgs: Is that `<4` a hacky way to check if the partition is not logical? In that case, you can now ask that to `partition.type`.

src/lib/bootloader/mbr_update.rb
```diff
@@ -183,8 +156,8 @@ def activatable_partitions(disk)
       disk.partitions.reject { |p| p.id.is?(:swap, :bios_boot) }
     end
 
-    def extended_partition(disk)
-      part = activatable_partitions(disk).find { |p| p.type == Y2Storage::PartitionType::EXTENDED }
+    def extended_partition(partition)
+      part = partition.siblings.find { |p| p.type.is?(:extended) }
```
ancorgs: When using `partition.siblings` take into account the the devicegraph looks like this if there are logical partitions:
```
          sda
  _____________________
 |       |      |      |
sda1    sda2   sda3   sda4
                      ____        
                     |    |
                   sda5  sda6
```
So `sda5.siblings` will only return `sda6`, not all the partitions. In the same way, `sda1.siblings` will not include sda5 nor sda6.

src/lib/bootloader/grub2base.rb
jreidinger: so, I need to do it set this flags in proposal and let storage do it during its write section and for runtime, call it manually. Right? Or can I call it always in this location, even in bootloader installation in first stage?

src/lib/bootloader/mbr_update.rb
jreidinger: To be honest I am not sure, because it was quite old code. But I think we can change it now.

src/lib/bootloader/mbr_update.rb
jreidinger: ah, so sda5 is child of sda4? so for extended partition I in fact need like `partition.parent` ? as I call it only for logical partitions.

jreidinger: in fact, what will return `partition.disk` ? if partition is logical? as disk is not direct parent.

ancorgs: Of course, `partition.disk` returns the disk for any partition. I don't see why the fact that the disk is not it's direct parent should change that. In fact, the disk is not the direct parent for primary partitions either. The direct parent would be a `PartitionTable` object in that case.

ancorgs: I would avoid using methods like `siblings` because the organization of the devicegraph can be surprising sometimes. In general, if there is a more "semantic" way of doing things (instead of using methods that traverse the graph at "low level"), that's by far preferred.

For example, something like `partition.partitionable.partitions`. Once again, if you thing is worth it, we can even add `Partition#other_partitions` or something like that.

jreidinger: ok, taking note about it. I just think that siblings make code more readable. I think it would be worth to mention there surprises in siblings documentation.

src/lib/bootloader/grub2base.rb
ancorgs: Oh, well, if this happens during installation before preparing the disks. Then it's fine. The `inst_prep_disk` client will commit the changes when the user press "install" in the summary screen.

But when running the bootloader module standalone or when doing changes at the end of installation, after installing the packages and all that, I guess you have to commit the changes (I'm not aware of any client doing a final commit).

src/lib/bootloader/mbr_update.rb
```diff
@@ -72,54 +72,26 @@ def install_generic_mbr
...
-    end
-
-    def can_activate_partition?(disk, num)
+    def can_activate_partition?(partition)
       # if primary partition on old DOS MBR table, GPT do not have such limit
-      gpt_disk = gpt?(disk)
+      gpt_disk = partition.disk.gpt?
 
-      !(Yast::Arch.ppc && gpt_disk) && (gpt_disk || num <= 4)
+      !(Yast::Arch.ppc && gpt_disk) && (gpt_disk || partition.is?(:primary, :extended))
```
ancorgs: You meant `partition.type.is?(:primary, :extended)`

src/lib/bootloader/mbr_update.rb
```diff
@@ -72,54 +72,26 @@ def install_generic_mbr
...
+        next unless can_activate_partition?(partition)
 
-        log.info "Activating partition #{num} on #{disk}"
+        log.info "Activating partition #{partition.inspect}"
         # set corresponding flag only bnc#930903
         if gpt?(disk)
-          set_parted_flag(disk, num, "legacy_boot")
+          # for legacy_boot storage_ng do not reset others, so lets
+          # do it manually
+          partition.siblings.select { |d| d.is?(:partition) }.each { |p| p.legacy_boot = false }
```
ancorgs: `siblings` is properly used here? Regarding the already discussed logical vs primary partitions issue, I mean.

jreidinger: yes, as gpt do not have primary versus logical and what I need is to reset other partitions, so I hope it will work, but of course I can change it to `partition.disk.partitions` instead

ancorgs:
Please, fix this in `BootSupportCheck.rb`
```ruby
Y2Storage::StorageManager.instance.y2bootloader_staging
```

Is not visible in the final diff, but I believe you fixed it and then un-fixed it.

PR #424, opened 2017-05-30
Storage timestamp

jreidinger:


src/modules/BootStorage.rb
```diff
@@ -175,7 +182,8 @@ def bootloader_installable?
 
     # Sets properly boot, root and mbr disk.
     def detect_disks
-      return unless @RootPartitionDevice.empty? # quit if already detected
+      # Use cachced value if already detected and cache still valid
```
ancorgs: typo in comment: cachced

src/lib/bootloader/proposal_client.rb
```diff
@@ -36,9 +36,16 @@ def initialize
...
+      storage_changed = Yast::BootStorage.storage_changed?
+      # redetect disks if cache is invalid as first part
+      Yast::BootStorage.detect_disks if storage_changed
+
+      # logic in this condition:
+      # when reset is forced or user do not modify proposal, reset proposal,
+      # but only when not using auto_mode
+      # But if storage changed, always repropose as it can be very wrong.
+      if ((force_reset || !Yast::Bootloader.proposed_cfg_changed) &&
+          !auto_mode) || storage_changed
```
ancorgs: NP: maybe moving this to a method would allow more clear conditions, like (just a possible example):
```ruby
def reset_needed?(force_reset)
  return true if storage_changed
  return false if Yast::Mode.autoinst || Yast::Mode.autoupgrade
  force_reset || !Yast::Bootloader.proposed_cfg_changed
end
```

jreidinger: make sense for me

src/modules/BootStorage.rb
jreidinger: thanks

ancorgs:
Looks better now, so my approval still stands. Travis is failing, nevertheless.

PR #427, opened 2017-05-30
Storage timestamp ng

jreidinger:
intentionally do not increase version.

### DIFF
--- a/test/boot_storage_test.rb
+++ b/test/boot_storage_test.rb
@@ -89,6 +89,18 @@ xdescribe Yast::BootStorage do
 
       expect(subject.mbr_disk).to eq "/dev/vda"
     end
+
+    it "skips cache if storage gets changed" do
+      subject.RootPartitionDevice = "/dev/sda1"
+      subject.BootPartitionDevice = "/dev/sda2"
+      subject.mbr_disk = "/dev/sda"
+
+      allow(Yast::Storage).to receive(:GetTargetChangeTime).and_return(Time.now.to_i)
+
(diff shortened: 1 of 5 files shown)


test/boot_storage_test.rb
```diff
@@ -89,6 +89,18 @@
 
       expect(subject.mbr_disk).to eq "/dev/vda"
     end
+
+    it "skips cache if storage gets changed" do
+      subject.RootPartitionDevice = "/dev/sda1"
+      subject.BootPartitionDevice = "/dev/sda2"
+      subject.mbr_disk = "/dev/sda"
+
+      allow(Yast::Storage).to receive(:GetTargetChangeTime).and_return(Time.now.to_i)
```
ancorgs: Hmmm

test/boot_storage_test.rb
```diff
@@ -89,6 +89,18 @@
...
+
+      subject.detect_disks
+
+      expect(subject.RootPartitionDevice).to eq "/dev/vda1"
+    end
```
ancorgs: Now that I saw the previous test passing even it was wrong, I'm kind of worried. Can we have a complementary test or something?

jreidinger: see xdescribe on top of file. This test is disabled for now. As its testing require solid set of device graph mocking, which is currently a bit out of scope of this PR.

ancorgs: Ah, ok

jreidinger:
@ancorgs thanks

PR #429, opened 2017-06-12
Fix nil device

jreidinger:


jreidinger:
@schubi2 test case in that part is not so easy, as it require to create target map from which it is proposed. And due to plan to drop it for SLE15, I do not plan to invest too much into it. Btw this problem won't happen in storage-ng branch.

PR #430, opened 2017-06-14
Udev for prep

jreidinger:


src/lib/bootloader/stage1_proposal.rb
```diff
@@ -198,7 +199,9 @@ class PPC < Stage1Proposal
       def propose
         partition = proposed_prep_partition
         if partition
-          assign_bootloader_device([:custom, partition])
+          # ensure that stage1 device is in udev (bsc#1041692)
```
mvidner: No. "... is in udev **form**"
I am sick of careless language usage *that obscures the meaning which it is intended to illuminate*.

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jun 14 14:54:38 UTC 2017 - jreidinger@suse.com
+
+- Use udev device for prep partition if it is available
```
mvidner: No. There are no udev devices. There are only udev device **names**.
Also, which is the thing that is available: the udev name or the prep partition?

src/lib/bootloader/stage1_proposal.rb
jreidinger: Well, it is about priorities. I spend almost whole day implementing and testing it. Of course with my English skill, I can easily spend another couple of hours to try to produce precise an English sentences, but I have other tasks in my queue and in scrum board.
So for me question is where my time bring the most value to customers and also to rest of team. So if a comment in a code is understandable and developer understand what I want to say, even if not precisely and perfectly correct, it is enough for me.

PR #434, opened 2017-06-28
fixed lookup to handle DASDs

aschnell:
For https://trello.com/c/lSOCC9R8/643-1-storageng-s-390-bootloader-proposal-failure.

### DIFF
--- a/src/modules/BootSupportCheck.rb
+++ b/src/modules/BootSupportCheck.rb
@@ -179,6 +179,9 @@ module Yast
     end
 
     def check_zipl_part
+# storage-ng
+=begin
+
       # if partitioning worked before upgrade, it will keep working (bnc#886604)
       return true if Mode.update
 
@@ -190,6 +193,8 @@ module Yast
         add_new_problem(_("Missing ext partition for booting. Cannot install boot code."))
         return false
       end
+
+=end
 
       true
     end
--- a/src/modules/BootStorage.rb
+++ b/src/modules/BootStorage.rb
@@ -236,7 +236,7 @@ module Yast
     end
 
     def disk_with_boot_partition
-      disk = boot_partition.disk
+      disk = boot_partition.partitionable
 
       log.info "Boot device - disk: #{disk}"
 


shundhammer:
LGTM

src/modules/BootSupportCheck.rb
```diff
@@ -179,6 +179,9 @@ def efi?
     end
 
     def check_zipl_part
+# storage-ng
+=begin
```
ancorgs: Well, translating this code to storage-ng should be quite easy. Any reason to disable it instead of updating it?

src/modules/BootSupportCheck.rb
```diff
@@ -191,6 +194,8 @@ def check_zipl_part
         return false
       end
```
ancorgs: This should work. Isn't it?
```ruby
boot_fs = staging.filesystems.find { f| f.mountpoint == "/boot/zipl" }
boot_fs ||= staging.filesystems.find { f| f.mountpoint == "/boot" }
boot_fs ||= staging.filesystems.find { f| f.mountpoint == "/" }

# It would be better to improve Filesystems::Type#is? to support :ext, but this will also make it
if ![:ext2, :ext3, :ext4].include? boot_fs.type.to_sym
  add_new_problem(_("Missing ext partition for booting. Cannot install boot code."))
  return false
end
```

aschnell: Overall this is code duplication since AFAIK the storage proposal has boot requirement checkers.
Those should be used.

jreidinger: well, this is protection against expert partitioner known issues from past. I am fine if it will be completely in storage.

ancorgs: @aschnell just to clarify, the comment ("it would be better.... bla") is just for you. Not expected to be in the final code. :-)

ancorgs: Ok. Reusing the logic from `Y2Storage::BootRequirementsChecker` sounds reasonable. But then this code should be substituted by a call to that checker, not completely commented out without even including a comment explaining why the code was commented instead of updated.

That being said. This logic does not match 100% with the logic implemented in `BootRequirementsChecker` as it is [documented here](https://github.com/yast/yast-storage-ng/blob/master/doc/boot-requirements.md#needed-partitions-in-a-s390-system). According to that, the checker always requires a separate extX partition mounted at `/boot/zipl`. But the code we just commented is fine if that partition is not completely separate, as long as the filesystem containing `/boot/zipl` is extX. We need to clarify what is more correct.

jsrain: I'm fine leaving this discussion to later; fixing the BootRequirementChecker (in bootloader module) is in the backlog, we can discuss there which one to use as well as individual checks.

jreidinger: According to my knowledge, ext partition can be even just "/". And I check it elsewhere and for installing gentoo they suggest to have simple "/" as ext4 and no other partitions, see https://wiki.gentoo.org/wiki/S390/Install#Initializing_and_formatting_the_DASD_disks

PR #435, opened 2017-07-04
Do not change mode when detecting disks in AutoYaST

imobachgs:
It looks like workaround for [bsc#950105](https://bugzilla.suse.com/show_bug.cgi?id=950105) is not needed anymore.


imobachgs:
Thanks!

PR #436, opened 2017-07-05
 Set StorageDevices flag disks_valid to true

schubi2:


package/yast2-bootloader.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Jul  5 08:56:43 CEST 2017 - schubi@suse.de
+
+- AuotYaST configuration: Set StorageDevices flag disks_valid
```
ancorgs: Typo: AuotYaST

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,12 @@
...
+  to true. So libstorage can scan valid disks and "/" partition
+  can be found. (bnc#1046738)
```
ancorgs: Since we have to merge this to SP3 and master branches, to keep the branches diff in good shape, I would also add a reference to bcn#1043132 (the original bug, this fix is just a backport) here.

src/modules/BootStorage.rb
```diff
@@ -192,8 +192,10 @@ def detect_disks
       if Mode.config
         Mode.SetMode("normal")
         log.info "Initialize libstorage in readonly mode" # bnc#942360
+        # Set StorageDevices flag disks_valid to true. So InitLibstorage
+        # can scan valid disks. (bnc#1046738)
```
ancorgs: Same than above I would reference both bugs here so merging to SP3/master don't kill the old number (the original bug one).

PR #437, opened 2017-07-05
added bnc#1046738 to decription

schubi2:


schubi2:
have to merge it from sp2

PR #438, opened 2017-07-05
Sle 12 sp3 bnc 1046738 1

schubi2:


src/modules/BootStorage.rb
```diff
@@ -193,7 +193,7 @@ def detect_disks
         Mode.SetMode("normal")
         log.info "Initialize libstorage in readonly mode" # bnc#942360
         # Set StorageDevices flag disks_valid to true. So InitLibstorage
-        # can scan valid disks. (bnc#1043132)
+        # can scan valid disks. (bnc#1046738, bnc#1046738)
```
ancorgs: Meck! We repeated the same bug number here. The error is already present in the SP2 branch. Let's fix it there before merging.

PR #441, opened 2017-07-05
bnc#1046738

schubi2:


schubi2:
no info available about merge. open another one

PR #450, opened 2017-07-31
[WIP] add test to simulate double entry write

jreidinger:


jreidinger:
fixed by https://github.com/config-files-api/config_files_api_grub2/pull/19

PR #451, opened 2017-08-02
not everything is a disk, use libstorage's partitionable instead

wfeldt:
This is for: https://trello.com/c/mUEtIRQN

### DIFF
--- a/src/lib/bootloader/stage1_proposal.rb
+++ b/src/lib/bootloader/stage1_proposal.rb
@@ -101,8 +101,9 @@ module Bootloader
     private
 
       def any_boot_flag_partition?(disk_name)
-        disk = devicegraph.disks.find { |d| d.name == disk_name }
-        return false unless disk.partition_table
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/stage1_proposal.rb
```diff
@@ -101,8 +101,9 @@ def propose
     private
 
       def any_boot_flag_partition?(disk_name)
-        disk = devicegraph.disks.find { |d| d.name == disk_name }
-        return false unless disk.partition_table
+        disk = Y2Storage::Partitionable.find_by_name(devicegraph, disk_name)
+        log.info "any_boot_flag_partition?: #{disk ? disk.name : 'nil'}"
```
jreidinger: if you add logging I suggest to add also parameter like
```
any_boot_flag_partition?(#{disk_name}): #{disk.inspect}"
```

disk inspect is more useful, as it shows more details about disk.

PR #452, opened 2017-08-04
more fixes where code assumes disk devices instead of libstorage's partitionable

wfeldt:
Most notably, dm devices are not disks in libstorage.

Also, the logic in #to_partition seemed wrong. It must return nil if there
are no partitions, not the main device, as the function using it assumes a
typical partition class is returned.

### DIFF
--- a/src/lib/bootloader/mbr_update.rb
+++ b/src/lib/bootloader/mbr_update.rb
@@ -106,7 +106,7 @@ module Bootloader
     def activate_partitions
       partitions_to_activate.each do |partition|
         num = partition.number
-        disk = partition.disk
+        disk = partition.partitionable
         if num.nil? || disk.nil?
           raise "INTERNAL ERROR: Data for partition to activate is invalid."
         end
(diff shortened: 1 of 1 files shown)


src/lib/bootloader/mbr_update.rb
```diff
@@ -206,25 +206,24 @@ def partition_to_activate(loader_device)
...
+      if device
         # (bnc # 337742) - Unable to boot the openSUSE (32 and 64 bits) after installation
         # if loader_device is disk Choose any partition which is not swap to
         # satisfy such bios (bnc#893449)
-        partition = activatable_partitions(mbr_dev).first
+        partition = activatable_partitions(device).first
         log.info "loader_device is disk device, so use its partition #{partition.inspect}"
       else
-        partition = device
+        partition = nil
```
jreidinger: this does not work. method should for partition return partition, in your case it will return nil as I expect partition is not partitionable. So I suggest to keep old code and simply instead of `device.is?(:disk)` use `device.is?(:partitionable)`

wfeldt: There's an 'if partition' in the function calling this method. So clearly the assumption is that the function should return nil if something goes wrong. Also, returning a class != Partition will not work as typical Partition methods are used on the result.

It would be different if there were a **logical** reason to return the main device but I couldn't think of any.

jreidinger: no, it should return Partition class. Just if dev_name is partition name, it should return Partition class and not nil.

calling method have if partition, for devices that is not partition, but data is taken from boot devices and it can look like /dev/sda, /dev/sda1 and in /dev/sda1 case, this method have to return Partition class and not nil.

wfeldt: Ah, you mean `#to_partition` can also receive a partition and then it should return it. Ok.

jreidinger: yeah, exactly, sorry for confusing you

PR #453, opened 2017-08-07
release 3.4.2

wfeldt:


ancorgs:
So far, in libstorage-ng we don't need to increase the package versions (except yast2-storage-ng if we change the API) since we don't submit this anywhere. Isn't it?

ancorgs:
s/in libstorage-ng/in the storage-ng branches/

wfeldt:
Well, I don't see the changes submitted. A jenkins problem, then?

jreidinger:
@wfeldt one think is that changes are too long, second problem is missing bnc or fate number for submission. ( I suggest to use storage-ng FATE number )

Second part is that it should be submitted even without changes entry. I will check jenkins and I see this submit as last one - https://ci.opensuse.org/view/Yast/job/yast-bootloader-storage-ng/31/console

if it is not the last change, then probably polling somehow failed or github hook.

wfeldt:
I don't see any bnc or fate numbers in the last submissions, so I don't think this is a problem.

The last poll run on Aug. 2; no idea if that's a problem on the jenkins or github side.

jreidinger:
@wfeldt I think it is no longer needed, right?

wfeldt:
yes, it's history

PR #454, opened 2017-08-10
bsc#1014167: allow installation without grub

gilsonsouza:


SUPPORTED_SCENARIOS.md
```diff
@@ -58,3 +58,7 @@ The goal of this document is to have a single source of information  which scena
 * /boot
 * extended partition
   * /boot has to be on a logical partition
+  
+# Required software
+
+* During installation, yast-bootloader requires installing the packages kexec_tools and grub2. If one decides for not installing grub in MBR <b>and</b> "/" then the system can be installed without these packages.
```
jreidinger: this is actually even more tricky. There is more tools needed depending on scenario. I prefer to have here enlisted what packages is in which scenario required.

src/lib/bootloader/proposal_client.rb
```diff
@@ -226,7 +224,23 @@ def single_click_action(option, value)
...
     end
+
+    def update_required_packages
+      bl = ::Bootloader::BootloaderFactory.current
+      bootloader_resolvables = Yast::PackagesProposal.GetResolvables("yast2-bootloader", :package)
+
+      if (bl.name == "none" || bl.stage1.devices.empty?) &&
+          !bootloader_resolvables.empty?
+        log.info "packages #{bl.packages} are no longer required"
+        Yast::PackagesProposal.RemoveResolvables("yast2-bootloader", :package, bl.packages)
```
jreidinger: this is wrong. It should remove `bootloader_resolvables` and not bl.packages, which for `none` should be empty.

test/bootloader_finish_client_test.rb
```diff
@@ -118,5 +119,24 @@
         subject.write
       end
     end
+
+    context "when there is no device to install grub" do
```
jreidinger: well, if we allows to install grub without stage1, does it have warning in such case? as usually it does not work, unless user do own changes.

SUPPORTED_SCENARIOS.md
gilsonsouza: Are the scenarios "grub", "grub-efi" (depending on architecture), and "none" (upgrade scenario)?

jreidinger: see e.g. grub2-efi packages, which require also shim if secure boot is used https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/grub2efi.rb#L94

and on other hand kexec-tools is not always installed see https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/bootloader_base.rb#L56

test/bootloader_finish_client_test.rb
gilsonsouza: There is a warning showing in red in the interface.

src/lib/bootloader/finish_client.rb
```diff
@@ -43,7 +43,7 @@ def write
       end
 
       # we do not manage bootloader, so relax :)
-      return true if bl_current.name == "none"
+      return true if bl_current.name == "none" || bl_current.stage1.devices.empty?
```
jreidinger: do you test it with EFI? EFI does not have stage1. In general, I am not sure if this check is good idea. Bug is about selecting "none", so deselection lower looks reasonable, but not this additional for empty stage1.

src/lib/bootloader/proposal_client.rb
```diff
@@ -226,7 +224,23 @@ def single_click_action(option, value)
...
+      elsif bootloader_resolvables.empty? && bl.name != "none" && !bl.stage1.devices.empty?
+        log.info "propose to install #{bl.packages}"
+        Yast::PackagesProposal.AddResolvables("yast2-bootloader", :package, bl.packages)
+      end
```
jreidinger: in general why does not use simple

```
Yast::PackagesProposal.RemoveResolvables("yast2-bootloader", :package, bootloader_resolvables)
Yast::PackagesProposal.AddResolvables("yast2-bootloader", :package, bl.packages)
```

for all scenarios? this way it will also fix switching between variants with different packages and it is clear code. Deselect old ones and select new packages.

SUPPORTED_SCENARIOS.md
jreidinger: scenarios in fact depends on configuration, so it is more like "shim if secure boot is used" but it is only for grub2efi and only on x86_64...so in general I suggest to split it to those three and then check code in which case what is installed. Logic is quite straighforward and can be found in `def packages` methods.

SUPPORTED_SCENARIOS.md
```diff
@@ -58,3 +58,21 @@ The goal of this document is to have a single source of information  which scena
 * /boot
 * extended partition
   * /boot has to be on a logical partition
+  
+# Required packages
+
+As we've previously mentioned, we have 3 options for bootloaders: grub2, grub2-efi, and none. The bootloader option, the system configuration, and its architecture will define the required packages. Besides that, the system will always require the package <b>kexec-tools</b> unless the installation is happening through a live medium.
+
+## grub2
+This is the most common option and requires only <b>grub2</b>.
```
jreidinger: in fact here it also depends on architecture, similar to grub2-efi

src/lib/bootloader/proposal_client.rb
```diff
@@ -227,5 +222,13 @@ def single_click_action(option, value)
 
       Yast::Bootloader.proposed_cfg_changed = true
     end
+
+    def update_required_packages
+      bl = ::Bootloader::BootloaderFactory.current
+      bootloader_resolvables = Yast::PackagesProposal.GetResolvables("yast2-bootloader", :package)
+      log.info "proposed packages to install #{bl.packages}"
+      Yast::PackagesProposal.RemoveResolvables("yast2-bootloader", :package, bootloader_resolvables)
```
lslezak: Instead of `RemoveResolvables(old)` + `AddResolvables(new)` you could use one `SetResolvables(new)` which replaces the current setting. See https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/PackagesProposal.rb#L134-L159

PR #455, opened 2017-08-11
Merge branch 'storage-ng' into 'master'

wfeldt:


.rubocop.yml
```diff
@@ -124,3 +124,12 @@ Style/VariableName:
...
   EnforcedStyle: prefer_alias_method
+
+# This two are not equivalent if using refinements (and we use them to add
+# functionality to the classes coming from libstorage)
+#   using RefinementAddingFooToThing
+#   things.map { |s| s.foo } # it works
+#   things.map(&:foo) # it does not work
+# See https://github.com/bbatsov/rubocop/issues/2720
+Style/SymbolProc:
+  Enabled: false
```
jreidinger: probably not needed anymore. @ancorgs ?

.travis.yml
```diff
@@ -6,7 +6,7 @@ services:
 before_install:
   - docker build -t yast-bootloader-image .
 script:
-  # the "yast-travis-ruby" script is included in the base yastdevel/ruby image
-  # see https://github.com/yast/docker-yast-ruby/blob/master/yast-travis-ruby
-  - docker run -it -e TRAVIS=1 -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" yast-bootloader-image yast-travis-ruby
+  # the "storage-ng-travis-ruby" script is included in the base yastdevel/storage-ng image
+  # see https://github.com/yast/docker-storage-ng/blob/master/storage-ng-travis-ruby
+  - docker run -it -e TRAVIS=1 -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" yast-bootloader-image storage-ng-travis-ruby
```
jreidinger: do we still need storage-ng travis? I expect after merge it is not needed.

Dockerfile
```diff
@@ -1,2 +1,2 @@
-FROM yastdevel/ruby
+FROM yastdevel/storage-ng
```
jreidinger: same here as above

package/yast2-bootloader.spec
```diff
@@ -66,6 +68,7 @@ This package contains the YaST2 component for bootloader configuration.
 %setup -n %{name}-%{version}
 
 %check
+# storage-ng
```
jreidinger: I do not get it

src/lib/bootloader/device_map.rb
```diff
@@ -118,14 +119,24 @@ def order_boot_device
...
     def fill_mapping
+      # storage-ng
+      # BIOS-ID is not supported in libstorage-ng, so let's simply create a
+      # mapping entry per disk for the time being (see commented code for the
+      # real expected behavior)
+      staging = Y2Storage::StorageManager.instance.staging
+      staging.disks.each_with_index do |disk, index|
+        add_mapping("hd#{index}", disk.name)
+      end
+# rubocop:disable Style/BlockComments
```
jreidinger: please remove whole this commented out code, as we do not plan to implement it, so lets use just common order.

src/lib/bootloader/grub2efi.rb
```diff
@@ -31,19 +32,28 @@ def read
...
+
+        efi_partition = efi_partition.plain_blk_devices.first
+
+        efi_disk = efi_partition.disk
+
+# storage-ng
+# rubocop:disable Style/BlockComments
+=begin
         # get underlaying disk as it have to be set there and not on virtual one (bnc#981977)
         device = ::Bootloader::Stage1Device.new(efi_disk)
```
jreidinger: I think this is not longer needed, as above is used "disk" method so it should be physical device, so can be removed.

src/lib/bootloader/mbr_update.rb
```diff
@@ -149,7 +152,14 @@ def disks_to_rewrite
 
     def first_base_device_to_boot(md_device)
       md = ::Bootloader::Stage1Device.new(md_device)
+      # storage-ng
+      # No BIOS-ID support in libstorage-ng, so just return first one
+      md.real_devices.first
+# rubocop:disable Style/BlockComments
```
jreidinger: fine for me to remove it.

src/lib/bootloader/mbr_update.rb
```diff
@@ -167,64 +177,73 @@ def bios_id_for(device)
...
+    # Raises an exception if no suitable partition to activate was found.
+    #
+    # @param loader_device [String] the device to install the bootloader to
+    #
+    # @return [Y2Storage::Partition]
+    #
     def partition_to_activate(loader_device)
+      # storage-ng
+      # FIXME
+      # going through 'real' device(s) here is almost certainly wrong; atm the
```
jreidinger: well, tricky part is that we still does not use udev device names here. So if we passing around storage object and not device name, it can just work.

.travis.yml
wfeldt: currently it is, as storage-ng is not yet available

Dockerfile
wfeldt: dto, travis will just fail otherwise

src/lib/bootloader/stage1_device.rb
```diff
@@ -53,61 +56,66 @@ def underlaying_devices_for(dev)
...
-        if part["type"] == :lvm
-          lvm_group = Yast::Storage.GetDisk(tm, disk_data["disk"])
-          return lvm_underlaying_devices(lvm_group)
-        elsif part["type"] == :sw_raid
-          return devices_on(part)
-        elsif part["fstype"] == Yast::Partitions.dmraid_name
-          mapper = Yast::Storage.GetDisk(tm, disk_data["disk"])
-          return mapper["devices"]
-        end
+      # FIXME: there is nothing like this right now in libstorage-ng
```
jreidinger: I think this is already implemented in yast2-storage-ng, right @ancorgs ?

src/lib/bootloader/stage1_device.rb
```diff
@@ -53,61 +56,66 @@ def underlaying_devices_for(dev)
...
+      # DMRAID main device - not implemented in libstorage-ng
+      # return disk["devices"]
+
+      # TODO: storage-ng
+      # DMRAID part device - not implemented in libstorage-ng
+      # return mapper["devices"]
+      #
+      # TODO: storage-ng
+      # sw_raid
+      # return devices_on(part)
```
jreidinger: well, it is already implemented in yast2-storage, so maybe it make sense to implement it?

src/lib/bootloader/stage1_device.rb
wfeldt: Josef, this is just about merging the current state, not fixing issues in the code.

src/lib/bootloader/udev_mapping.rb
```diff
@@ -49,6 +48,9 @@ def to_mountby_device(dev)
 
       log.info "#{dev} looked as kernel device name: #{kernel_dev}"
 
+# storage-ng
+# rubocop:disable Style/BlockComments
```
jreidinger: well, this is regression as we should use udev names. So please add FIXME

.rubocop.yml
ancorgs: Shouldn't be needed anymore, indeed. But not sure if removing the exception from here would also imply fixing the code in some parts.

src/lib/bootloader/udev_mapping.rb
ancorgs: All the `# storage-ng` comments are kind of FIXMEs, according to this https://github.com/yast/yast-storage-ng/blob/master/doc/installer-hacks.md#temporarily-commented-code

src/lib/bootloader/stage1_device.rb
ancorgs: Yes, please review the merge (like travis configuration, resulting changelog, etc.) but don't re-review the code. Merging is the important thing in this PR. The rest can be addressed, if you feel in fixing-stuff mood, in separate PRs. 😉

package/yast2-bootloader.changes
```diff
@@ -1,50 +1,98 @@
...
   to true. So libstorage can scan valid disks and "/" partition
   can be found. (bnc#1046738, bnc#1043132)
-- 3.3.0
 
 -------------------------------------------------------------------
 Wed Jun 14 14:54:38 UTC 2017 - jreidinger@suse.com
 
 - Use udev device for prep partition if it is available
   (bsc#1041692)
-- 3.2.22
```
ancorgs: As discussed in irc, loosing this kind of information or adding entries down there in the changelog is a no-go for me from the package history point of view.

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,86 @@
 -------------------------------------------------------------------
+Wed Aug  2 16:00:44 CEST 2017 - snwint@suse.de
+
+- not everything is a disk, use libstorage's partitionable instead
+
```
ancorgs: I'm missing the `3.3.1` version change here... and I can already hear some bot complaining about the lack of fate entry. I'm using `fate#318196` in other packages. Not sure if the bot will enforce us to add it to every single entry or just to the last one

jreidinger: one per diff is enough. not all entries have to have it.

package/yast2-bootloader.changes
```diff
@@ -1,4 +1,93 @@
 -------------------------------------------------------------------
+Fri Aug 11 14:14:33 UTC 2017 - snwint@suse.com
+
+- merge storage-ng branch to master
+- note: all changes below with this date belong to the merge
```
jreidinger: at least one bsc or FATE number have to be there, otherwise won't be accepted.

PR #457, opened 2017-08-22
bsc#1039712 and bsc#1052006: Find correct mbr device to  install grub

gilsonsouza:


src/modules/BootStorage.rb
```diff
@@ -125,11 +125,11 @@ def extended_partition_for(device)
...
       mp = Storage.GetMountPoints
-      boot_disk = Ops.get_string(
-        mp,
-        ["/boot", 2],
-        Ops.get_string(mp, ["/", 2], "")
+      tm = Yast::Storage.GetTargetMap
+      dev = Ops.get_string(
+        mp, ["/boot", 0], Ops.get_string(mp, ["/", 0], "")
       )
+      boot_disk = Yast::Storage.GetDisk(tm, dev)["device"]
```
shundhammer: I just tried to understand that Storage.GetDisk() function and Storage.GetDiskPartitionTg() which it calls. The latter function seems to take the partition name and from that tries to deduce the disk name, then searching the target map for a disk that also has that partition as a child. It also does not check the type of the object (it should test for CT_DISK), so it seems to work only because the names don't match otherwise (or it might find the CT_BTRFS container for /dev/btrfs as well as in the bug that this patch is about to fix).

So this does work, but entirely for the wrong reasons (which is the fault of that old storage code, not of your patch).

PR #458, opened 2017-08-24
Switch to yastdevel/ruby Docker image

imobachgs:


.travis.yml
```diff
@@ -6,7 +6,6 @@ services:
 before_install:
   - docker build -t yast-bootloader-image .
 script:
-  # the "storage-ng-travis-ruby" script is included in the base yastdevel/storage-ng image
-  # see https://github.com/yast/docker-storage-ng/blob/master/storage-ng-travis-ruby
-  - docker run -it -e TRAVIS=1 -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" yast-bootloader-image storage-ng-travis-ruby
-  - docker run -it yast-bootloader-image rake check:doc
+  # the "yast-travis-ruby" script is included in the base yastdevel/ruby image
+  # see https://github.com/yast/docker-yast-ruby/blob/master/yast-travis-ruby
+  - docker run -it -e TRAVIS=1 -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" yast-bootloader-image yast-travis-ruby
```
mchf: missing ```check:doc``` is it intentional?

imobachgs: Yes, it is already done in the [yast-travis-ruby](https://github.com/yast/docker-yast-ruby/blob/master/yast-travis-ruby#L22) script.

imobachgs: Yes, it is already done in the [yast-travis-ruby](https://github.com/yast/docker-yast-ruby/blob/master/yast-travis-ruby#L22) script.

mchf: ok

imobachgs:
Thanks!

.travis.yml
jreidinger: @imobachgs no, it is different. In standard image is called only yardoc and it does not fail if there is any warnings or coverage is lower then defined. `rake check:doc` fail if there is any yardoc warning or documentation coverage is lower then defined, so cannot be overlooked.

PR #461, opened 2017-09-05
Test enablement

jreidinger:
more test re enablement after storage-ng and some cleanup of no longer used stuff


imobachgs:
Nice clean-up :) LGTM

PR #462, opened 2017-09-08
Fix size of device map dialog

jreidinger:
it by default extend size of dialog. see old and new screenshots 

## Old
snimek obrazovky_2017-09-08_17-02-42

snimek obrazovky_2017-09-08_16-52-38

## New

snimek obrazovky_2017-09-08_17-03-26

snimek obrazovky_2017-09-08_16-51-50


lslezak:
Use the `TERM=xterm-color` workaround: https://bugzilla.suse.com/show_bug.cgi?id=1054448

PR #463, opened 2017-09-14
Udev fix

jreidinger:
note: udev mapping is almost new file, so make sense to review it from scratch.

### DIFF
--- a/src/lib/bootloader/udev_mapping.rb
+++ b/src/lib/bootloader/udev_mapping.rb
@@ -5,7 +5,6 @@ require "bootloader/exceptions"
 require "y2storage"
 
 Yast.import "Mode"
-Yast.import "Arch"
 
 module Bootloader
   # Class manages mapping between udev names of disks and partitions.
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/udev_mapping.rb
```diff
@@ -38,8 +38,26 @@ def to_kernel_device(dev)
       end
     end
 
-    # Converts udev or kernel device (disk or partition) to udev name according to mountby
-    # option or kernel device if such udev device do not exists
+    # Converts udev or kernel device (disk or partition) to udev name that fits best.
+    #
+    # Here is description of strategy for finding the best possible udev persistent name.
+    # There is three scenarious we consider:
```
ancorgs: There are three scenarios we consider:

src/lib/bootloader/udev_mapping.rb
```diff
@@ -38,8 +38,26 @@ def to_kernel_device(dev)
...
+    # S1. disk with boot configuration is moved to different PC
+    # S2. disk dies and its content is loaded to new disk from backup
+    # S3. path to disk dies and disk is moved to diffent one
```
ancorgs: typo: diffent

src/lib/bootloader/udev_mapping.rb
```diff
@@ -38,8 +38,26 @@ def to_kernel_device(dev)
...
+    #
+    # Strategy is:
+    #
+    # 1. if device have filesystem and it have its mount_by, then respect it
+    # 2. if there is by-label use it, as it allows to handle S1, S2 and S3 just with using same
+    #    label
+    # 3. if there is by-uuid then use it as it can also handle S1, S2 and S3 as uuid can be
+    #    changed, but it is harder to do
+    # 4. if there is by-id use it, as it can handle S3 in some scenarious, but not always.
```
ancorgs: scenarios

test/udev_mapping_test.rb
```diff
@@ -4,50 +4,21 @@
...
-xdescribe Bootloader::UdevMapping do
+describe Bootloader::UdevMapping do
   subject { Bootloader::UdevMapping }
   before do
-    # always invalidate cache to use new mocks
-    allow(subject.instance).to receive(:cache_valid?).and_return false
-    allow(Yast::Arch).to receive(:ppc).and_return(false)
-    allow(Yast::Storage).to receive(:GetContVolInfo).and_return(false)
+    # always call proper method
     allow(Bootloader::UdevMapping).to receive(:to_kernel_device).and_call_original
```
ancorgs: Please, remind me why we have these `allow(X).to receive(:y).and_call_original` here. Is to "disable" some mocking coming from the test_helper? In that case, "always call proper method" is not clear and obvious enough to me.

jreidinger: yes, exactly. test_helper mock it by default, so I want to be sure, here is called real method. I will udpate comment

PR #464, opened 2017-09-15
Fix udev mapping

jreidinger:


package/yast2-bootloader.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Sep 15 11:54:25 UTC 2017 - jreidinger@suse.com
+
+- always use for underlaying devices detection kernel name
```
ancorgs: I had to leave the code to understand the sentence. Reordering words should make it more readable.
"Use always kernel name for underlaying devices detection"

PR #466, opened 2017-10-26
bsc#1059757 and bsc#1060172: filter out multipath wires and change dialog to add disks

gilsonsouza:


src/lib/bootloader/udev_mapping.rb
```diff
@@ -79,7 +79,7 @@ def udev_to_kernel(dev)
       # in mode config if not found, then return itself
       return dev if Yast::Mode.config
 
-      devices = Y2Storage::BlkDevice.all(staging)
+      devices = staging.disk_devices
```
jreidinger: Tests failed. Main problem is that disk_devices gets only disk and similar devices, but we need all devices including also partitions, that is why tests failed, as it tries to get kernel name for partition.

jreidinger:
I am also missing test case for problematic scenario. Feel free to add new yaml file with definition or even xml one and add test for it.

src/lib/bootloader/udev_mapping.rb
jreidinger: in general we need something like `staging.find_by_udev_name` or at least something like `staging.devices_with_udev_links`

src/lib/bootloader/udev_mapping.rb
```diff
@@ -80,8 +80,8 @@ def udev_to_kernel(dev)
       return dev if Yast::Mode.config
 
       devices = Y2Storage::BlkDevice.all(staging)
+      devices.reject! { |d| d.is?(:disk) && d.descendants.any? { |i| i.is?(:multipath) } }
```
jreidinger: this needs at least some comment why remove such disks from udev mapping.

src/lib/bootloader/udev_mapping.rb
```diff
@@ -80,8 +80,9 @@ def udev_to_kernel(dev)
       return dev if Yast::Mode.config
 
       devices = Y2Storage::BlkDevice.all(staging)
+      # removes possible multipath wires from the list since they should not appear as disk devices
```
jreidinger: Hmm, this explanation looks wrong for me. This method is only about translating udev device received to kernel name. Only reason I see to filter it out is that this multipath wires have identical udev device as its mutlitpath device and we want to translate that udev device to multipath device and not wire. So if I am true, this should be described here.

gilsonsouza: do you think that the following comment fits better?
"As wires devices have identical udev devices as its multipath device, we remove possible multipath wires from the list since we want to translate udev device to multipath device and not wire"

jreidinger: yes, that is perfect one.

PR #475, opened 2017-12-19
working RAID

jreidinger:
https://bugzilla.suse.com/show_bug.cgi?id=1072908

### Fix
    
Legacy (non-EFI) x86: Fixed multi-device booting problems (mostly RAID, also LVM and multipath). Now the boot loader is written to ALL physical devices that host the boot partition (or the MBR). Before, only one of them was picked and the BIOS could pick a different one to boot from.

Related UI change: "Boot Loader Location" has been simplified. We kept the option "Boot from Master Boot Record", but the remaining 3 options, for booting from the Boot, Root, or Extended partitions, have been unified to a single option "Boot from Partition".

### Code

Before, the code remembered the "boot partition" and installed stage 1 to one of its underlying devices.

Now, it remembers the "boot mount point" (which is / or /boot) and uses #stage1disk or #stage1partition (according to a UI choice) to get ALL underlying devices for it (as an Array).

### Documenting this

The main idea is implemented in BootStorage#stage1...for..., see
- the comments at: https://github.com/yast/yast-bootloader/blob/39e6ac67f6f9df2ea9fdb8b72bde925ae929f532/src/modules/BootStorage.rb#L124-L130 
- the commit message for https://github.com/yast/yast-bootloader/commit/a50da5512634d01b376a863280071f8dc3d7fdb0


src/lib/bootloader/grub2.rb
```diff
@@ -184,34 +184,18 @@ def disk_order_summary
...
-        # TRANSLATORS: extended is here for extended partition. Keep translation short.
-        locations << Yast::BootStorage.extended_partition.name + _(" (extended)")
-      end
+      partition_location = Yast::BootStorage.boot_partitions.map(&:name).join(", ")
+      locations << partition_location + _(" (/boot)") unless partition_location.empty?
       if stage1.mbr?
         # TRANSLATORS: MBR is acronym for Master Boot Record, if nothing locally specific
         # is used in your language, then keep it as it is.
-        locations << Yast::BootStorage.mbr_disk.name + _(" (MBR)")
+        locations << Yast::BootStorage.boot_disks.map(&:name) + _(" (MBR)")
```
ancorgs: beware, you are trying to Array + String here

jreidinger: good catch, thanks. I plan to do some intensive testing and hopefully also testsuite will catch it.

ancorgs:
Overall, the idea looks sane. But after two meetings, my brain is not in best conditions for a more precise review.

mvidner:
OK, this is a big change, and after skimming the diff I still don't know where to start. So far, while looking for a high-level entrance, I've found this:

### Changes

Thanks for the summary in the PR description. We need to write that to an anppropriate place in code comments.

Let's change this description to make it accurate:

> RAID booting fixed. Before, the system would not boot if you had any partition on any kind of RAID (hardware, software, RAID0, RAID1, RAID2018). Now it works for any kind except World of Warcraft raids. (bsc#000000)

> UI changes: "Boot Loader Location" has been simplified. We kept the option "Boot from Master Boot Record", but the remaining 3 options, for booting from the Boot, Root, or Extended partitions, have been unified to a single option "Boot from Partition".

### Tests

Which of them are "mock adjustments to make it work" and which are "actual tests of the fixes/new features"?

### Code

I thought I would start looking at the added classes but there aren't any :) A class was deleted (Stage1Device) and more methods were deleted than added.

I am suspicious: a substantial fix becomes so well dispersed in the changed code that it is almost invisible. Should we add comments to some methods? Like "This is what makes RAID work". Maybe even factor out a class??

mvidner:
Here are the answers I got from Pepa in an in-person explanation:

### The Fix

https://bugzilla.suse.com/show_bug.cgi?id=1072908

> Legacy (non-EFI) x86: Fixed multi-device booting problems (mostly RAID, also LVM and multipath). Now the boot loader is written to ALL physical devices that host the boot partition (or the MBR). Before, only one of them was picked and the BIOS could pick a different one to boot from.

> Related UI change: "Boot Loader Location" has been simplified. We kept the option "Boot from Master Boot Record", but the remaining 3 options, for booting from the Boot, Root, or Extended partitions, have been unified to a single option "Boot from Partition".

### Code

Before, the code remembered the "boot partition" and installed stage 1 to one of its underlying devices.

Now, it remembers the "boot mount point" (which is / or /boot) and uses #stage1disk or #stage1partition (according to a UI choice) to get ALL underlying devices for it (as an Array).

### Tests

So far they are just adjustments of existing tests. The central methods are covered because higher level tests call them. Also, openQA tests Legacy boot a lot.

src/lib/bootloader/grub2.rb
```diff
@@ -184,34 +184,18 @@ def disk_order_summary
...
       locations = []
 
-      partition_location = boot_partition_location
-      locations << partition_location unless partition_location.empty?
-      if stage1.extended_partition?
-        # TRANSLATORS: extended is here for extended partition. Keep translation short.
-        locations << Yast::BootStorage.extended_partition.name + _(" (extended)")
-      end
+      partition_location = Yast::BootStorage.boot_partitions.map(&:name).join(", ")
+      locations << partition_location + _(" (/boot)") if stage1.boot_partition?
```
mvidner: IMHO this string should not be translated

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -149,16 +149,9 @@ def import_stage1(data, bootloader)
...
-            data["location"] == "extended") &&
-            Yast::BootStorage.extended_partition.nil?
-          raise "boot_extended used in autoyast profile, but there is no extended partition"
-        end
-
-        STAGE1_DEVICES_MAPPING.each do |key, device|
+        STAGE1_DEVICES_MAPPING.each do |key, method|
           if data["global"][key] == "true" || data["boot_#{key}"]
-            stage1.add_udev_device(Yast::BootStorage.public_send(device).name)
+            stage1.add_udev_device(stage1.public_send(method))
```
mvidner: This new line is not covered by tests.

Can this PR break previously valid AY profiles?

src/modules/BootStorage.rb
```diff
@@ -213,36 +111,123 @@ def available_swap_partitions
...
+      partitions.map! { |p| extended_for_logical(p) }
+      partitions.uniq!
+
+      log.info "stage1 partitions for #{device.inspect} is #{partitions.inspect}"
+
+      partitions
+    end
+
+    # If passed partition is logical, return extended, otherwise return argument
+    def extended_for_logical(partition)
```
mvidner: Not covered by tests, and neither is `extended_partition`.

src/lib/bootloader/autoyast_converter.rb
jreidinger: no. In fact as you can see in module we no longer export this value ( we do not export it even in SLE12 ). It is used mainly as workaround when bootloader propose wrong value and user want explicitly overwrite it.

But I can try to add test for this code.

src/lib/bootloader/grub2.rb
jreidinger: are you sure that `(` and `)` is same in all languages? because e.g. in czech `"` is not same and depends if it is opening or closing one.

src/modules/BootStorage.rb
jreidinger: I will check if I have proper storage data with logical partition.

src/lib/bootloader/autoyast_converter.rb
jreidinger: test added and code fixed

src/modules/BootStorage.rb
jreidinger: test added

src/modules/BootSupportCheck.rb
```diff
@@ -101,11 +101,11 @@ def correct_loader_type(lt)
     def check_gpt_reserved_partition
       return true unless stage1.mbr?
 
-      disk = BootStorage.mbr_disk
-      boot_device = BootStorage.boot_partition
-      return true unless disk.gpt?
-      return true if boot_device.filesystem_type != ::Y2Storage::Filesystems::Type::BTRFS
-      return true if disk.partitions.any? { |p| p.id.is?(:bios_boot) }
+      boot_fs = BootStorage.boot_mountpoint
```
mvidner: the rest of the method is changed but untested

src/lib/bootloader/grub2.rb
```diff
@@ -225,28 +209,15 @@ def mbr_line
...
-          _(
-            "Do not install bootcode into /boot partition " \
-              "(<a href=\"enable_boot_boot\">install</a>)"
-          )
-        end
-      elsif stage1.root_partition?
+      if stage1.boot_partition?
         _(
-          "Install bootcode into \"/\" partition " \
-            "(<a href=\"disable_boot_root\">do not install</a>)"
```
mvidner: ..._boot_root is still present in src/lib/bootloader/proposal_client.rb, remove it?

src/lib/bootloader/grub2.rb
```diff
@@ -225,28 +209,15 @@ def mbr_line
...
+          "Install bootcode into partition " \
```
mvidner: "Install boot code into a partition"
Should we also tell the user here *which* partition it is?

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -708,32 +708,33 @@ def handle(event)
...
+          stage1.boot_disk_names.each { |d| stage1.add_udev_device(d) }
+        end
       end
 
       return unless Yast::UI.QueryWidget(:custom, :Value)
 
       devs = Yast::UI.QueryWidget(:custom_list, :Value)
       devs.split(",").each do |dev|
-        stage1.add_udev_device(dev.strip)
+        # add it exactly how user specify
```
mvidner: The user specifi**es**. Anyway:
> Add it exactly as specified by the user

src/modules/BootStorage.rb
```diff
@@ -27,36 +27,12 @@ module Yast
   class BootStorageClass < Module
     include Yast::Logger
 
-    # Disk where to place MBR code. By default one with /boot partition
-    # @return [Y2Storage::Disk]
-    def mbr_disk
+    # Moint point for /boot. If there is not separated /boot, / is used instead.
```
mvidner: When I hear "mount point" I think of a filesystem path, a String. But this method returns a Filesystem, whatever that is. All callers assign it to a variable named `fs`. Rename it to `boot_filesystem`?

src/lib/bootloader/grub2.rb
jreidinger: right

src/lib/bootloader/grub2.rb
jreidinger: it will be mentioned few lines above in installation destination when user decide to install it there. I worry it can be quite long here.

src/modules/BootStorage.rb
jreidinger: yes, makes sense. I start with this one and later prefer using filesystem instead.

src/lib/bootloader/grub2.rb
mvidner: Hum, you have a point, https://www.mediawiki.org/wiki/Localisation#Symbols,_colons,_brackets,_etc._are_parts_of_messages I'll have to look closer.

src/modules/BootStorage.rb
```diff
@@ -213,36 +111,119 @@ def available_swap_partitions
...
+    # If passed partition is logical, return extended, otherwise return argument
+    def extended_for_logical(partition)
+      partition.type.is?(:logical) ? extended_partition(partition) : partition
+    end
+
+    # get stage1 device suitable for stage1 location
+    # @param [Y2Storage::Device] device to check
+    # @return [Array<Y2Storage::Device] devices suitable for stage1
+    def stage1_disks_for(device)
+      disks = ([device] + device.ancestors + device.descendants).select { |a| a.is?(:disk) }
```
mvidner: Why *both* ancestors and descendants? Isn't that wrong? An example please?

src/modules/BootStorage.rb
```diff
@@ -142,31 +142,48 @@ def stage1_device_for_name(dev_name)
...
-    # If passed partition is logical, return extended, otherwise return argument
+    # In a device graph, starting at *device* (inclusive), find the parents
+    # that match *predicate*.
+    # NOTE that once the predicate matches, the search stops **for that node**
+    # but continues for other nodes.
+    # @param device [Y2Storage::Device] starting point
+    # @yieldparam [Y2Storage::Device]
+    # @yieldreturn [Boolean]
+    # @return [Array<Y2Storage::Device]
+    def find_ancestors(device, &predicate)
```
jreidinger: better name will be `select_ancestors`

jreidinger: and second note, as it is internal helper only, please move it to internal private section of class.

mvidner:
@jreidinger Please check whether https://github.com/yast/yast-bootloader/blob/master/SUPPORTED_SCENARIOS.md is still valid, I cannot tell.

src/lib/bootloader/grub2.rb
mvidner: In practice (in yast-translations) we don't have a language that would use non-ASCII parentheses so this is fine.

jreidinger:
@mvidner checked, still good from supportability POV.

src/lib/bootloader/grub2.rb
mvidner: (done)

src/modules/BootStorage.rb
mvidner: (done)

DimStar77:
Commenting here, so I won't forget till Monday: this yast2-bootloader is now in Staging:E, and the first observations are in:

The bootloader is now reported to be installed on the MBR in a standard install, no longer to / partition

Compare openQA screenshots:

* old:   https://openqa.opensuse.org/tests/574935#step/installation_overview/5
* new: https://openqa.opensuse.org/tests/575133#step/installation_overview/5

jreidinger:
@DimStar77 yes, part of this PR is that we change default boot location to MBR unless we know we will have problem to install it there.

PR #476, opened 2018-01-09
Use the new Devicegraph#find_by_any_name

ancorgs:
See https://bugzilla.suse.com/show_bug.cgi?id=1073254

Depends on https://github.com/yast/yast-storage-ng/pull/479

I manually tested the 13.1 -> TW upgrade including the three patches (this, https://github.com/yast/yast-storage-ng/pull/479 and https://github.com/yast/yast-update/pull/91) and it worked. :+1:

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Jan  9 07:25:09 UTC 2018 - ancor@suse.com
+
+- Rely on the new Y2Storage::Devicegraph#find_by_any_name when
+  matching udev names to their corresponding kernel device names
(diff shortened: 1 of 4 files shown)


ancorgs:
Ok. I just merged the storage-ng PR. Let's wait some minutes and retry travis then

ancorgs:
I debugged the travis error using docker locally and turns out that this libstorage-ng method:
```ruby
Storage::BlkDevice.find_by_any_name(probed, "whatver crazy string")
```
Always return the /dev/sda device (which is probably the first block device).

It works as expected if the `--priviledged` option is used to run docker. So definitely the library doesn't work reliably in a non-priviledged docker container. Time to add a mock to the library call!

PR #478, opened 2018-01-16
Fix disk in partitions

jreidinger:


test/mbr_update_test.rb
```diff
@@ -134,9 +134,19 @@ def stage1(devices: [], activate: false, generic_mbr: false)
           expect(Yast::Execute).to receive(:locally)
             .with(/parted/, "-s", "/dev/sda", "set", 1, "boot", "on")
           expect(Yast::Execute).to receive(:locally)
-            .with(/parted/, "-s", "/dev/sdb", "set", 1, "boot", "on")
+            .with(/parted/, "-s", "/dev/sdb", "set", 2, "boot", "on")
 
-          subject.run(stage1(activate: true, devices: ["/dev/sda1", "/dev/sdb1"]))
+          subject.run(stage1(activate: true, devices: ["/dev/sda1", "/dev/sdb2"]))
```
ancorgs: Any reason for this change from sdb1 to sdb2? Just curious

package/yast2-bootloader.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Jan 16 16:51:54 UTC 2018 - jreidinger@suse.com
+
+- fix crash for getting partition to activate when stage1 device
+  is disk (bsc#10761210
```
ancorgs: Matching closing parenthesis not found. Buffer overflow. ;-)

test/mbr_update_test.rb
jreidinger: well, for partition it just use any partition which is in stage1, but it does not make any sense to have stage1 in swap partition.

jreidinger: so it is just to have reasonable test cases

ancorgs: Oh, is explained below

package/yast2-bootloader.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Jan 16 16:51:54 UTC 2018 - jreidinger@suse.com
+
+- fix crash for getting partition to activate when stage1 device
+  is disk (bsc#10761210)
```
ancorgs: Bug #10761210 does not exist, according to bugzilla.suse.com. Wrong copy&paste?

ancorgs: Yes, you wrote an extra 0 at the end.

test/mbr_update_test.rb
jreidinger: yeah, I do not notice it first time. I just use sdb1 in other test case and was surprised why sdb2 was picked and then inspect partitioning which makes it clear :)

package/yast2-bootloader.changes
jreidinger: ah, typo. I use 0 instead of ) and appending ) does not fix it :)

PR #482, opened 2018-02-06
Fix activate

jreidinger:


Dutchglory:
please also merge this with yast-bootloader for Leap 15 beta...

jreidinger:
@Dutchglory currently master is used for SLE15, Leap 15 and TW. We share code as long as possible, only when we have to start doing changes for Leap 15.1 we split it to own branch.

Dutchglory:
ok thanks..  :)

PR #485, opened 2018-02-15
Fix mount by

jreidinger:


src/lib/bootloader/udev_mapping.rb
```diff
@@ -121,6 +121,8 @@ def udev_name_for(device)
     def mount_by_udev(device)
       filesystem = device.filesystem
       return nil unless filesystem
+      # mount_by is nil, so not mounted and we need to use our own strategy
+      return nil if filesystem.mount_by.nil?
```
joseivanlopez: The result is the same, but maybe the code is clearer if we use `filesystem.mount_point.nil?`. Anyway, I am fine with this version.

PR #490, opened 2018-03-14
Fix alternative raid

jreidinger:


wfeldt:
Josef, could you please fix the English grammar in your comments.

jreidinger:
@wfeldt I try to improve grammar in the comments. If you have something specific in mind, can you please point me to it?

wfeldt:
I would have been happy when you just added the 3rd-person 's'.

You may not believe it but it really sticks out if one reads your comments. If I browse through yast code I immediately know 'ah, that was Josef'.

jreidinger:
@wfeldt OK, that is fair request and I can focus on that single thing.

wfeldt:
'a fair request' :-)

PR #493, opened 2018-04-24
Serial console

jreidinger:
PBI: https://trello.com/c/c66dIsRy/1417-3-15-ga-please-dont-throw-an-exception-on-valid-grubterminal-options

how it looks:
bootloader_qt_terminal
bootloader_ncurses_terminal

### DIFF
--- a/src/lib/bootloader/config_dialog.rb
+++ b/src/lib/bootloader/config_dialog.rb
@@ -19,11 +19,17 @@ module Bootloader
 
     def run
       guarded_run
-    rescue ::Bootloader::BrokenConfiguration => e
-      ret = Yast::Report.AnyQuestion(_("Broken Configuration"),
+    rescue ::Bootloader::BrokenConfiguration, ::Bootloader::UnsupportedOption => e
+      msg = if e.is_a?(::Bootloader::BrokenConfiguration)
         # TRANSLATORS: %s stands for readon why yast cannot process it
         _("YaST cannot process current bootloader configuration (%s). " \
-          "Propose new configuration from scratch?") % e.reason,
+          "Propose new configuration from scratch?") % e.reason
+      else
+        e.message
+      end
+
+      ret = Yast::Report.AnyQuestion(_("Unsupported Configuration"),
+        msg,
         _("Propose"),
         _("Quit"),
         :yes) # focus proposing new one
--- a/src/lib/bootloader/exceptions.rb
+++ b/src/lib/bootloader/exceptions.rb
@@ -26,6 +26,8 @@ module Bootloader
     end
   end
 
+  # Represent unsupported value in given option. Used mainly when value contain something that
+  # bootloader does not understand yet.
   class UnsupportedOption < RuntimeError
     include Yast::I18n
     attr_reader :option
@@ -39,7 +41,6 @@ module Bootloader
         " To use YaST2 Bootloader proposal have to be reproposed " \
         "or configuration have to be edit manually.") % option
     end
-
   end
 
   # Represents error when serial console arguments are not valid
--- a/src/lib/bootloader/exceptions.rb
+++ b/src/lib/bootloader/exceptions.rb
@@ -37,9 +37,9 @@ module Bootloader
       textdomain "bootloader"
 
       # TRANSLATORS: %s is name of option that has unsupported value
-      super _("Option '%s' contain unsupported value.\nYaST2 Bootloader cannot handle it.\n" \
-        "To use YaST2 Bootloader configuration have to be reproposed\n" \
-        "or configuration have to be edited manually.") % option
+      super _("Option '%s' contains an unrecognized value.\nYaST Bootloader cannot handle it.\n" \
+        "Either the whole bootloader configuration needs to be proposed again\n" \
+        "or YaST must quit and you can edit it manually.") % option
     end
   end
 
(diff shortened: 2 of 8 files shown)


src/modules/Bootloader.rb
```diff
@@ -146,10 +146,17 @@ def Read
         ::Bootloader::BootloaderFactory.current = ::Bootloader::BootloaderFactory.proposed
         ::Bootloader::BootloaderFactory.current.propose
       rescue ::Bootloader::BrokenConfiguration => e
-        ret = Yast::Report.AnyQuestion(_("Broken Configuration"),
+        msg = if e.is_a?(::Bootloader::BrokenConfiguration)
```
lslezak: This looks strange with the `rescue ::Bootloader::BrokenConfiguration` statement above, `e` cannot be something else. Or is `::Bootloader::UnsupportedOption` missing in the `rescue`?

src/lib/bootloader/grub2base.rb
```diff
@@ -217,11 +212,15 @@ def merge_grub_default(other)
...
-        val = other.public_send(attr)
-        default.public_send((attr.to_s + "=").to_sym, val) if val
+      begin
+        # string attributes
+        [:serial_console, :terminal, :timeout, :hidden_timeout, :distributor,
+         :gfxmode, :theme, :default].each do |attr|
+          val = other.public_send(attr)
+          default.public_send((attr.to_s + "=").to_sym, val) if val
+        end
+      rescue RuntimeError
```
lslezak: Um, I'm not sure if catching all run time exceptions is a good approach here. Maybe a comment would help here...

src/lib/bootloader/exceptions.rb
```diff
@@ -26,6 +26,23 @@ def initialize(msg)
...
+    include Yast::I18n
+    attr_reader :option
+
+    def initialize(option)
+      @reason = option
+      textdomain "bootloader"
+
+      # TRANSLATORS: %s is name of option that has unsupported value
+      super _("Option '%s' contain unsupported value.\nYaST2 Bootloader cannot handle it.\n" \
+        "To use YaST2 Bootloader configuration have to be reproposed\n" \
```
lslezak: IMHO this part is too complicated, simpler

`The configuration needs to be proposed again or edited manually.` 

would be better.

src/lib/bootloader/exceptions.rb
```diff
@@ -26,6 +26,23 @@ def initialize(msg)
...
+  class UnsupportedOption < RuntimeError
+    include Yast::I18n
+    attr_reader :option
+
+    def initialize(option)
+      @reason = option
+      textdomain "bootloader"
+
+      # TRANSLATORS: %s is name of option that has unsupported value
+      super _("Option '%s' contain unsupported value.\nYaST2 Bootloader cannot handle it.\n" \
```
lslezak: "contains an unsupported value"

src/lib/bootloader/config_dialog.rb
```diff
@@ -19,11 +19,17 @@ class ConfigDialog
...
-          "Propose new configuration from scratch?") % e.reason,
+          "Propose new configuration from scratch?") % e.reason
+      else
+        e.message
+      end
+
+      ret = Yast::Report.AnyQuestion(_("Unsupported Configuration"),
+        msg,
         _("Propose"),
         _("Quit"),
```
lslezak: "Quit" means exiting which does not make much sense during installation. The message says "repropose or edit manually" so the appropriate label could be `Edit` or something like that.

src/modules/Bootloader.rb
```diff
@@ -146,10 +146,17 @@ def Read
...
+            "Propose new configuration from scratch?") % e.reason
+        else
+          e.message
+        end
+
+        ret = Yast::Report.AnyQuestion(_("Unsupported Configuration"),
+          # TRANSLATORS: %s stands for readon why yast cannot process it
+          msg,
           _("Propose"),
           _("Quit"),
```
lslezak: Is `Quit` also OK here?

src/modules/Bootloader.rb
jreidinger: yes, it is missing.

src/lib/bootloader/grub2base.rb
jreidinger: yep, but cfa_grub2 raise generic RuntimeError, I hope it is just temporary solution. Also cfa_grub2 can be improved to raise specific exception, but currently for SLE15 GA I do not want to touch it.

src/lib/bootloader/config_dialog.rb
jreidinger: IT cannot happen during installation. It is runtime only where user edit configuration with something else then YaST2.

src/modules/Bootloader.rb
jreidinger: yes, Quit exits YaST2.

src/lib/bootloader/grub2base.rb
lslezak: OK, then just add comment with that info.

src/lib/bootloader/config_dialog.rb
jreidinger: And edit manually means edit by something other than yast2 bootloader.

PR #499, opened 2018-05-02
do not use alias in mocking

jreidinger:
it cause failure in test on s390.

Status: ready to go for SLE15


mvidner:
I don't understand this. In particular I don't understand what "alias" is.

Why is it necessary to change the YAML file?

jreidinger:
@mvidner alias stands for udev alias for device. YAML file change is necessary, as this YAML file is used for that testing and I need to distinguish those two devices by different UUID. and for testing mounting by LABEL I need to have label there. So in short I need to extend testing data to allow full testing of udev aliases.

PR #501, opened 2018-05-03
warn when autoyast have wrong loader type

jreidinger:
trello: https://trello.com/c/hjG3dkgA/1431-2-bootloader-in-autoupgrade-autoinstallation-tell-the-user-that-their-profile-is-wrong-instead-of-throwing-an-exception-please

TODO:

- [x] changes
- [x] manual test

screenshot:
snimek obrazovky_2018-05-04_14-39-44

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Fri May  4 12:43:44 UTC 2018 - jreidinger@suse.com
 
-- Inform user when bootloader is in supported in autoyast profile
-  together with allowed values (bsc#1091284)
+- Inform user when an unsupported bootloader is defined in the
+  AutoYaST profile, display the valid values (bsc#1091284)
 - 4.0.29
 
 -------------------------------------------------------------------
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri May  4 12:43:44 UTC 2018 - jreidinger@suse.com
+
+- Inform user when bootloader is in supported in autoyast profile
+  together with allowed values (bsc#1091284)
+- 4.0.29
+
+-------------------------------------------------------------------
 Wed May  2 10:35:56 UTC 2018 - jreidinger@suse.com
 
 - Fix test failure on s390 (no functionality change)(bsc#1091631)
--- a/src/lib/bootloader/auto_client.rb
+++ b/src/lib/bootloader/auto_client.rb
@@ -16,6 +16,8 @@ Yast.import "PackagesProposal"
 module Bootloader
   # Autoyast client for bootloader
   class AutoClient < ::Installation::AutoClient
+    include Yast::I18n
+
     class << self
       attr_accessor :changed
     end
@@ -32,6 +34,7 @@ module Bootloader
       begin
         Yast::Bootloader.Import(data)
       rescue Bootloader::UnsupportedBootloader => e
+        textdomain "bootloader"
         possible_values = BootloaderFactory.supported_names + ["default"]
         AutoInstall.issues_list.add(:invalid_value, "bootloader", "loader_type",
             e.bootloader_name,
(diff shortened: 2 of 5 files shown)


src/lib/bootloader/auto_client.rb
```diff
@@ -27,7 +29,18 @@ def run
     end
 
     def import(data)
-      Yast::Bootloader.Import(data)
+      begin
+        Yast::Bootloader.Import(data)
+      rescue Bootloader::UnsupportedBootloader => e
+        possible_values = BootloaderFactory.supported_names + ["default"]
```
lslezak: Um, I do not like this magic string "default", it should be a constant...

src/lib/bootloader/auto_client.rb
```diff
@@ -27,7 +29,18 @@ def run
...
+        AutoInstall.issues_list.add(:invalid_value, "bootloader", "loader_type",
+            e.bootloader_name,
+            _("Bootloader is not supported on this architecture. Possible values: ") +
```
lslezak: `Bootloader is not supported on this architecture` it a generic message saying that bootloaders do not work here. That's not true.

Proposals:

- `The selected bootloader is not supported on this architecture.`
- Or if you know what was in the profile you could include is as well:
  `The selected bootloader ('%{bootloader}') is not supported on this architecture.`

src/lib/bootloader/auto_client.rb
```diff
@@ -27,7 +31,19 @@ def run
     end
 
     def import(data)
-      Yast::Bootloader.Import(data)
+      begin
+        Yast::Bootloader.Import(data)
+      rescue Bootloader::UnsupportedBootloader => e
+        textdomain "bootloader"
+        possible_values = BootloaderFactory.supported_names + ["default"]
```
lslezak: Um, I do not like this magic string `"default"`, it should be a constant...

src/lib/bootloader/auto_client.rb
```diff
@@ -27,7 +31,19 @@ def run
...
+        AutoInstall.issues_list.add(:invalid_value, "bootloader", "loader_type",
+            e.bootloader_name,
+            _("Bootloader is not supported on this architecture. Possible values: ") +
```
lslezak: `Bootloader is not supported on this architecture` it a generic message saying that bootloaders do not work here. That's not true.

Proposals:

- `The selected bootloader is not supported on this architecture.`
- Or if you know what was in the profile you could include is as well:
  `The selected bootloader ('%{bootloader}') is not supported on this architecture.`

jreidinger: it is already passed as parameter above. So I do not want to be duplicite, so first one looks good.

src/lib/bootloader/auto_client.rb
```diff
@@ -27,7 +31,19 @@ def run
...
     def import(data)
-      Yast::Bootloader.Import(data)
+      begin
+        Yast::Bootloader.Import(data)
+      rescue Bootloader::UnsupportedBootloader => e
+        textdomain "bootloader"
+        possible_values = BootloaderFactory.supported_names + [BootloaderFactory::DEFAULT_KEYWORD]
+        AutoInstall.issues_list.add(:invalid_value, "bootloader", "loader_type",
+            e.bootloader_name,
+            _("The selected bootloader is not supported on this architecture. Possible values: ") +
```
lslezak: I just noticed it's a single line string. How is that rendered, is it displayed in a RichText? Will that fit the 80x25 ncurses screen?

src/lib/bootloader/bootloader_factory.rb
```diff
@@ -18,6 +18,8 @@ class BootloaderFactory
       "grub2-efi"
     ].freeze
 
+    DEFAULT_KEYWORD = "default".freeze
```
lslezak: IMHO A comment would nice here, form the name it's hard to guess what that means...

src/lib/bootloader/auto_client.rb
jreidinger: I have no clue, it is up to autoyast how it will show it.

schubi2: It should fit. There will be line breaks: https://trello.com/c/g5GxOY2c/1416-5-ay-implement-checking-for-firewall-part-of-the-profile

package/yast2-bootloader.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri May  4 12:43:44 UTC 2018 - jreidinger@suse.com
+
+- Inform user when bootloader is in supported in autoyast profile
```
lslezak: "Inform user when an unsupported bootloader is defined in the AutoYaST profile, display the valid values"

PR #507, opened 2018-05-08
Use none bootloader in case of nfs root boot filesystem

teclator:
[Trello Card](https://trello.com/c/siYsznJj/1433-3-help-with-nfs-root-in-mbrgap-bug)

Tested manually

nfsrootbootloader


ancorgs:
Shouldn't this go to the SLE-15-GA branch first? We want it in the media for sure

src/lib/bootloader/finish_client.rb
```diff
@@ -44,6 +44,7 @@ def write
 
       # we do not manage bootloader, so relax :)
       return true if bl_current.name == "none"
+      return true if Yast::BootStorage.boot_filesystem.is?(:nfs)
```
ancorgs: Isn't this somehow redundant with the previous condition? I mean, if `boot_filesystem.is?(:nfs)` couldn't we be quite sure than `bl_current.name == "none"` (according to `ProposalClient#make_proposal`)

teclator: It should be redundant, it was the first defensive check I added, before proposing a "none" bootloader.

teclator:
@ancorgs modified branch and removed the redundant check.

src/modules/Bootloader.rb
```diff
@@ -204,7 +204,8 @@ def Summary(simple_mode: false)
       # kokso: additional warning that root partition is nfs type -> bootloader will not be installed
       if BootStorage.boot_filesystem.is?(:nfs)
         log.info "Bootloader::Summary() -> Boot partition is nfs type, bootloader will not be installed."
-        return _("The boot partition is of type NFS. Bootloader cannot be installed.")
+        msg = _("The boot partition is of type NFS. Bootloader cannot be installed.")
+        return (simple_mode ? [msg] : [Yast::HTML.Colorize(msg, "red")])
```
jreidinger: why having it red? It is not warning. It is correct proposal for NFS.

src/modules/Bootloader.rb
```diff
@@ -204,7 +204,8 @@ def Summary(simple_mode: false)
       # kokso: additional warning that root partition is nfs type -> bootloader will not be installed
       if BootStorage.boot_filesystem.is?(:nfs)
         log.info "Bootloader::Summary() -> Boot partition is nfs type, bootloader will not be installed."
-        return _("The boot partition is of type NFS. Bootloader cannot be installed.")
+        msg = _("The boot partition is of type NFS. Bootloader cannot be installed.")
```
jreidinger: I would rephrase it like `"The boot partition is of type NFS. No bootloader is installed."` Or if you want to have it red and with warning, then I would write something like `"The boot partition is of type NFS. No bootloader is installed. Please ensure that your machine can properly boot from network."` So at least user have hint, but for me it is not needed, as if user want nfs root ( not trivial to setup ), then he knows what he is doing.

src/modules/Bootloader.rb
teclator: I agree with no highlight but definitely should be shown. I just used the same warning was defined in the NoneBootloader [Summary](https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/none_bootloader.rb#L11)

src/modules/Bootloader.rb
teclator: Just used the same message, but also fine about rewording it.

ancorgs: If the original message is already translated, then I wouldn't change it

PR #508, opened 2018-05-09
Merge SLE-15-GA into master

teclator:
It merges #507


src/lib/bootloader/proposal_client.rb
```diff
@@ -32,6 +32,10 @@ def initialize
     ].freeze
 
     def make_proposal(attrs)
+      if Yast::BootStorage.boot_filesystem.is?(:nfs)
+        ::Bootloader::BootloaderFactory.current_name = "none"
+        return construct_proposal_map
+      end
```
jreidinger: now when I see it I see few problems with this early exit.

1. in update we should still use line 56
2. as you mention going back and forward, it really won't work as it cache old result and what is worse it will SEGFAULT as storage objects no longer will work.

When I see it, my idea is to move this proposal to - https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/bootloader_factory.rb#L94 where bootloader factory propose bootloader type. so in the end it really propose None for NFS and on other hand keep other functionality.

teclator: @jreidinger taking in account that NFS seems like an unsupported scenario for offline updating and that we tested back and forth with success do you still expect some adaptation here?

jreidinger: no, it is fine as it is.

PR #509, opened 2018-05-09
Fix bios raid 15

jreidinger:
trello: https://trello.com/c/XUunQlOe/1438-3-sle-15-ga-grub2-installation-failed-on-intel-rste-in-raid0-mode

Successfully tested by customer with respective hardware.


src/modules/BootStorage.rb
```diff
@@ -192,10 +192,13 @@ def stage1_disks_for(device)
       # Eg. 2 Disks are parents of 1 Multipath, the disks are just "wires"
       # to the real disk.
       multipaths = component.select { |a| a.is?(:multipath) }
-
       multipath_wires = multipaths.each_with_object([]) { |m, r| r.concat(m.parents) }
 
-      result = multipaths + disks - multipath_wires
+      # And same for bios raids
+      bios_raids = component.select { |a| a.is?(:bios_raid) }
```
teclator: The code structure is almost the same an probably could be unified in one variable and with `component.select { |a| a.is?(:multipath) || a.is?(:bios_raid) }` but to be honest I prefer to maintain as it is

jreidinger: yes, me too. Even if there is some code and also abstracted similarities, in the end it is different technology that I would like to keep separately as this similarity is a bit accident and in other aspects it is different.

jreidinger: for me the main indicator why not to unify is that I have no clue how to name its variable :)

PR #513, opened 2018-05-17
more logging  and fix for detected multipath and bios raids

jreidinger:


src/modules/BootStorage.rb
```diff
@@ -197,8 +197,9 @@ def stage1_disks_for(device)
 
       # And same for bios raids
       bios_raids = component.select { |a| a.is?(:bios_raid) }
-      raid_members = bios_raids.each_with_object([]) { |m, r| r.concat(m.parents) }
-      log.info "bios_raids devices #{multipaths.inspect} and its members #{multipath_wires.inspect}"
+      # raid can be more complex, so we need even not even direct parents when we see bios raid
```
wfeldt: 'even not even'?

jreidinger: thanks, it should be probably "not only direct parents, but all ancestors that creates raid"

PR #516, opened 2018-05-29
fix pmbr propose for udev links

jreidinger:
in testing now

### DIFF
--- a/src/modules/BootStorage.rb
+++ b/src/modules/BootStorage.rb
@@ -71,10 +71,20 @@ module Yast
       # if bootloader do not know its location, then we do not care
       return false unless current_bl.respond_to?(:stage1)
 
+      !gpt_boot_disks.empty?
+    end
+
+    def gpt_boot_disks
+      require "bootloader/bootloader_factory"
+      current_bl = ::Bootloader::BootloaderFactory.current
       targets = current_bl.stage1.devices.map { |dev_name| staging.find_by_any_name(dev_name) }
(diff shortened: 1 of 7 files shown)


src/modules/BootStorage.rb
```diff
@@ -66,13 +66,13 @@ def gpt_boot_disk?
       require "bootloader/bootloader_factory"
       current_bl = ::Bootloader::BootloaderFactory.current
 
-      # efi require gpt disk, so it is always one
-      return true if current_bl.name == "grub2efi"
+      # efi bootloader does not have any boot disk and currently we support only GPT for it
+      return true if current_bl.name == "grub2-efi"
```
wfeldt: That's actually the point that's worrying me a bit. That check has been wrong for over two years without anyone noticing. What's the effect? Does it come down to whether we set the bit for EFI?

wfeldt: Ok, I've just checked and the bits seems not to be set in EFI mode. So I'm fine with it.

jreidinger: well, reality is that it has not effect. It is just relict of past, as now when proposing on grub2-efi it always remove pmbr - https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/grub2efi.rb#L64

and when write it detect once again what is disk where /boot/efi lives - https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/grub2efi.rb#L48

So basically not used anymore. It was useful in past. But I do not want to potentially change too much before GMC.

wfeldt: ok

PR #519, opened 2018-06-07
Full path parted

jreidinger:


ancorgs:
Don't we want this fix for Leap 15 as well? Only master?

jreidinger:
@ancorgs yep, only master makes sense as it is useful only in net installer run from "live" DVDs, so insts-sys only. On common system this is usually not problem.

PR #522, opened 2018-06-14
Warning upgrade

jreidinger:
screenshot with older text. Still probably space for improvement.

snimek obrazovky_2018-06-14_14-14-22

### DIFF
--- a/src/lib/bootloader/proposal_client.rb
+++ b/src/lib/bootloader/proposal_client.rb
@@ -7,6 +7,38 @@ require "yast2/popup"
 module Bootloader
   # Proposal client for bootloader configuration
   class ProposalClient < ::Installation::ProposalClient
+
+    class MismatchBootloader < RuntimeError
+      include Yast::I18n
+
+      def initialize(old_bootloader, new_bootloader)
+        @old_bootloader = old_bootloader
+        @new_bootloader = new_bootloader
+
+        super("Mismatching bootloaders")
+      end
+
+      def user_message
+        textdomain "bootloader"
+
+        boot_map = {
+          # TRANSLATORS: kind of boot. It is term for way how x86_64 can boot
+          "grub2" => _("Legacy BIOS boot"),
+          # TRANSLATORS: kind of boot. It is term for way how x86_64 can boot
+          "grub2-efi" => _("EFI boot")
+        }
+        # TRANSLATORS: keep %{} intact. It will be replaced by kind of boot
+        format(_(
+          "Cannot update bootloader as there is mismatch of boot technology.\n" \
+            "System to update is booted via %{old_boot} and media is boot via %{new_boot}.\n" \
+            "This scenario is not support. Updated system may stop booting when continue or \n" \
+            "update itself can fail."
+          ),
+          old_boot: boot_map[@old_bootloader], new_boot: boot_map[@new_bootloader]
+        )
+      end
+    end
+
     include Yast::I18n
     include Yast::Logger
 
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/proposal_client.rb
```diff
@@ -7,6 +7,38 @@
...
   class ProposalClient < ::Installation::ProposalClient
+    # Error when during update media is booted by different technology than target system.
+    class MismatchBootloader < RuntimeError
+      include Yast::I18n
+
+      def initialize(old_bootloader, new_bootloader)
+        @old_bootloader = old_bootloader
+        @new_bootloader = new_bootloader
+
+        super("Mismatching bootloaders")
```
lslezak: Do we display the exception somewhere? Is it OK to use an untranslated string?

test/bootloader_proposal_client_test.rb
```diff
@@ -222,5 +222,14 @@
       expect(Yast::PackagesProposal).to receive(:AddResolvables).with("yast2-bootloader", :package, packages_to_propose)
       subject.make_proposal({})
     end
+
+    it "returns warning if old system use different boot technology then new one" do
+      Yast.import "Mode"
```
lslezak: I'd move the import to the top of the file (and remove it from the other places), it will stay imported for the other tests anyway.

src/lib/bootloader/proposal_client.rb
```diff
@@ -7,6 +7,38 @@
...
+        }
+        # TRANSLATORS: keep %{} intact. It will be replaced by kind of boot
+        format(_(
+                 "Cannot update bootloader as there is mismatch of boot technology.<br>" \
+                   "System to update is booted via %{old_boot} and " \
+                   "media is booted via %{new_boot}.<br>" \
+                   "This scenario is not supported. Updated system may stop booting " \
+                   "when continue or update itself can fail."
+        ),
+          old_boot: boot_map[@old_bootloader], new_boot: boot_map[@new_bootloader])
```
lslezak: What if `@old_bootloader` or `@new_bootloader` value is missing in the `boot_map`? That will result in `nil` and empty strings. To make it robust I'd suggest to use the original values without mapping. The class name `MismatchBootloader` suggests it can handle any bootloaders, not only grub2.

Or alternatively rename it to something like `MismatchGrub2Bootloader` so it is clear it's designed only for grub2.

src/lib/bootloader/proposal_client.rb
```diff
@@ -7,6 +7,38 @@
...
+
+        boot_map = {
+          # TRANSLATORS: kind of boot. It is term for way how x86_64 can boot
+          "grub2"     => _("Legacy BIOS boot"),
+          # TRANSLATORS: kind of boot. It is term for way how x86_64 can boot
+          "grub2-efi" => _("EFI boot")
+        }
+        # TRANSLATORS: keep %{} intact. It will be replaced by kind of boot
+        format(_(
+                 "Cannot update bootloader as there is mismatch of boot technology.<br>" \
```
lslezak: Better wording + some nice rich text formatting to make it better readable:
```ruby
"<p>Cannot upgrade the bootloader because of a mismatch of the boot technology. " \
"The upgraded system uses <i>%{old_boot}</i> while the installation medium " \
"has been booted using <i>%{new_boot}</i>.</p>" \
"<p>This scenario is not supported, the upgraded system may not boot " \
"or the upgrade process can fail later.</p>"
```

src/lib/bootloader/proposal_client.rb
jreidinger: no, it is catched in only place where it is available. And if we display it, then it uses user_message.

src/lib/bootloader/proposal_client.rb
jreidinger: but then it will be less useful for users to see grub2 and grub2-efi instead of that legacy BIOS and EFI. I can ensure it in constructor, that it raises if something else is passed. Is it ok for you?

lslezak: Yes, that's OK.

PR #523, opened 2018-06-19
[no merge] More logging sle12

jreidinger:
just for moving forward L3


jreidinger:
no longer needed as L3 closed

PR #526, opened 2018-06-28
Added additional searchkeys to desktop file

schubi2:


src/desktop/bootloader.desktop
```diff
@@ -15,6 +15,7 @@ X-SuSE-YaST-SortKey=
 X-SuSE-YaST-AutoInstResource=
 X-SuSE-YaST-AutoInstPath=install
 X-SuSE-YaST-AutoInstSchema=bootloader.rnc
+X-SuSE-YaST-Keywords=bootloader,GRUB,system
```
jreidinger: maybe add also GRUB2 and EFI . Not sure if we should add also all supported archs like PPC, S390, ARM, AARCH, x86, i386 ?

PR #528, opened 2018-07-09
Do not crash with internal error when GRUB_TERMINAL contains multiple values

mchf:
Bnc#1053559

depends on https://github.com/config-files-api/config_files_api_grub2/pull/22
when merged, new CFA for GRUB2 gem based on ^^^ has to be released

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,3 +47,7 @@ that holds and also can propose the bootloader implementation. So now let's expl
 - [Boot Record backup](https://www.rubydoc.info/github/yast/yast-bootloader/master/Bootloader/BootRecordBackup) creates a backup of boot record for devices which the code touched
 - [Serial Console](https://www.rubydoc.info/github/yast/yast-bootloader/master/Bootloader/SerialConsole) converts between grub2 serial console configuration and kernel serial console configuration
 - [Sysconfig](https://www.rubydoc.info/github/yast/yast-bootloader/master/Bootloader/Sysconfig) read/write sysconfig configuration for bootloader. Sysconfig basically holds currently used grub2 implementation and configuration that is not in grub2 itself like if use secure boot or trusted grub
+
+### See Also
+
+The bootloader module uses custom CFA (config files API) implementation - [See CFA for Grub](https://github.com/config-files-api/config_files_api_grub2)
--- a/README.md
+++ b/README.md
@@ -50,4 +50,4 @@ that holds and also can propose the bootloader implementation. So now let's expl
 
 ### See Also
 
-The bootloader module uses custom CFA (config files API) implementation - [See CFA for Grub](https://github.com/config-files-api/config_files_api_grub2)
+The bootloader module uses CFA (config files API) module for GRUB2 - [See CFA for GRUB2 repository](https://github.com/config-files-api/config_files_api_grub2)
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -482,12 +482,12 @@ module Bootloader
     end
 
     def init
-      enable = grub_default.terminal == :serial
+      enable = grub_default.terminal.include?(:serial)
       Yast::UI.ChangeWidget(Id(:console_frame), :Value, enable)
       args = grub_default.serial_console || ""
       Yast::UI.ChangeWidget(Id(:console_args), :Value, args)
 
-      enable = grub_default.terminal == :gfxterm
+      enable = grub_default.terminal.include?(:gfxterm)
       Yast::UI.ChangeWidget(Id(:gfxterm_frame), :Value, enable)
 
       Yast::UI.ChangeWidget(Id(:gfxmode), :Items, vga_modes_items)
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -515,15 +515,15 @@ module Bootloader
     def store
       use_serial = Yast::UI.QueryWidget(Id(:console_frame), :Value)
       use_gfxterm = Yast::UI.QueryWidget(Id(:gfxterm_frame), :Value)
+      use_console = !use_serial && !use_gfxterm
 
-      use_gfxterm = false if use_gfxterm && use_serial
+      grub_default.terminal = []
+      grub_default.terminal = [:gfxterm] if use_gfxterm
 
       if use_serial
         console_value = Yast::UI.QueryWidget(Id(:console_args), :Value)
         BootloaderFactory.current.enable_serial_console(console_value)
-      elsif use_gfxterm
-        grub_default.terminal = [:gfxterm]
-      else
+      elsif use_console
         grub_default.terminal = [:console]
       end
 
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -47,3 +47,7 @@ that holds and also can propose the bootloader implementation. So now let's expl
 - [Boot Record backup](https://www.rubydoc.info/github/yast/yast-bootloader/master/Bootloader/BootRecordBackup) creates a backup of boot record for devices which the code touched
 - [Serial Console](https://www.rubydoc.info/github/yast/yast-bootloader/master/Bootloader/SerialConsole) converts between grub2 serial console configuration and kernel serial console configuration
 - [Sysconfig](https://www.rubydoc.info/github/yast/yast-bootloader/master/Bootloader/Sysconfig) read/write sysconfig configuration for bootloader. Sysconfig basically holds currently used grub2 implementation and configuration that is not in grub2 itself like if use secure boot or trusted grub
+
+### See Also
+
+The bootloader module uses custom CFA (config files API) implementation - [See CFA for Grub](https://github.com/config-files-api/config_files_api_grub2)
```
jreidinger: I would not write custom CFA, but instead "uses CFA module for GRUB2"

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -497,8 +497,6 @@ def init
       Yast::UI.ChangeWidget(Id(:gfxmode), :Value, mode) if mode && mode != ""
 
       Yast::UI.ChangeWidget(Id(:theme), :Value, grub_default.theme || "")
-    # FIXME: just temporary workaround for terminal that does contain
-    # more complex string (bsc#1053559)
```
jreidinger: you remove only comment, but keep catching of exception which expect GRUB_TERMINAL which is wrong.

src/lib/bootloader/grub2base.rb
```diff
@@ -212,17 +212,11 @@ def merge_grub_default(other)
...
-        end
-      # FIXME: only temporary solution to catch too complex grub terminal option (bsc#1053559)
-      # will be removed when cfa_grub2 and yast understand more complex terminal configuration
-      rescue RuntimeError
-        raise ::Bootloader::UnsupportedOption, "GRUB_TERMINAL"
+      # string attributes
+      [:serial_console, :terminal, :timeout, :hidden_timeout, :distributor,
+       :gfxmode, :theme, :default].each do |attr|
+        val = other.public_send(attr)
+        default.public_send((attr.to_s + "=").to_sym, val) if val
```
jreidinger: So terminal is still string? I do not think so, as above it is checked against symbol inclusion.

README.md
mchf: ok

src/lib/bootloader/grub2_widgets.rb
mchf: I've added support for parsing multiple values for that option ... so the exception means something different now.

jreidinger: OK, so maybe it is time to do it properly and define own exception class only for this case and catch this specialized class? When I did this workaround in past, I try to do minimal changes, but when you touched cfa already, it makes more sense for me.

src/lib/bootloader/grub2base.rb
```diff
@@ -212,17 +212,17 @@ def merge_grub_default(other)
...
+      [:serial_console, :timeout, :hidden_timeout, :distributor,
+       :gfxmode, :theme, :default].each do |attr|
+        val = other.public_send(attr)
+        default.public_send((attr.to_s + "=").to_sym, val) if val
+      end
+
+      # string attributes with multiple values allowed
+      [:terminal].each do |attr|
+        val = other.public_send(attr)
+        default.public_send((attr.to_s + "=").to_sym, val.split) if val
```
jreidinger: hmm, what is goal of split? It looks for me like bad place. Terminal should return array not? so it should be just val not?

src/lib/bootloader/grub2base.rb
```diff
@@ -212,17 +212,17 @@ def merge_grub_default(other)
...
-      rescue RuntimeError
-        raise ::Bootloader::UnsupportedOption, "GRUB_TERMINAL"
+      # string attributes
+      [:serial_console, :timeout, :hidden_timeout, :distributor,
+       :gfxmode, :theme, :default].each do |attr|
+        val = other.public_send(attr)
+        default.public_send((attr.to_s + "=").to_sym, val) if val
+      end
+
+      # string attributes with multiple values allowed
```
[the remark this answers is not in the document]
mchf: hm ... i wanted to express something else here ... but as it seems I was unclear ;-)

test/grub2base_test.rb
```diff
@@ -455,6 +455,16 @@
       expect(subject.password).to be_password
     end
 
+    it "use terminal configuration specified in the merged object" do
+      TERMINAL_DEFINITION = "console serial"
```
jreidinger: this looks wrong to me. I expect it should return array and not string, not?

jreidinger: I expect something like `[:console, :serial]`

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -535,9 +522,9 @@ def store
         console_value = Yast::UI.QueryWidget(Id(:console_args), :Value)
         BootloaderFactory.current.enable_serial_console(console_value)
       elsif use_gfxterm
-        grub_default.terminal = :gfxterm
+        grub_default.terminal = [:gfxterm]
```
jreidinger: this is in fact wrong as currently the code does not allow to set both to true ( https://github.com/yast/yast-bootloader/pull/528/files#diff-8a7a995c9c06bae0fbc6acd4ccd71a23R519 ). This can be now removed and it should here add/remove `:gfxterm` and `:console` according to user configuration.

jreidinger: ah, I mention gfxterm and console, but mean gfxterm and serial console which can be used in parallel. For gfxterm and console it does not make much sense.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -535,9 +522,9 @@ def store
...
       else
-        grub_default.terminal = :console
+        grub_default.terminal = [:console]
```
jreidinger: so maybe current logic with gfxterm versus console is correct. But user should be able to set in parallel serial console and console/gfxterm.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -528,16 +515,16 @@ def validate
...
 
       if use_serial
         console_value = Yast::UI.QueryWidget(Id(:console_args), :Value)
         BootloaderFactory.current.enable_serial_console(console_value)
-      elsif use_gfxterm
-        grub_default.terminal = :gfxterm
-      else
-        grub_default.terminal = :console
+      elsif use_console
+        grub_default.terminal = [:console]
```
jreidinger: code wise it is good now, only missing piece for me is to document how it works. Ideally in widget documentation mention what we will set in which case.
Thanks

PR #529, opened 2018-07-20
switching to installed system for reading grub settings

schubi2:
While calling Bootloader.Read() SCR still points to inst-sys. So we should switch to /mnt before reading the grub settings.

### DIFF
--- a/src/modules/Bootloader.rb
+++ b/src/modules/Bootloader.rb
@@ -422,16 +422,17 @@ module Yast
         Propose()
       else
         progress_orig = Progress.set(false)
-        if Mode.update
+        if Stage.initial && Mode.update
           # SCR has been currently set to inst-sys. So we have
           # set the SCR to installed system in order to read
           # grub settings
           old_SCR = WFM.SCRGetDefault
-          new_SCR = WFM.SCROpen("chroot=/mnt:scr", false)
+          new_SCR = WFM.SCROpen("chroot=#{Yast::Installation.destdir}:scr",
+            false)
           WFM.SCRSetDefault(new_SCR)
         end
         Read()
-        if Mode.update
+        if Stage.initial && Mode.update
           # settings have been read from the target system
           current_bl.read
           # reset target system to inst-sys
--- a/src/modules/Bootloader.rb
+++ b/src/modules/Bootloader.rb
@@ -422,7 +422,22 @@ module Yast
         Propose()
       else
         progress_orig = Progress.set(false)
+        if Mode.update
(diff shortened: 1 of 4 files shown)


src/modules/Bootloader.rb
```diff
@@ -422,7 +422,22 @@ def ReadOrProposeIfNeeded
         Propose()
       else
         progress_orig = Progress.set(false)
+        if Mode.update
+          # SCR has been currently set to inst-sys. So we have
+          # set the SCR to installed system in order to read
+          # grub settings
+          old_SCR = WFM.SCRGetDefault
+          new_SCR = WFM.SCROpen("chroot=/mnt:scr", false)
```
jreidinger: maybe use Installation.destdir instead of hardcoded "/mnt"

src/modules/Bootloader.rb
```diff
@@ -422,7 +422,22 @@ def ReadOrProposeIfNeeded
         Propose()
       else
         progress_orig = Progress.set(false)
+        if Mode.update
```
jreidinger: please also check Stage.initial, as it can be broken in autoupgrade in second stage.

src/modules/Bootloader.rb
```diff
@@ -422,16 +422,17 @@ def ReadOrProposeIfNeeded
...
         progress_orig = Progress.set(false)
-        if Mode.update
+        if Stage.initial && Mode.update
           # SCR has been currently set to inst-sys. So we have
           # set the SCR to installed system in order to read
           # grub settings
           old_SCR = WFM.SCRGetDefault
-          new_SCR = WFM.SCROpen("chroot=/mnt:scr", false)
+          new_SCR = WFM.SCROpen("chroot=#{Yast::Installation.destdir}:scr",
+            false)
```
dgdavid: NP: it seems not necessary to add a new line here :)

src/modules/Bootloader.rb
```diff
@@ -422,7 +422,23 @@ def ReadOrProposeIfNeeded
         Propose()
       else
         progress_orig = Progress.set(false)
+        if Stage.initial && Mode.update
+          # SCR has been currently set to inst-sys. So we have
+          # set the SCR to installed system in order to read
+          # grub settings
+          old_SCR = WFM.SCRGetDefault
+          new_SCR = WFM.SCROpen("chroot=#{Yast::Installation.destdir}:scr",
```
jreidinger: I know that it probably gets loaded by some other module, but it is not imported. So please add import of "Installation".

PR #530, opened 2018-07-23
Master 1094031

schubi2:


jreidinger:
@Fraser-Bell it is result of old ycp to ruby conversion. So yes, it should be rb now, but it does not matter much and I usually remove it from new code, as anyone can read what is file name.

PR #536, opened 2018-08-29
Avoid to crash when a required package is not installed

dgdavid:
Fix for https://bugzilla.suse.com/show_bug.cgi?id=1089829

> avoiding to continue writing the configuration and reverting
the sysconfig when user rejects the installation of required package

### DIFF
--- a/src/lib/bootloader/bootloader_base.rb
+++ b/src/lib/bootloader/bootloader_base.rb
@@ -99,7 +99,8 @@ module Bootloader
 
     # Writes the sysconfig readed in the initialization
     #
-    # Useful to "rollback" sysconfig changes if something is fails before finish the configuration
+    # Useful to "rollback" sysconfig changes if something fails before finish writing the
+    # configuration
     def restore_initial_sysconfig
       @initial_sysconfig.write
     end
--- a/src/lib/bootloader/bootloader_base.rb
+++ b/src/lib/bootloader/bootloader_base.rb
@@ -17,8 +17,8 @@ module Bootloader
 
     # Prepares the system to (before write the configuration)
     #
-    # Writes the new sysconfig and, when the Mode.normal is set, tries to install the required packages.
-    # If user decides to cancel the installation, it restores the previous sysconfig.
+    # Writes the new sysconfig and, when the Mode.normal is set, tries to install the required
+    # packages. If user decides to cancel the installation, it restores the previous sysconfig.
     #
     # @return [Boolean] true whether the system could be prepared as expected;
     #                   false when user cancel the installation of needed packages
@@ -66,11 +66,8 @@ module Bootloader
     def packages
       res = []
 
-      # added kexec-tools fate# 303395
-      if !Yast::Mode.live_installation &&
-          Yast::Linuxrc.InstallInf("kexec_reboot") != "0"
-        res << "kexec-tools"
-      end
+      # added kexec-tools fate#303395
+      res << "kexec-tools" if include_kexec_tools_package?
 
       res
     end
@@ -82,13 +79,6 @@ module Bootloader
       prewrite ? sysconfig.pre_write : sysconfig.write
     end
 
-    # Writes the sysconfig readed in the initialization
-    #
-    # Useful to "rollback" sysconfig changes if something is fails before finish the configuration
-    def restore_initial_sysconfig
-      @initial_sysconfig.write
-    end
-
     # merges other bootloader configuration into this one.
     # It have to be same bootloader type.
     def merge(other)
@@ -96,6 +86,22 @@ module Bootloader
 
       @read ||= other.read?
       @proposed ||= other.proposed?
+    end
+
+  private
+
+    # @return [Boolean] true when kexec-tools package should be included; false otherwise
+    def include_kexec_tools_package?
+      return false if Yast::Mode.live_installation
+
+      Yast::Linuxrc.InstallInf("kexec_reboot") != "0"
+    end
+
+    # Writes the sysconfig readed in the initialization
+    #
+    # Useful to "rollback" sysconfig changes if something is fails before finish the configuration
+    def restore_initial_sysconfig
+      @initial_sysconfig.write
     end
   end
 end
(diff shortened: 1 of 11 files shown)


src/lib/bootloader/bootloader_base.rb
```diff
@@ -9,16 +10,38 @@
...
+      Yast2::Popup.show(_("Cannot continue without install required packages"))
+      false
     end
 
     # writes configuration to target disk
     def write
-      write_sysconfig
-      # in running system install package, for other modes, it need specific handling
-      Yast::PackageSystem.InstallAll(packages) if Yast::Mode.normal
+      raise NotImplementedError
```
jreidinger: this cannot work. All children call super, so it raise exception. If there is nothing to do, just keep it empty.

src/lib/bootloader/grub2base.rb
```diff
@@ -93,8 +93,6 @@ def read
     end
 
     def write
-      super
```
jreidinger: ah, you remove it. I think it is bad idea. If we have something to do always, then we can add it to base. I prefer keeping it empty and calling super here.

dgdavid: Ok.

src/lib/bootloader/bootloader_base.rb
jreidinger: EmptyBootloader will still fail.

dgdavid: I remove from children, but I will to leave it empty.

src/lib/bootloader/grub2.rb
```diff
@@ -157,6 +149,20 @@ def write_sysconfig(prewrite: false)
 
   private
 
+    # @return [Boolean] true if syslinux package should be included; false otherwise
+    def include_syslinux_package?
+      return false if Yast::Stage.initial
```
jreidinger: please move here also explanation from above why we skip it in insts-sys.

src/lib/bootloader/main_dialog.rb
```diff
@@ -52,7 +52,7 @@ def run
         "ws_start" => "read",
         "read"     => { abort: :abort, next: "main" },
         "main"     => { abort: :abort, next: "write" },
-        "write"    => { abort: :abort, next: :next }
+        "write"    => { abort: "main", next: :next }
```
jreidinger: so if you click on abort during write it will return you to main? Is it required?

src/lib/bootloader/write_dialog.rb
```diff
@@ -9,11 +9,13 @@ class WriteDialog
...
-    # @return `:abort` if aborted and `:next` otherwise
+    #
+    # @return [Symbol] :abort if aborted
+    #                  :next otherwise
     def run
       Yast::Wizard.RestoreHelp(help_text)
-      ret = Yast::Bootloader.Write
-      ret ? :next : :abort
+      return :next if Yast::Bootloader.Write
+      :abort
```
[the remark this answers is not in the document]
dgdavid: In fact, I though first in your proposed improvement, `Yast::Bootloader.Write ? :next : :abort`, but usually I try to avoid the ternary operator. Moreover, the use of "guard clauses" if looks more readable and natural to me.

That said, if you feel uncomfortable with this style I can to use your suggestion instead.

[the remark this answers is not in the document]
jreidinger: well, for this case I do not like much guard clauses as it hides that real action is done behind if at the end of line, which is not nature for me.
If you want to use guard clause then I prefer something like
```
failed = !Yast::Bootloader.Write
return :abort if failed

:next
```
including new line so it is clear separation.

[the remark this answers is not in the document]
dgdavid: I (partially) agree with 

>  hides that real action is done behind if at the end of line

So, I am going to use the ternary operator for this case, in order to keep it simple and not add extra lines only for use an specific technique.

Thank you.

src/lib/bootloader/main_dialog.rb
jreidinger: then I suggest to use different symbol like `:cancel` and not :abort as abort is used also when user click on abort.

dgdavid: Thank you @jreidinger 

The thing is how to distinguish when a false from Bootloader.Write is "translated" to `:cancel` and when to `:abort`.

I will think about it. Suggestions are welcome :)

src/lib/bootloader/bootloader_base.rb
```diff
@@ -74,5 +80,29 @@ def merge(other)
...
+    # @return [Boolean] true when kexec-tools package should be included; false otherwise
+    def include_kexec_tools_package?
+      return false if Yast::Mode.live_installation
+
+      Yast::Linuxrc.InstallInf("kexec_reboot") != "0"
+    end
+
+    # done in common write but also in installation pre write as kernel update need it
+    # @param prewrite [Boolean] true only in installation when scr is not yet switched
+    def write_sysconfig(prewrite: false)
```
jreidinger: why it is moved to private? This cause exception during installation at https://github.com/yast/yast-bootloader/blob/master/src/clients/inst_bootloader.rb#L45

src/lib/bootloader/grub2.rb
```diff
@@ -155,8 +166,6 @@ def write_sysconfig(prewrite: false)
       prewrite ? sysconfig.pre_write : sysconfig.write
     end
 
-  private
```
jreidinger: and same here. Write sysconfig have to be public method as it is used in installation.

src/lib/bootloader/main_dialog.rb
jreidinger: well, not so easy. so lets keep it as it is. Maybe just keep here abort and see if someone complains.

src/lib/bootloader/bootloader_base.rb
dgdavid: Sorry, I did not realize about it.

src/lib/bootloader/main_dialog.rb
dgdavid: @jreidinger  just to clarify, you mean to keep it as it was

```ruby
"write"    => { abort: :abort, next: :next }
```

right?

jreidinger: yes

PR #537, opened 2018-09-14
patch to not show popup when editenv list failed

jreidinger:
not for merge, just to unblock OES


jreidinger:
closing as it was one shot for OES and not change that should go to upstream.

PR #540, opened 2018-10-23
Improve broken configuration detection

jreidinger:
https://bugzilla.suse.com/show_bug.cgi?id=1111236


test/boot_storage_test.rb
```diff
@@ -155,6 +155,17 @@
 
         expect(subject.gpt_boot_disk?).to eq false
       end
+
+      it "raises BrokenConfiguration if in stage1 is unknown device" do
```
mvidner:
```suggestion
      it "raises BrokenConfiguration if there is an unknown device in stage1" do
```

PR #544, opened 2018-11-22
Secure boot arm

jreidinger:


ggardet:
> cannot merge now as shim and mokutil is x86_64 only.

And there will be no shim for aarch64. Here is my PR (untested yet): https://github.com/yast/yast-bootloader/pull/565

jreidinger:
@ggardet how it can work without shim? I probably need more details to know about how it should work ( ideally some persistent URL, so I can link it from code for future references.

ggardet:
> @ggardet how it can work without shim? I probably need more details to know about how it should work ( ideally some persistent URL, so I can link it from code for future references.

The only problem I have is that signed grub is loaded properly, but is not configured. 
I still need to figure out how to fix it.

jreidinger:
replaced by https://github.com/yast/yast-bootloader/pull/565

PR #547, opened 2018-12-20
Backport booting fx

jreidinger:
this is backport of SLE12 SP4 fix from https://github.com/yast/yast-bootloader/pull/524 to SLE 12 SP3 as OES needs it.

### DIFF
--- a/src/lib/bootloader/proposal_client.rb
+++ b/src/lib/bootloader/proposal_client.rb
@@ -84,7 +84,7 @@ module Bootloader
         if grub2_update?(current_bl)
           ::Yast2::Popup.show(
             # TRANSLATORS: popup text when user click on link and we forbid to continue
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/proposal_client.rb
```diff
@@ -76,6 +77,18 @@ def make_proposal(attrs)
     end
 
     def ask_user(param)
+      if Yast::Mode.update
+        current_bl = ::Bootloader::BootloaderFactory.current
```
dgdavid: NP: I know that this is a backport, but I must to say that I prefere `current_bootlader` as param name. More meaningful, IMO :angel:

jreidinger: it is just two line scope. So for me it was fine. I picking better names if it has longer scope. So single or two letter for single line is fine and then increasing it.

PR #550, opened 2019-02-06
Better Error Handling for Broken by-path Device Names

shundhammer:
## Trello

https://trello.com/c/3HUj2DWo

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1122008
https://bugzilla.suse.com/show_bug.cgi?id=1116305

## Problem

If a user uses a `/dev/disk/by-path/...` device name (YaST never does that by itself!) for the bootloader configuration, that device name may have become invalid upon a future reboot. That doesn't take much: If the hardware changes only slightly, if the kernel parameters change, the kernel may decide to change the numbering scheme for devices, and so this kind of device name may change, making the old device name invalid.

This caused a hard to understand error message:

```
Internal error. Please report a bug report with logs.
Run save_y2logs to get complete logs.

Details: Error reading the bootloader configuration files.
Please use YaST2 bootloader to fix it.
Details: Unknown device /dev/disk/by-path/pci-0000:00:08.0
```

This gave the impression that this is a crash caused by a programming error in bug which it is clearly not.

## Fix

Changed the error message to something that should be clearer to understand:

```
Error reading the bootloader configuration files:
Invalid device name /dev/disk/by-path/pci-0000:00:08.0

This by-path device name may have changed after a reboot
if the hardware or kernel parameters changed.

Please use YaST2 bootloader to fix this.
```

It is still a fatal error, and there is no way to automatically recover from this. It still requires user interaction. But at least it should be much clearer now what the problem is, and it no longer looks (so much) like a problem caused by a bug in the YaST code.

We agreed to limit it to making the error message clearer. This was approved by @jsrain.

### DIFF
--- a/src/lib/bootloader/exceptions.rb
+++ b/src/lib/bootloader/exceptions.rb
@@ -26,6 +26,27 @@ module Bootloader
     end
   end
 
+  # Specialized exception for invalid by-path device names
+  # (bsc#1122008, bsc#1116305)
+  class BrokenByPathDeviceName < RuntimeError
+    include Yast::I18n
+    attr_reader :dev_name
+
+    def initialize(dev_name)
+      @dev_name = dev_name
+      textdomain "bootloader"
+
+      # TRANSLATORS: %s is the device name
+      super _("Error reading the bootloader configuration files:\n" \
+        "Invalid device name %s\n" \
+        "\n" \
+        "This by-path device name may have changed after a reboot\n" \
+        "if the hardware or kernel parameters changed.\n" \
+        "\n" \
+        "Please use YaST2 bootloader to fix this.\n") % dev_name
+    end
+  end
+
   # Represent unsupported value in given option. Used mainly when value contain something that
   # bootloader does not understand yet.
   class UnsupportedOption < RuntimeError
--- a/src/modules/BootStorage.rb
+++ b/src/modules/BootStorage.rb
@@ -72,8 +72,7 @@ module Yast
     # @return [Array<String>] gpt disks only
     def gpt_disks(devices)
       targets = devices.map do |dev_name|
-        staging.find_by_any_name(dev_name) or
-          raise ::Bootloader::BrokenConfiguration, "Unknown device #{dev_name}"
+        staging.find_by_any_name(dev_name) or handle_unknown_device(dev_name)
       end
       boot_disks = targets.each_with_object([]) { |t, r| r.concat(stage1_disks_for(t)) }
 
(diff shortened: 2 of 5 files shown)


jsrain:
Text-wise it is OK with me.

src/lib/bootloader/exceptions.rb
```diff
@@ -26,6 +26,27 @@ def initialize(msg)
...
+  class BrokenByPathDeviceName < RuntimeError
+    include Yast::I18n
+    attr_reader :dev_name
+
+    def initialize(dev_name)
+      @dev_name = dev_name
+      textdomain "bootloader"
+
+      # TRANSLATORS: %s is the device name
+      super _("Error reading the bootloader configuration files:\n" \
```
joseivanlopez: NP: I am fine with `%` usage for formatting the string, but it would be preferable to use `format` method, that is:

`format(_("... %{dev} ..."), dev: dev_name)`

src/modules/BootStorage.rb
```diff
@@ -275,6 +274,25 @@ def select_ancestors(device, &predicate)
...
+    # Handle an "unknown device" error: Raise an appropriate exception.
+    # @param dev_name [String]
+    def handle_unknown_device(dev_name)
+      # rubocop:disable Style/GuardClause
+      #
+      # I flatly refuse to make my code LESS readable because of a third-rate
+      # check tool. This is the CLASSIC use case for if...else, even if this
+      # mindless rubocop thinks otherwise.
+      #
+      # 2019-02-06 shundhammer
```
joseivanlopez: xD

src/modules/BootStorage.rb
```diff
@@ -275,6 +274,25 @@ def select_ancestors(device, &predicate)
...
+
+      if dev_name =~ %r{/by-path/} # bsc#1122008, bsc#1116305
+        raise ::Bootloader::BrokenByPathDeviceName, dev_name
+      else
+        raise ::Bootloader::BrokenConfiguration, "Unknown device #{dev_name}"
+      end
+      # rubocop:enable Style/GuardClause
+    end
```
joseivanlopez: I guess rubocop expects something like:

```
raise ::Bootloader::BrokenByPathDeviceName, dev_name if dev_name =~ %r{/by-path/}

raise ::Bootloader::BrokenConfiguration, "Unknown device #{dev_name}"
```

Or

```
if dev_name =~ %r{/by-path/} # bsc#1122008, bsc#1116305
  raise ::Bootloader::BrokenByPathDeviceName, dev_name
end

raise ::Bootloader::BrokenConfiguration, "Unknown device #{dev_name}"  
```

Maybe this last version continues being quite readable and rubocops will be fine.

src/lib/bootloader/exceptions.rb
ancorgs: In fact, I believe that `%` and `format` are basically aliases. So you can use both specifications (`%{var_name}` or `%s`) with both methods. And in general I also prefer the `%{variable_name}` way.

BUT translators are more likely to mess up things when we use the Ruby-specific way than when we use the more standard `%s` one. So I would stick if possible to the more widespread format.

PR #553, opened 2019-02-21
Relax lvm checks

jreidinger:
trello: https://trello.com/c/7l3D6O7p/734-sle15-sp1-p3-1125792-bootloader-installation-shows-error-messages-but-appears-to-work-nonetheless

bzilla: https://bugzilla.suse.com/show_bug.cgi?id=1125792

### DIFF
--- a/src/lib/bootloader/grub_install.rb
+++ b/src/lib/bootloader/grub_install.rb
@@ -2,14 +2,26 @@ require "yast"
 require "yast2/execute"
 
 Yast.import "Arch"
+Yast.import "Report"
 
 module Bootloader
   # Wraps grub install script for easier usage.
   class GrubInstall
+    include Yast::Logger
+    include Yast::I18n
+
     def initialize(efi: false)
       @efi = efi
+      textdomain "bootloader"
     end
 
+    # Runs grub2 install command.
+    #
+    # @param devices[Array<String>] list to which grub2 should be installed.
+    #   Ignored when grub2 does not need device.
+    # @param secure_boot [Boolean] if secure boot variant should be used
+    # @param trusted_boot [Boolean] if trusted boot variant should be used
+    # @return [Array<String>] list for which install failed
     def execute(devices: [], secure_boot: false, trusted_boot: false)
       raise "cannot have secure boot without efi" if secure_boot && !efi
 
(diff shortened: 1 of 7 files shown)


src/lib/bootloader/grub_install.rb
```diff
@@ -2,30 +2,72 @@
...
+            false
+          rescue Cheetah::ExecutionFailed => e
+            log.warn "Failed to install grub to device #{device}. #{e.inspect}"
+            last_failure = e
+            true
+          end
+        end
+
+        # Failed to install to all devices
+        report_failure(last_failure) if res.size == devices.size
```
mchf: so it can fail for all devices but reports only one? It is a bit strange in light of fact that error is reported *only* when grub could not be written to all devices.

src/lib/bootloader/grub_install.rb
```diff
@@ -2,30 +2,72 @@
...
 
       cmd = basic_cmd(secure_boot, trusted_boot)
 
       if no_device_install?
         Yast::Execute.on_target(cmd)
+        []
       else
-        devices.each { |d| Yast::Execute.on_target(cmd + [d]) }
+        return [] if devices.empty?
+        last_failure = nil
```
mchf: seems useless

src/lib/bootloader/grub_install.rb
```diff
@@ -2,30 +2,72 @@
...
+    include Yast::I18n
+
     def initialize(efi: false)
       @efi = efi
+      textdomain "bootloader"
     end
 
+    # Runs grub2 install command.
+    #
+    # @param devices[Array<String>] list to which grub2 should be installed.
```
mchf: ... list of devices where ....

src/lib/bootloader/grub_install.rb
jreidinger: in logs you see all failures, but reported is only last one as usually it is just single reason for all devices ( like missing bios grub, missing grub2 at all or some other issues ). So goal is to report user that there is a error ( so we do not hide that we have problem to make system bootable ), but it is not shown for all failures.

src/lib/bootloader/grub_install.rb
jreidinger: in fact no, I just try it and inside block is only local scope. Try it yourself:

```
irb(main):003:0> [1, 2].select { |a| abc = 5 }
=> [1, 2]
irb(main):004:0> abc
Traceback (most recent call last):
        2: from /usr/bin/irb:11:in `<main>'
        1: from (irb):4
NameError (undefined local variable or method `abc' for main:Object)
```

so it needs to be defined on higher scope.

src/lib/bootloader/grub_install.rb
mchf: well, I meant that just a report "Failed to install grub." would be enough instead of appending a device name ... but that's only NP for me.

src/lib/bootloader/grub_install.rb
```diff
@@ -2,30 +2,73 @@
...
         Yast::Execute.on_target(cmd)
+        []
       else
-        devices.each { |d| Yast::Execute.on_target(cmd + [d]) }
+        return [] if devices.empty?
+
+        last_failure = nil
+        res = devices.select do |device|
+          begin
+            Yast::Execute.on_target!(cmd + [device])
```
mchf: just for record: what about shell escaping ```device``` ?

jreidinger: Cheetah is not affected by shell execution as it does not use shell and pass arguments directly. See https://github.com/openSUSE/cheetah#features

mchf: ok, thanks. I will have to study cheetah more deeply.

jreidinger:
Manual testing confirm that fix works as expected.

PR #558, opened 2019-03-25
Master bsc 1128707 2

schubi2:


jreidinger:
hmmm, I do not get it. You add it with https://github.com/yast/yast-bootloader/pull/557 and now removing.

schubi2:
Yes, #557 has been wrong because it has already been defined later.

PR #561, opened 2019-04-18
Mitigations

jreidinger:


Dockerfile
```diff
@@ -1,2 +1,2 @@
-FROM registry.opensuse.org/yast/head/containers/yast-ruby:latest
+FROM yastdevel/ruby:sle15-sp1
```
jreidinger: @lslezak SP1 uses latest?

lslezak: For SP1 you should use `yastdevel/ruby:sle15-sp1`, that's correct.

lslezak: Ah, this is for master, for master use `registry.opensuse.org/yast/head/containers/yast-ruby:latest`.

PR #562, opened 2019-04-18
Cpu mitigation

jreidinger:


src/lib/bootloader/grub2_widgets.rb
```diff
@@ -113,30 +113,46 @@ def store
...
     end
 
     def label
-      _("Disable Simultaneous &Multithreading")
+      _("CPU Mitigations")
+    end
+
+    def items
+      [
+        [:nosmt, _("Auto + No SMT")],
```
lslezak: I'd use more verbose labels like "Automatic with SMT disabled"...

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -113,30 +113,46 @@ def store
...
+        [:auto, _("Keep SMT")],
```
lslezak: ... and "Keep SMT Enabled" or "Automatic with SMT enabled" (to be similar to the previous one).

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -113,30 +113,46 @@ def store
...
+        [:off, _("Off")],
+        [:manual, _("Manually")]
```
lslezak: Manual would be more consistent:

```suggestion
        [:manual, _("Manual")]
```

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -113,30 +113,46 @@ def store
...
+      ]
     end
 
     def help
+      # TODO: adapt
       _(
-        "<p><b>Disable Simultaneous Multithreading</b><br>\n" \
-          "To disable sharing physical cores by more virtual ones."
+        "<p><b>CPU Speculation</b><br>\n" \
```
lslezak: The title here is "CPU Speculation" but the widget uses the "CPU Mitigations" label. Which one is correct?

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -113,30 +113,46 @@ def store
...
+          "Controls CPU speculative execution. <i>Secure</i> is the most secure option " \
```
lslezak: The "Secure" option has been renamed. Looks like the help has not been updated.

src/lib/bootloader/grub2base.rb
```diff
@@ -28,6 +28,7 @@
 
 module Bootloader
   # Common base for GRUB2 specialized classes
+  # rubocop:disable Metrics/ClassLength
```
lslezak: I'd prefer adding it to `.rubocop.yml` instead of polluting the code...

src/lib/bootloader/grub2base.rb
```diff
@@ -338,4 +359,5 @@ def propose_encrypted
       grub_default.cryptodisk.value = !!Yast::BootStorage.encrypted_boot?
     end
   end
+  # rubocop:enable Metrics/ClassLength
```
lslezak: Please (re)move to `.rubocop.yml`.

src/lib/bootloader/grub2base.rb
jreidinger: it is already in master, so I just use it.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -113,30 +114,54 @@ def store
...
 
     def help
+      # TODO: adapt
       _(
-        "<p><b>Disable Simultaneous Multithreading</b><br>\n" \
-          "To disable sharing physical cores by more virtual ones."
+        "<p><b>CPU Mitigations</b><br>\n" \
+          "The option selects which default settings should be used for CPU " \
+          "side channels mitigations. A highlevel description is on our Technical Information " \
+          "Document TID 7023836. Following options are available:<ul>" \
```
mvidner: Here it is. A link to this document should be in the class YARD and commit messages:
https://www.suse.com/support/kb/doc/?id=7023836
("The Article Cannot Be Found" but I assume it will be published later)

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -113,30 +114,54 @@ def store
...
+    end
+
+    def items
+      ::Bootloader::CpuMitigations::ALL.map do |m|
+        [m.value, m.to_human_string]
+      end
     end
 
     def help
+      # TODO: adapt
```
mvidner: Still TODO??

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -113,30 +114,54 @@ def store
...
       _(
-        "<p><b>Disable Simultaneous Multithreading</b><br>\n" \
-          "To disable sharing physical cores by more virtual ones."
+        "<p><b>CPU Mitigations</b><br>\n" \
+          "The option selects which default settings should be used for CPU " \
```
mvidner: `\n` are missing at the end of **every** line. 
```suggestion
          "The option selects which default settings should be used for CPU\n" \
```

src/lib/bootloader/grub2base.rb
```diff
@@ -54,7 +54,7 @@ def initialize
       @grub_default = ::CFA::Grub2::Default.new
       @sections = ::Bootloader::Sections.new
       @pmbr_action = :nothing
-      @smt = nil # nil means not set explicitly, otherwise boolean
+      @explicit_cpu_speculation = false
```
mvidner: Gotcha!
```suggestion
      @explicit_cpu_mitigations = false
```

src/lib/bootloader/grub2_widgets.rb
jreidinger: it is rich text so it does not matter.

src/lib/bootloader/grub2_widgets.rb
jreidinger: not anymore, will remove it.

src/lib/bootloader/grub2base.rb
jreidinger: good catch, it was old name.

src/lib/bootloader/grub2_widgets.rb
mvidner: It does matter to the translators who will have to read it with their eyes. Likewise we do format HTML code even if the browser renders it anyway

src/autoyast-rnc/bootloader.rnc
```diff
@@ -86,7 +86,7 @@ boot_root = element boot_root { "true" | "false" }
 boot_boot = element boot_boot { "true" | "false" }
 boot_extended = element boot_extended { "true" | "false" }
 boot_mbr = element boot_mbr { "true" | "false" }
-smt = element smt { "true" | "false" }
+cpu_mitigations = element cpu_mitigations { "nosmt", "auto", "off", "manual" }
```
lslezak: Looks like the same problem as in https://github.com/yast/yast-installation-control/pull/81/files#diff-1e14931051b4c04ff21df76fc3588959R232

src/lib/bootloader/cpu_mitigations.rb
```diff
@@ -0,0 +1,75 @@
...
+    ALL = KERNEL_MAPPING.keys.map { |k| CpuMitigations.new(k) }
+    DEFAULT = CpuMitigations.new(:auto)
+
+    def self.from_kernel_params(kernel_params)
+      log.info "kernel params #{kernel_params.inspect}"
+      param = kernel_params.parameter("mitigations")
+      log.info "mitigation param #{param.inspect}"
+      param = nil if param == false
+      reverse_mapping = KERNEL_MAPPING.invert
+      raise "Unknown mitigations value #{param.inspect}" if !reverse_mapping.key?(param)
```
lslezak: Um, where is this actually used? Is the exception handled somewhere?

I mean what happens when I boot with `mitigations=invalid_valued` boot parameter? Does it crash?

src/autoyast-rnc/bootloader.rnc
jreidinger: yep, you are right.

src/lib/bootloader/cpu_mitigations.rb
jreidinger: yep, it is more like assertion that there is no unknown parameter. I am not fan of making it ConfigurationError, as we can be wrong and proposing it from scratch does not help as it failing in proposal.

lslezak: OK, in that case I'd prefer a more verbose message so the user knows what wrong and how to fix it. Something like:

`Unknown mitigations value #{param.inspect} in the kernel command line, supported values are: #{KERNEL_MAPPING.values.reject{|v| v.nil?}.map(&:inspect).join(", ")}.`

mvidner: `reject{|v| v.nil?}` is IIRC called `compact`

PR #564, opened 2019-05-24
Fix resume with uuid partition

jreidinger:
Explanation: UUID is known after partition is created, but kernel
parameters including resume is proposed in advance. So we need to try to
translate device to uuid later.

https://bugzilla.suse.com/show_bug.cgi?id=1134895


jreidinger:
closed for now as it was not merged in SP5 window

PR #565, opened 2019-05-28
[RFC] Enable secureboot on aarch64 (boo#1136601)

ggardet:
https://bugzilla.suse.com/show_bug.cgi?id=1136601

Not tested yet

### DIFF
--- a/src/lib/bootloader/grub_install.rb
+++ b/src/lib/bootloader/grub_install.rb
@@ -72,10 +72,12 @@ module Bootloader
     # creates basic command for grub2 install without specifying any stage1
     # locations
     def basic_cmd(secure_boot, trusted_boot)
-      if secure_boot
+      if secure_boot && !Yast::Arch.aarch64
         cmd = ["/usr/sbin/shim-install", "--config-file=/boot/grub2/grub.cfg"]
       else
         cmd = ["/usr/sbin/grub2-install", "--target=#{target}"]
+        # On aarch64, we do not use shim, but '--suse-signed-grub' option (bsc#1136601)
+        cmd << "--suse-signed-grub" if secure_boot && Yast::Arch.aarch64
         # Do skip-fs-probe to avoid error when embedding stage1
         # to extended partition
         cmd << "--force" << "--skip-fs-probe"
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Jun  5 12:59:27 UTC 2019 - Guillaume GARDET <guillaume.gardet@opensuse.org>
+
+- Enable Secure Boot on AArch64 (boo#1136601)
+- 4.2.4
+
+-------------------------------------------------------------------
 Fri May 31 12:26:38 UTC 2019 - Stasiek Michalski <hellcp@mailbox.org>
 
 - Add metainfo (fate#319035)
--- a/package/yast2-bootloader.spec
+++ b/package/yast2-bootloader.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-bootloader
-Version:        4.2.3
+Version:        4.2.4
 Release:        0
 Summary:        YaST2 - Bootloader Configuration
 License:        GPL-2.0-or-later
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -977,7 +977,7 @@ module Bootloader
     end
 
     def secure_boot_widget?
-      (Yast::Arch.x86_64 || Yast::Arch.i386) && grub2.name == "grub2-efi"
+      (Yast::Arch.x86_64 || Yast::Arch.i386 || Yast::Arch.aarch64) && grub2.name == "grub2-efi"
     end
 
     def trusted_boot_widget?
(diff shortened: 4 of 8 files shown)


src/lib/bootloader/grub_install.rb
```diff
@@ -87,6 +87,13 @@ def basic_cmd(secure_boot, trusted_boot)
 
       cmd << "--no-nvram" << "--removable" if removable_efi?
 
+      # FIXME: Grub does boot, but is unconfigured
+      if trusted_boot && Yast::Arch.aarch64
+        cmd << "&&" << "cp" << "/usr/share/grub2/arm64-efi/grub.efi" << "/boot/efi/EFI/opensuse"
```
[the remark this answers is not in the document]
ggardet: Ok, I will rework this part.
Thanks for the review.

src/lib/bootloader/grub_install.rb
```diff
@@ -90,6 +94,26 @@ def basic_cmd(secure_boot, trusted_boot)
...
+    def fixup_no_shim_secureboot(secure_boot, trusted_boot)
+      # Fixup only for secureboot on aarch64
+      return if !secure_boot || !Yast::Arch.aarch64
+      if trusted_boot
+        cmd = ["cp", "/usr/share/grub2/arm64-efi/grub-tpm.efi", "/boot/efi/EFI/*suse/bootaa64.efi"]
+      else
+        cmd = ["cp", "/usr/share/grub2/arm64-efi/grub.efi", "/boot/efi/EFI/*suse/bootaa64.efi"]
+      end
+      Yast::Execute.on_target(cmd)
+      []
```
jreidinger: why is this empty array needed?

src/lib/bootloader/grub_install.rb
```diff
@@ -90,6 +94,26 @@ def basic_cmd(secure_boot, trusted_boot)
...
 
+    # Fixup for SecureBoot without shim, on aarch64 (boo#1136601)
+    # FIXME: remove me when/if 'grub2-install' command supports secureboot
+    def fixup_no_shim_secureboot(secure_boot, trusted_boot)
+      # Fixup only for secureboot on aarch64
+      return if !secure_boot || !Yast::Arch.aarch64
+      if trusted_boot
+        cmd = ["cp", "/usr/share/grub2/arm64-efi/grub-tpm.efi", "/boot/efi/EFI/*suse/bootaa64.efi"]
+      else
+        cmd = ["cp", "/usr/share/grub2/arm64-efi/grub.efi", "/boot/efi/EFI/*suse/bootaa64.efi"]
```
jreidinger: does this *suse works? Execute does not use shell for expansion of path.

jreidinger:
@afaerber Can you please check it as arm expert if it is reasonable approach?

src/lib/bootloader/grub_install.rb
ggardet: I do not know, I am currently trying to get some review of the idea before trying it.
Any idea how to get the folder name for `/boot/efi/EFI/` ? It should be suse or opensuse, depending on the flavor.

jreidinger: let me show example code in irb ( irb is your ruby friend ). I create /tmp/suse and /tmp/opensuse.
```
irb(main):001:0> Dir["/tmp/*suse"]
=> ["/tmp/opensuse", "/tmp/suse"]
```
so `Dir[]` with glob returns array of matched entries. If there is just one, then I suggest to use `Dir["/boot/efi/EFI/*suse"].first`.

ggardet: With my latest update, I just make use of `--suse-signed-grub` option for `grub2-mkimage`.

src/lib/bootloader/grub_install.rb
```diff
@@ -72,10 +72,12 @@ def report_failure(exception)
     # creates basic command for grub2 install without specifying any stage1
     # locations
     def basic_cmd(secure_boot, trusted_boot)
-      if secure_boot
+      if secure_boot && !Yast::Arch.aarch64
         cmd = ["/usr/sbin/shim-install", "--config-file=/boot/grub2/grub.cfg"]
       else
         cmd = ["/usr/sbin/grub2-install", "--target=#{target}"]
+        # On aarch64, we do not use shim, but '--suse-signed-grub' option
```
jreidinger: please reference there link to bugzilla entry like `(bsc#1136601)`. It will help in future for reasoning and more detailed info. Similar way how it was at https://github.com/yast/yast-bootloader/pull/565/files#diff-b0238c1d8aa9d0846b284a3b971bc543L64.

jreidinger:
in general look good here. I just need to wait till new grub2 reaches TW as we need to send it together.

afaerber:
@jreidinger The logic changes look good to me (we may want to extend the architecture list later for 32-bit or other non-x86 archs, this becomes easier now), and the `grub2-install` approach was the requested one by the team. There were discussions to use that also for x86 as a wrapper for consistency, but changes can always be done incrementally. Can't fully review the Ruby code myself.

src/lib/bootloader/grub_install.rb
ggardet: Done

jreidinger:
@afaerber thanks for info. Ruby code looks good. So now I am just waiting till modified grub will be send to factory.

PR #566, opened 2019-05-29
Set kernel parameters without resume parameter. (JIRA#SLE-6926)

schubi2:


src/lib/bootloader/grub2base.rb
```diff
@@ -114,7 +114,8 @@ def propose
       propose_encrypted
 
       if grub_default.kernel_params.empty?
-        kernel_line = Yast::BootArch.DefaultKernelParams(propose_resume)
+        # Set kernel parameters without "resume" parameter. (JIRA#SLE-6926)
+        kernel_line = Yast::BootArch.DefaultKernelParams("")
```
jreidinger: sorry, this is wrong. JIRA speak only about mainframe ( so s390 ). Here it will disable it for all. Also please try to modify it in method propose_resume and not here. ( as it is for s390 specifically. BTW another option is also to extend that BootArch::ResumeAvailable.

PR #569, opened 2019-05-29
Adapt Rakefile and Dockerfile for SLE-12-SP5

dgdavid:
### Problem

`Rakefile` and `Dockerfile` were not adapted to use `SLE-12-SP5`

I realized about it trying to do an SR via `rake osc:sr`, which actually generated a [MR for SLE-12-SP4](https://build.suse.de/request/show/193762) (already revoked).

### Solution

Properly adapt those files.


dgdavid:
Submit request created, 

https://build.suse.de/request/show/193959

PR #573, opened 2019-07-03
Replace --suse-signed-grub with --suse-force-signed to follow update from boo#1136601

ggardet:
Grub2 update to Factory https://build.opensuse.org/request/show/713148


dgdavid:
Related to changes introduced in #565

PR #574, opened 2019-07-19
Added "BuildRequires: update-desktop-files"

lslezak:
- The recent fix in YaST RPM macros now updates the desktop files
- The `%suse_update_desktop_file` macro is defined in the `update-desktop-files` package
- Without it the macro is not expanded and the shell treats `%` as a job control character and fails with error `fg: no job control`


DimStar77:
alternative, more generic fix: https://github.com/yast/yast-devtools/pull/146

PR #575, opened 2019-07-23
Do not abort the upgrade if previous Grub config is broken

ancorgs:
### Problem

Reported as [bug#1138930](https://bugzilla.suse.com/show_bug.cgi?id=1138930) which was closed because the system under test was indeed broken.

When upgrading a system which contains a badly broken Grub2 configuration (e.g. contains references to udev links that do not longer exist), the process could indeed go wild in several points. All of them resulted in the upgrade being interrupted with an ugly internal error.

The first problem could arise in the summary screen. Like shown in this screenshots.

bootloader_error_in_proposal

If the error didn't pop-up or if user managed to recover from it, it was possible to reach the final phase of the upgrade process. But then the following internal error popped up.

error

For more details see https://trello.com/c/MMbbUuwZ/1140-2-sle12-sp5-p2-1138930-build-00770201-openqa-test-fails-in-awaitinstall

### Solution

In the summary screen, catch the unhandled exception and display a proper warning with a tip on what to do to fix the problem. The details provided by the exception (which are pretty informative) are displayed as well.

rescue_proposal

The user can ignore the problem or click on "booting" to fix it. In the latter case, the usual (already existing before this pull request) pop-up with instructions will appear.

rescue_dialog

If the final stage of the upgrade process is reached without fixing the error, the exception is now handled and the internal error is replaced by an informative message that does not interrupt the process.

rescue_finish

### DIFF
--- a/src/lib/bootloader/proposal_client.rb
+++ b/src/lib/bootloader/proposal_client.rb
@@ -84,12 +84,12 @@ module Bootloader
         option = chosen_id[/(enable|disable)_boot_(.*)/, 2]
         single_click_action(option, value)
       else
-        settings = Yast::Bootloader.Export
+        settings = export_settings
         result = ::Bootloader::MainDialog.new.run_auto
-        if result != :next
-          Yast::Bootloader.Import(settings)
-        else
+        if result == :next
           Yast::Bootloader.proposed_cfg_changed = true
+        elsif settings
+          Yast::Bootloader.Import(settings)
         end
       end
       # Fill return map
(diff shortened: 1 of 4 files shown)


jreidinger:
OK, in general looks good. Just that use yast2-bootloader looks a bit strange in installation context.
Way how to reproduce can be e.g. to add to grub_installdevice some non-existing udev link during rpm install ( so before install phase but after initial partitioning )

src/lib/bootloader/proposal_client.rb
```diff
@@ -197,6 +195,32 @@ def construct_proposal_map
...
+    # is raised while calculating the proposal
+    #
+    # @param err [Bootloader::BrokenConfiguration] raised exception
+    # @return [Hash]
+    def broken_configuration_proposal(err)
+      {
+        "label_proposal" => [],
+        "warning_level"  => :error,
+        # TRANSLATORS: %s is a string containing the technical details of the error
+        "warning"        => _("Error reading the bootloader configuration files. %s.") % err.reason
```
jreidinger: well, this is kind of new as it can happen only during upgrade where we read previous configuration.
But it should contain what user can do? Ideally user should click on "Booting" and here he should be able to repropose configuration from scratch, same as during runtime.

So I suggest to adapt error message here and also allow user to repropose after clicking on Booting.

ancorgs: What I'm fixing right now is that if you click on "Booting" the error pops-up. So the exception has to be captured in a third place.

PR #578, opened 2019-07-25
Fix test aarch

jreidinger:


package/yast2-bootloader.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Jul 25 13:21:44 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
+
+- Fix testsuite on AArch64. Related to changes done for bsc#1138930
```
ancorgs: Why is this related to the changes of bsc#1138930? I don't see the direct connection.

jreidinger: well, because previously if on all architecture if secure boot is passed, then shim-install is used. Now for aarch it using instead some special parameter for grub.

ancorgs: But that is bug#1136601, not bug#1138930

PR #581, opened 2019-09-30
Autoyast secure boot

jreidinger:
https://trello.com/c/i6UZefuO/1341-feature-jscsle-8787-autoyast-option-for-non-secure-boot
JSC#SLE-8787

### DIFF
--- a/src/lib/bootloader/autoyast_converter.rb
+++ b/src/lib/bootloader/autoyast_converter.rb
@@ -35,6 +35,7 @@ module Bootloader
         data["global"] ||= {}
 
         import_grub2(data, bootloader)
+        import_grub2efi(data, bootloader)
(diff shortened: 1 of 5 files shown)


src/lib/bootloader/autoyast_converter.rb
```diff
@@ -207,6 +220,17 @@ def bootloader_from_data(data)
         end
       end
 
+      # only for grub2, not for others
+      GRUB2EFI_BOOLEAN_MAPPING = {
```
joseivanlopez: NP: Should it be a private constant?

jreidinger: good point, also others are not private, so lets mark all of them as private.

PR #583, opened 2019-11-25
Abort the module execution when running w/o enough privileges

dgdavid:
## Problem

When `yast-bootloader` module is executed without enough privileges/permissions, it crash because an `Errno::EACCES` error when trying access to the _grub.cfg_ file.

<details>
<summary>Click to show/hide an screenshot</summary>

<hr/> 

bootloader_crash

</details>

## Solution

Abort the execution _nicely_, giving human readable information to the user.

:bulb:  **Proposed message:** 

> The module is running without enough privileges to perform all possible actions.
> 
> Cannot continue. Please, try again as root.

<hr/> 

bootloader_aborting_nicely

## Tests

* Tested manually in a running system.

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Nov 26 12:22:50 UTC 2019 - David Diaz <dgonzalez@suse.com>
+
+- Abort the execution when the module run without enough
+  permissions (related to bsc#1137688).
+- 4.2.13
+
+-------------------------------------------------------------------
 Wed Nov 20 16:07:58 UTC 2019 - Ancor Gonzalez Sosa <ancor@suse.com>
 
 - Improved detection of the root filesystem in non-standard
--- a/src/modules/Bootloader.rb
+++ b/src/modules/Bootloader.rb
@@ -135,7 +135,7 @@ module Yast
       return false if testAbort
 
       begin
-        ret = ::Bootloader::BootloaderFactory.current.read
+        ::Bootloader::BootloaderFactory.current.read
       rescue ::Bootloader::UnsupportedBootloader => e
         ret = Yast::Report.AnyQuestion(_("Unsupported Bootloader"),
           _("Unsupported bootloader '%s' detected. Use proposal of supported configuration instead?") %
@@ -166,13 +166,17 @@ module Yast
 
         ::Bootloader::BootloaderFactory.current = ::Bootloader::BootloaderFactory.proposed
         ::Bootloader::BootloaderFactory.current.propose
-      rescue ::Bootloader::InsufficientPrivileges => e
-        Yast2::Popup.show(e.message, headline: :error)
+      rescue Errno::EACCES
+        # TRANSLATORS: warn to the user that is running the module without enough permissions
+        error_msg = _(
+          "The module is running without enough privileges to perform all possible actions.\n\n" \
+          "Cannot continue. Please, try again as root."
+        )
+
+        Yast2::Popup.show(error_msg, headline: :error)
 
         return false
       end
-
-      return false unless ret
 
       Progress.Finish
 
--- a/src/modules/Bootloader.rb
+++ b/src/modules/Bootloader.rb
@@ -167,13 +167,16 @@ module Yast
         ::Bootloader::BootloaderFactory.current = ::Bootloader::BootloaderFactory.proposed
         ::Bootloader::BootloaderFactory.current.propose
       rescue Errno::EACCES
-        # TRANSLATORS: warn to the user that is running the module without enough permissions
-        error_msg = _(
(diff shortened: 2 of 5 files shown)


src/modules/Bootloader.rb
```diff
@@ -166,8 +166,14 @@ def Read
 
         ::Bootloader::BootloaderFactory.current = ::Bootloader::BootloaderFactory.proposed
         ::Bootloader::BootloaderFactory.current.propose
+      rescue ::Bootloader::InsufficientPrivileges => e
```
dgdavid: ## A self pre-review / Request for comments

I added this new exception just to follow the pattern used here (I mean, rescuing its own exception). But, to be honest, I'm not sure if

* I like it
* It is a good idea

The other way I see to do it without a custom exception is to simply rescue the `Errno::ACCES` here directly.

```ruby

...
rescue Errno::ACCES
  error_msg = _(
    "The module is running without enough privileges to perform all possible actions.\n\n" \
    "Cannot continue. Please, try again as root."
  )

  Yast2::Popup.show(error_msg, headline: :error)

  return false
end

...
```

Shorter than the previous option but, again, I'm not sure if

* I like it
* It is a good idea

**What do you think? Do you have any other suggestions?**

Thanks in advance for your input!

ancorgs: I agree there is no need to invent a new type of exception that is only used to directly replace `Errno::ACCESS`.

The current custom exceptions defined in yast2-bootloader exist because they actually describe a pretty specific circumstance that is not covered by the standard ruby. But that's not the case with that new (and unnecessary) `InsufficientPrivileges` class.

ancorgs:
Thinking about this more widely, I wonder whether we should just use `Yast::Confirm.MustBeRoot` at the beginning of the wizard/client instead of waiting for the permission problems to appear later.

I'd say that's what most traditional YaST modules do.

dgdavid:
@ancorgs 


I didn't know about `Yast::Confirm.MustBeRoot` :roll_eyes:  However, after some small tests, I don't think it will be a "real" solution for this issue. It only warning to the user that "**if you continue**, things might not work". But allows to continue.

<details>
<summary>Click to show/hide</summary>

yast_confirm_mustberoot

</details>


And what I'd like to do is abort the execution. To me, it makes no sense to 


>  - YaST: "hey, if you continue something wrong could happen"
>  - User: "it doesn't matter, let's _Continue_"
>  - YaST: "ok, here you have an error"

  <details>
  <summary>Click to show/hide</summary>

  yast_confirm_mustberoot_but_continue

  </details>

or even worst


>  - YaST: "hey, if you continue something wrong could happen"
>  - User: "it doesn't matter, let's _Continue_"
>  - YaST: "you don't know what are you doing my dear user. Let's _Abort_"


Am I missing something? Sorry for the noise :(

ancorgs:
Yes, you are right. The approach you implemented makes more sense from a user POV. But, please add a comment in that `rescue` block mentioning that not using `MustBeRoot` was a conscious decision.

package/yast2-bootloader.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Nov 26 12:22:50 UTC 2019 - David Diaz <dgonzalez@suse.com>
+
+- Abort the execution when the module run without enough
+  permissions (related to bsc#1137688).
```
ancorgs: It's actually hard to see the relationship with bsc#1137688, but I trust you ;-)

src/modules/Bootloader.rb
```diff
@@ -166,6 +166,19 @@ def Read
 
         ::Bootloader::BootloaderFactory.current = ::Bootloader::BootloaderFactory.proposed
         ::Bootloader::BootloaderFactory.current.propose
+      rescue Errno::EACCES
+        # If the access to any needed file (e.g., grub.cfg when using GRUB bootloader) is not
+        # allowed, just abort the execution. Using Yast::Confirm.MustBeRoot early in the
+        # wizard/client is not enough since it allows continue.
+
+        Yast2::Popup.show(
+          # TRANSLATORS: warn to the user that is running the module without enough permissions
```
ancorgs: Is really that useful to have a message to the translators to tell them what is already said in the string itself?

Something like `TRANSLATORS: pop-up message, beware the line breaks` would actually be useful, but the current one...

package/yast2-bootloader.changes
dgdavid: Yes, it's hard also to me. It's what I said in [a comment](https://trello.com/c/nhBiaGEa/1103-2-yast2-systemsettings-crashes-with-internal-error#comment-5ddbaa0431bc5f8fe0393e55) in the Trello card :man_shrugging:

src/modules/Bootloader.rb
dgdavid: Answer to the first question: actually no.

Thanks for the suggestion :)

PR #588, opened 2020-01-23
Improve error message for missing root 

joseivanlopez:
Improve the message when the root file system is not found. Now it suggests to unmount bind-mounted devices.

See https://bugzilla.suse.com/show_bug.cgi?id=1160176#c5 and https://github.com/yast/yast-bootloader/pull/586.

### DIFF
--- a/src/lib/bootloader/config_dialog.rb
+++ b/src/lib/bootloader/config_dialog.rb
@@ -27,7 +27,8 @@ module Bootloader
     rescue ::Bootloader::NoRoot
       Yast::Report.Error(
         _("YaST cannot configure the bootloader because it failed to find the root file system.\n" \
-          "This usually happens when there are bind-mounted devices.")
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/config_dialog.rb
```diff
@@ -27,7 +27,8 @@ def run
     rescue ::Bootloader::NoRoot
       Yast::Report.Error(
         _("YaST cannot configure the bootloader because it failed to find the root file system.\n" \
-          "This usually happens when there are bind-mounted devices.")
+          "This usually happens when there are bind-mounted devices. Please, unmount your bind \n" \
+          "mounts first and try again. You can mount it back after configuring your bootloader.")
```
ancorgs: Shouldn't it be "You can mount THEM back"? It's referring to "your bind mounts" (plural).

joseivanlopez: Sure.

PR #591, opened 2020-02-21
Delegate mount by logic

joseivanlopez:
## Problem

This module and *yast2-storage-ng* have a very similar logic to obtain the most reasonable udev link of a device. Such duplication (as every code/logic duplication) is error prone.

## Solution

Now on, yast2-bootloader simply delegates the logic to obtain an udev link to yast2-storage-ng module.

These changes require https://github.com/yast/yast-storage-ng/pull/1050.

## Testing

Added new unit tests.

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Feb 21 11:10:24 UTC 2020 - José Iván López González <jlopez@suse.com>
+
+- Delegate the logic for calculating a device udev link to
+  yast2-storage-ng module.
+- 4.2.16
+
+-------------------------------------------------------------------
 Fri Jan  3 13:08:10 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
 
 - deduplicate kernel parameters after merging them
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -2,7 +2,7 @@
 Fri Feb 21 11:10:24 UTC 2020 - José Iván López González <jlopez@suse.com>
 
 - Delegate the logic for calculating a device udev link to
-  yast2-storage-ng module.
+  yast2-storage-ng module (related to bsc#1151075).
 - 4.2.16
 
 -------------------------------------------------------------------
--- a/src/lib/bootloader/udev_mapping.rb
+++ b/src/lib/bootloader/udev_mapping.rb
@@ -76,19 +76,6 @@ module Bootloader
       device.name
     end
 
-    def kernel_to_udev(dev)
-      device = Y2Storage::BlkDevice.find_by_name(staging, dev)
-      if device.nil?
-        log.error "Cannot find #{dev}"
-        return dev
-      end
-
-      result = udev_name_for(device)
-      log.info "udev device for #{dev.inspect} is #{result.inspect}"
-
-      result
-    end
-
     # Udev name for the device
     #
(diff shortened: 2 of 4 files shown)


package/yast2-bootloader.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Feb 21 11:10:24 UTC 2020 - José Iván López González <jlopez@suse.com>
+
+- Delegate the logic for calculating a device udev link to
+  yast2-storage-ng module.
```
jreidinger: issue number?

src/lib/bootloader/udev_mapping.rb
```diff
@@ -38,26 +38,15 @@ def to_kernel_device(dev)
 
     # Converts udev or kernel device (disk or partition) to udev name that fits best.
     #
-    # Here is description of strategy for finding the best possible udev persistent name.
     # There are three scenarios we consider:
     # S1. disk with boot configuration is moved to different PC
     # S2. disk dies and its content is loaded to new disk from backup
     # S3. path to disk dies and disk is moved to different one
```
jreidinger: I think this part of comment is also not needed as it is related to strategy. It is scenarios we try to support.

src/lib/bootloader/udev_mapping.rb
```diff
@@ -105,48 +94,28 @@ def kernel_to_udev(dev)
...
+    #
+    # @return [String]
     def udev_name_for(device)
-      mount_by_udev(device) ||
-        udev_by_label(device) ||
-        udev_by_uuid(device) ||
-        udev_by_id(device) ||
-        udev_by_path(device) ||
-        device.name
+      mount_by_udev(device) || device.name
```
jreidinger: NP: I am not sure if it is worth to have it in two methods as logic now is quite small.

test/udev_mapping_test.rb
```diff
@@ -42,56 +42,142 @@
...
+
+        allow_any_instance_of(Y2Storage::MountPoint).to receive(:preferred_mount_by)
+          .and_return(mount_by)
+      end
+
+      let(:mount_by_option) { :label }
+
+      it "returns the udev link according to the preferred mount by" do
+        expect(subject.to_mountby_device(device.name)).to eq("/dev/disk/by-label/DATA")
+      end
```
jreidinger: I think we should not double test in storage. Simply delegates to storage-ng is enough for me here. As it is out of module bonds.

package/yast2-bootloader.changes
joseivanlopez: No issue number. This is an internal change addressed by ourselves. There is no real bug associated to it, but only a code improvement. I guess we don't need a bug number for that, right?

src/lib/bootloader/udev_mapping.rb
joseivanlopez: Yes, I was not sure about removing it or not, because somehow it documents what scenarios bootloader expects to support. But I agree that is something that does not depend on bootloader anymore. I will drop it out.

src/lib/bootloader/udev_mapping.rb
joseivanlopez: Ok

test/udev_mapping_test.rb
joseivanlopez: Ok, I will simplify it.

package/yast2-bootloader.changes
jreidinger: it won't be submitted without it.

PR #592, opened 2020-02-21
support s390 secure boot (jsc#SLE-9425)

wfeldt:
## Task

- https://jira.suse.com/browse/SLE-9471
- https://jira.suse.com/browse/SLE-9425

Implement secure boot handling for s390.

## Implementation

- extend existing secure boot logic to s390
- bundle a number of system-dependend logic around secure boot into a new `Systeminfo` class
- use `/etc/sysconfig/bootloader::SECURE_BOOT` to store the setting; `grub2-install` is expected to read it from there
- the secure boot setting can be changed directly from the installation summary screen
- to match it visually, the same has been done for trusted boot
- when the secure boot setting is changed, a popup is shown on s390 to remind the admin to 
adjust the HMC
- on s390, start with the currently active secure boot setting; not the value from sysconfig; this way the user doesn't run into problems when the sysconfig value and the current state disagree

### Note

The implementation is incomplete as there's no info on the tools to use to handle secure boot 
from the s390 side so far. So the current implementation does

- assume every s390 system can do secure boot, and
- that the setting is always `off`

## Notes on `auto` setting

- see [jsc#SLE-9425](https://jira.suse.com/browse/SLE-9425?focusedCommentId=985905&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-985905)

## Todo

- some extensive integration testing

## Screenshots

bar1

bar2

bar3

bar4

### DIFF
--- a/src/lib/bootloader/grub2.rb
+++ b/src/lib/bootloader/grub2.rb
@@ -113,8 +113,8 @@ module Bootloader
         )
       ]
 
-      result.push secure_boot_summary if Systeminfo.secure_boot_available?
-      result.push trusted_boot_summary if Systeminfo.trusted_boot_available?
+      result.push secure_boot_summary if Systeminfo.secure_boot_available?(name)
+      result.push trusted_boot_summary if Systeminfo.trusted_boot_available?(name)
 
       locations_val = locations
       if !locations_val.empty?
--- a/src/lib/bootloader/grub2.rb
+++ b/src/lib/bootloader/grub2.rb
@@ -7,6 +7,7 @@ require "bootloader/mbr_update"
 require "bootloader/device_map"
 require "bootloader/stage1"
 require "bootloader/grub_install"
+require "bootloader/systeminfo"
 
 Yast.import "Arch"
 Yast.import "BootStorage"
@@ -17,6 +18,7 @@ module Bootloader
   class Grub2 < Grub2Base
     attr_reader :stage1
     attr_reader :device_map
+    attr_accessor :secure_boot
 
     def initialize
       super
@@ -30,6 +32,8 @@ module Bootloader
     # Read settings from disk, overwritting already set values
     def read
       super
+
+      @secure_boot = Sysconfig.from_system.secure_boot
 
       begin
         stage1.read
@@ -65,7 +69,9 @@ module Bootloader
 
       # powernv must not call grub2-install (bnc#970582)
       unless Yast::Arch.board_powernv
-        failed = @grub_install.execute(devices: stage1.devices, trusted_boot: trusted_boot)
+        failed = @grub_install.execute(
+          devices: stage1.devices, secure_boot: @secure_boot, trusted_boot: trusted_boot
+        )
         failed.each { |f| stage1.remove_device(f) }
         stage1.write
       end
@@ -77,6 +83,8 @@ module Bootloader
     def propose
       super
 
+      @secure_boot = Systeminfo.secure_boot_active?
+
       stage1.propose
       # for GPT add protective MBR flag otherwise some systems won't
       # boot, safer option for legacy booting (bnc#872054)
@@ -87,6 +95,8 @@ module Bootloader
 
     def merge(other)
       super
+
+      @secure_boot = other.secure_boot unless other.secure_boot.nil?
 
       @device_map = other.device_map if !other.device_map.empty?
 
@@ -100,12 +110,12 @@ module Bootloader
         Yast::Builtins.sformat(
           _("Boot Loader Type: %1"),
           "GRUB2"
-        ),
-        Yast::Builtins.sformat(
-          _("Enable Trusted Boot: %1"),
-          trusted_boot ? _("yes") : _("no")
         )
       ]
+
+      result.push secure_boot_summary if Systeminfo.secure_boot_available?
+      result.push trusted_boot_summary if Systeminfo.trusted_boot_available?
+
       locations_val = locations
       if !locations_val.empty?
         result << Yast::Builtins.sformat(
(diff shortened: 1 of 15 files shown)


src/lib/bootloader/grub2.rb
```diff
@@ -100,12 +110,12 @@ def summary(simple_mode: false)
...
           _("Boot Loader Type: %1"),
           "GRUB2"
-        ),
-        Yast::Builtins.sformat(
-          _("Enable Trusted Boot: %1"),
-          trusted_boot ? _("yes") : _("no")
         )
       ]
+
+      result.push secure_boot_summary if Systeminfo.secure_boot_available?(name)
```
jreidinger: just style, usually it is used `<<` instead of use.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -336,6 +338,19 @@ def init
     def store
       grub2.secure_boot = value
     end
+
+    def validate
+      return true if Yast::Mode.config ||
+        !Yast::Arch.s390 ||
+        value == Systeminfo.secure_boot_active?
```
jreidinger: NP: I find a bit more readable to have it each on separate line and separate return as it is not related. So something like
```suggestion
return true if Yast::Mode.config
return true unless Yast::Arch.s390
return true if value == Systeminfo.secure_boot_active?
```

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -336,6 +338,19 @@ def init
...
+
+      Yast::Popup.ContinueCancel(
+        _(
+          "Make sure the Secure Boot setting matches the configuration of the HMC.\n\n" \
+          "Otherwise this system will not boot."
```
jreidinger: Well, buttons are continue/cancel. So I expect here a bit hint to user what when they click it. Like `Cancel to edit it` or `Continue to use setting`.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -336,6 +338,19 @@ def init
...
+        )
+      )
```
jreidinger: One more usability note. There is method focus which is often used in validate to "highlight" given widget. So here it can be something like
```
res = Yast::Popup.ContinueCancel(...)
focus unless res
res
```

https://github.com/yast/yast-yast2/blob/master/library/cwm/src/lib/cwm/abstract_widget.rb#L187

src/lib/bootloader/grub2base.rb
```diff
@@ -45,13 +45,22 @@ class Grub2Base < BootloaderBase
...
+    #   @return [Boolean] current secure boot setting
+    attr_accessor :secure_boot
+
     # @!attribute console
     #   @return [::Bootloader::SerialConsole] serial console or nil if none
     attr_reader :console
 
+    # @!attribute stage1
+    #   @return [::Bootloader::Stage1, nil] bootloader stage1, if one is needed
+    attr_reader :stage1
```
jreidinger: this is a bit strange as grub2EFI does not have stage1. So original intention is to have it only in grub2.rb and if something needs to depend on it, it should check bootloader or ask if method is defined.

src/lib/bootloader/grub2_widgets.rb
wfeldt: Again, this just mirrors the trusted boot dialog. Consistency across a UI is an important point. It doesn't help to do similar things in a completely different way.

src/lib/bootloader/grub2_widgets.rb
wfeldt: I personally wouldn't have formatted it this way. But this is what our rubocop rules allow/enforce. To completely reformat it might even run into some complexity check failures.

src/lib/bootloader/grub2_widgets.rb
wfeldt: You mean like here?:

https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/grub2_widgets.rb#L388-L390

I've never seen a dialog like 'continue to continue' or 'cancel to cancel'. That's why there's a button text.

src/lib/bootloader/proposal_client.rb
```diff
@@ -337,10 +343,28 @@ def handle_errors(ret)
...
+              "Otherwise this system will not boot."
+            ),
+            headline: :warning, buttons: :ok
+          )
+        end
+      end
       log.info "single_click_action #{option} #{value.inspect} #{devices}"
 
-      devices.each do |device|
+      devices&.each do |device|
```
jreidinger: this starting to feel a bit too hacky, as it is related only to "boot_" arguments. What about moving it inside case like:
```suggestion
      bootloader = ::Bootloader::BootloaderFactory.current
      case option
      when /boot_/
        stage1 = bootloader.stage1
        devices = option == "boot_mbr" ? stage1.boot_disk_names : stage1.boot_partition_names
        log.info "single_click_action #{option} #{value.inspect} #{devices}"
        devices.each do |device|
          value ? stage1.add_udev_device(device) : stage1.remove_device(device)
        end
      when "trusted_boot"
        bootloader.trusted_boot = value
      when "secure_boot"
        bootloader.secure_boot = value
      end
      
      Yast::Bootloader.proposed_cfg_changed = true
    end
```

src/lib/bootloader/proposal_client.rb
```diff
@@ -337,10 +343,28 @@ def handle_errors(ret)
...
+      case option
+      when "boot_mbr"
+        devices = stage1.boot_disk_names
+      when "boot_boot"
+        devices = stage1.boot_partition_names
+      when "trusted_boot"
+        ::Bootloader::BootloaderFactory.current.trusted_boot = value
+      when "secure_boot"
+        ::Bootloader::BootloaderFactory.current.secure_boot = value
+        if value && Yast::Arch.s390
```
jreidinger: in validation it checks when there is a mismatch, but here it shows always when enabling. Why conditions are different?

src/lib/bootloader/systeminfo.rb
```diff
@@ -0,0 +1,109 @@
...
+        Yast::Arch.s390
+      end
+
+      # Check if secure boot is currently active on an s390 machine.
+      #
+      # The 'real' state, not any config file setting.
+      #
+      # @return [Boolean] true if 390x machine has secure boot enabled
+      def s390_secure_boot_active?
+        # FIXME: this is just a stub - replace with real code later
```
jreidinger: well, I am not sure I want to approve with stub as I worry it will live here for another 20 years :)

src/lib/bootloader/grub2base.rb
wfeldt: That's a valid point and I considered doing it that way. The usage is here:

https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/proposal_client.rb#L338-L339

I my view this is a matter of class design. This proposal client code should work with any bootloader. So, basically with Grub2Base. That's why I decided to move the stage1 method.

Inconsistent class interfaces are imho one of the reasons yast code so easily runs into exceptions.

src/lib/bootloader/systeminfo.rb
wfeldt: This is what was agreed on for the current task. There will be another PBI once we get the missing info.

src/lib/bootloader/proposal_client.rb
wfeldt: Because it is initially always off. Note also the different texts.

PR #593, opened 2020-03-19
Fix calculation of udev mapping (bsc#1166096)

ancorgs:
## Problem 

Recently yast2-booloader was modified to not contain the logic about what is the best file path to reference a device, but to delegate that calculation to yast2-storage-ng. See #591.

Unfortunately, the logic implemented in that pull request is not really a direct translation of the previous one. If the device contains no filesystem (for example, PReP partitions and bios_boot ones), the logic implemented in #591 fallbacks immediately to use the kernel device name instead of trying to use the most stable name of that device.

Additionally, we found out the new version implemented in #591 didn't write much useful information to the YaST logs, making the debugging process harder.

**References:**

 - https://bugzilla.suse.com/show_bug.cgi?id=1166096
 - https://trello.com/c/SFYLkAIX/1703-3-tw-1166096-openqa-yast2bootloader-invalid-path-by-id-string

## Solution

Use the new API implemented in https://github.com/yast/yast-storage-ng/pull/1069 to bring back the original logic.

Opportunity taken to add some debugging output.

## Note

This can only be merged after https://github.com/yast/yast-storage-ng/pull/1069

## Testing

Added a regression test for the reported bug (PReP partition).


src/lib/bootloader/udev_mapping.rb
```diff
@@ -91,22 +91,37 @@ def kernel_to_udev(dev)
...
+          log.info "udev_path: using the udev name of the configured mount_by"
+          mount_by_name
+        else
+          log.info "udev_path: using the preferred udev name for the filesystem"
+          filesystem.preferred_name
+        end
+      else
+        log.info "udev_path: not formatted, using preferred udev name for the block device"
+        device.preferred_name
+      end
```
jreidinger: well, to my taste it is still too much storage code and ideally it would be device.preferred_name and it will return what should be used. But eve this way it looks ok.

ancorgs: Yes, I was unsure about where to draw the line. But the previous version was exposing even more storage knowledge/internals, so I decided to not go too far in hiding it.

PR #594, opened 2020-03-23
support s390 secure boot (jsc#SLE-9425, jsc#SLE-9471, bsc#1166736)

wfeldt:
## Task

- https://trello.com/c/7ZxtAqIG

Handle new Secure Boot setting for zSeries (z15+) in YaST.

## See also

- this is a follow-up to https://github.com/yast/yast-bootloader/pull/592
- technical summary: https://bugzilla.suse.com/show_bug.cgi?id=1166736#c9

## Implementation

- if `/sys/firmware/ipl/has_secure` is 1, the proposal will enable secure boot support
- the current state is stored in `/etc/sysconfig/bootloader::SECURE_BOOT` ("yes" or "no")
- the user may change the setting
- there is a warning about z15+ requirement if secure boot support is turned on
- reworked help texts to try to explain the situation

## Screenshots

bar1

bar2

bar3

bar4

### DIFF
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -377,7 +377,7 @@ module Bootloader
       textdomain "bootloader"
     end
 
-    def labe
+    def label
       _("Enable &Trusted Boot Support")
     end
 
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -328,7 +328,22 @@ module Bootloader
     end
 
     def help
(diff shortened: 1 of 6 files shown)


src/lib/bootloader/grub2_widgets.rb
```diff
@@ -361,7 +377,7 @@ def initialize
       textdomain "bootloader"
     end
 
-    def label
+    def labe
```
jreidinger: this looks like typo

jreidinger:
beside typo looks good, just changelog and version bump missing.

src/lib/bootloader/grub2_widgets.rb
wfeldt: Strange. Good spotting!

PR #595, opened 2020-03-27
Revert udev

ancorgs:
## Problem

yast-bootloader used to contain its own logic for calculating the best udev link to reference a device. We though that logic was redundant with the one in yast-storage-ng, so it was decided to unify that.

It was done in #591. But then we found the logic was not actually equivalent between both modules (see [bug#1166096](https://bugzilla.suse.com/show_bug.cgi?id=1166096)). We tried to fix it in #593, but we found more cases in which the different philosophy makes the unification hard.

We are coming up with a definitive solution... but we are too close to the deadline for releasing a new stable version of SLE  and openSUSE Leap.

## Solution

Revert the changes done in #591 and #593, and keep the separated logic in yast-bootloader. We will try again to reunify it later.

## Testing

Unit tests are reverted to its previous state (just like the code).

Since openQA caught the problems introduced in the two previous pull requests, it should also be enough to verify the old behavior is back.


src/lib/bootloader/udev_mapping.rb
```diff
@@ -76,52 +92,61 @@ def udev_to_kernel(dev)
...
+    def udev_by_uuid(device)
+      device.udev_full_uuid
+    end
+
+    def udev_by_label(device)
+      device.udev_full_label
+    end
+
+    def udev_by_id(device)
+      device.udev_full_ids.first
```
jreidinger: sadly this is also source of bugs, because udev does not consider all ids equal and some are better then others :(

ancorgs: libstorage-ng filters many of the ids out. Only the reliable ones (following libstorage-ng criteria) are included in `#udev_full_ids`.

PR #596, opened 2020-04-02
limit s390 secure boot to SCSI disks (bsc#1168165)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1168165
- https://trello.com/c/W1LvJYKF

Secure boot is only possible on SCSI disks.

## Solution

There are two options to deal with this

1. Hide the 'secure boot' check box when zipl is not on a SCSI disk.
2. Uncheck 'secure boot' when zipl is not on a SCSI disk.

I went for 2. basically so it is still possible to influence the secure boot setting. Otherwise the user might get stuck if `/etc/sysconfig/bootloader::SECURE_BOOT` is `yes` for whatever reason and they can't disable it because the UI hides the option for that.

## Question

How best to check for SCSI devices? See

- https://github.com/yast/yast-bootloader/blob/4226947aba875921b82d5dff4f2c64c07047d759/src/lib/bootloader/systeminfo.rb#L139-L141

- Answered: https://bugzilla.suse.com/show_bug.cgi?id=1168165#c11

## See also

Nice secure boot overview:

- https://linux.mainframe.blog/secure-boot

### DIFF
--- a/src/lib/bootloader/systeminfo.rb
+++ b/src/lib/bootloader/systeminfo.rb
@@ -124,7 +124,10 @@ module Bootloader
       # @return [Y2Storage::Partition, NilClass] zipl partition
       def zipl_device
         staging = Y2Storage::StorageManager.instance.staging
-        mountpoint = Y2Storage::MountPoint.find_by_path(staging, "/boot/zipl").first
+        mountpoint =
+          Y2Storage::MountPoint.find_by_path(staging, "/boot/zipl").first ||
+          Y2Storage::MountPoint.find_by_path(staging, "/boot").first ||
+          Y2Storage::MountPoint.find_by_path(staging, "/").first
         mountpoint.filesystem.blk_devices.first
       rescue StandardError
         nil
--- a/src/lib/bootloader/systeminfo.rb
+++ b/src/lib/bootloader/systeminfo.rb
@@ -31,7 +31,7 @@ module Bootloader
       # @param bootloader_name [String] bootloader name
       # @return [Boolean] true if secure boot setting is available with this bootloader
       def secure_boot_available?(bootloader_name)
-        efi_used?(bootloader_name) || s390_secure_boot_supported?
+        efi_used?(bootloader_name) || s390_secure_boot_available?
       end
 
       # Check current trusted boot state.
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/systeminfo.rb
```diff
@@ -108,6 +118,31 @@ def s390_secure_boot_active?
       rescue StandardError
         false
       end
+
+      # The partition where zipl is installed.
+      #
+      # @return [Y2Storage::Partition, NilClass] zipl partition
+      def zipl_device
+        staging = Y2Storage::StorageManager.instance.staging
+        mountpoint = Y2Storage::MountPoint.find_by_path(staging, "/boot/zipl").first
```
jreidinger, line 127: does it found if e.g. whole /boot is ext? and so no specific mp for zipl?

src/lib/bootloader/systeminfo.rb
```diff
@@ -108,6 +118,31 @@ def s390_secure_boot_active?
...
+      # Check if device is a SCSI device.
+      #
+      # param device [Y2Storage::Partition, NilClass] partition device (or nil)
+      #
+      # @return [Boolean] true if device is a SCSI device
+      def scsi?(device)
+        # in lack of a better idea: check if device name starts with 'sd'
+        # alternatively: device.udev_ids.any?(/^scsi-/)
+        # or: device.udev_paths.any?(/-zfcp-/)
+        device.name.start_with?("/dev/sd")
```
jreidinger, line 142: hmm, name check is a bit strange. I expect that such kind of info storage-ng contain. @ancorgs @aschnell is there such method for device? Or do you use hwinfo for it?

src/lib/bootloader/systeminfo.rb
wfeldt: Hm, good point. I was assuming /boot/zipl is always separate.

src/lib/bootloader/systeminfo.rb
wfeldt: The info does not exist. I've asked IBM if this would do and they agreed.

ancorgs: We have `Disk#transport` that can be one of these: `UNKNOWN, SBP, ATA, FC, ISCSI, SAS, SATA, SPI, USB, FCOE`

And we also have `BlkDevice#bus` that returns the bus reported by hwinfo and, more generally, `BlkDevice#hwinfo` that contains all the information from hwinfo.

I kind of expect that any of those can be used.

wfeldt: Interesting. I Looked at the XML and found no usable data there.

wfeldt: Some more thoughts:

- `transport` is the scsi transport layer, so we might go for `FC` and hope that's our use-case here; IMHO testing for scsi in general looks more correct (and also easier testable)
- `bus` is not in libstorage-ng and so not logged in the device graph XML; that again makes it hard to debug as the actual value shows up nowhere in any logs

PR #600, opened 2020-04-23
Arm workaround

jreidinger:
bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1167015

trello: https://trello.com/c/b1jtZQ8R/1816-sles15-sp2-p2-1167015-a64fx-uefi-does-not-boot-grub-after-installation

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Thu Apr 23 08:53:49 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
 
-- add SLE15 SP2 only workaround for arm to make grub2 works for 
-  every device. It can potentionnaly break multi-boot (bsc#1167015)
+- add SLE15 SP2 only workaround for ARM to make GRUB2 works for
+  every device. It can potentially break multi-boot (bsc#1167015)
 - 4.2.23
 
 -------------------------------------------------------------------
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Apr 23 08:53:49 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- add SLE15 SP2 only workaround for arm to make grub2 works for 
+  every device. It can potentionnaly break multi-boot (bsc#1167015)
+- 4.2.23
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


test/grub_install_test.rb
```diff
@@ -92,7 +96,17 @@ def expect_grub2_install(target, device: nil, removable: false)
...
+
+        subject.execute(devices: [])
+      end
+
+      it "runs twice as removable and non removable on aarch64 with efi vars (bsc#1167015)" do
+        stub_arch("aarch64")
+        stub_efivars
+        allow(::File).to receive(:directory?).and_return(true)
+        expect_grub2_install("arm64-efi", removable: false)
+        expect_grub2_install("arm64-efi", removable: true)
```
WenhuaChang, line 109: Did the sequence matter ? Looks to me the removable should be in front of non-removable ? 
````
expect_grub2_install("arm64-efi", removable: true)
expect_grub2_install("arm64-efi", removable: false)
````

Comment (jreidinger) [reply test/grub_install_test.rb:109]: no, in this test order does not matter. I can do it in to be dependent on order, but if I get it, order is not important, just calling both.

File: package/yast2-bootloader.changes [hunk 2]
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Apr 23 08:53:49 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- add SLE15 SP2 only workaround for arm to make grub2 works for
```
imobachgs, line 4: What about "ARM"?

package/yast2-bootloader.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Apr 23 08:53:49 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- add SLE15 SP2 only workaround for arm to make grub2 works for 
+  every device. It can potentionnaly break multi-boot (bsc#1167015)
```
imobachgs, line 5: potentionnaly -> potentially

test/grub_install_test.rb
WenhuaChang: Yes, order doesn't matter as long as we only want both installed. I was misunderstanding the test that it would expect the command execution order to match what's programmed in the source code.

ggardet:
It breaks on Raspberry Pi 3 : https://bugzilla.opensuse.org/show_bug.cgi?id=1172114

jreidinger:
@ggardet thanks for notify me. We need to discuss how to deal with that conflicting requirements with arm/EFI/shim experts

PR #601, opened 2020-05-05
allow optional types for string and map

jreidinger:
trello: https://trello.com/c/HkFkUQHj/1791-3-continue-with-new-xml-parser-xml-validation

depends on https://github.com/yast/yast-autoinstallation/pull/598

Agreed to postpone for now that trang and jing travis validation to not delay even more new xml parser.

### DIFF
--- a/src/autoyast-rnc/bootloader.rnc
+++ b/src/autoyast-rnc/bootloader.rnc
@@ -2,6 +2,8 @@ default namespace = "http://www.suse.com/1.0/yast2ns"
 namespace a = "http://relaxng.org/ns/compatibility/annotations/1.0"
 namespace config = "http://www.suse.com/1.0/configns"
 
+include "common.rnc"
+
 STRING_BOOL = STRING_ATTR, ("true" | "false")
 
 bootloader =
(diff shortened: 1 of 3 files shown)


src/autoyast-rnc/bootloader.rnc
```diff
@@ -2,28 +2,36 @@ default namespace = "http://www.suse.com/1.0/yast2ns"
 namespace a = "http://relaxng.org/ns/compatibility/annotations/1.0"
 namespace config = "http://www.suse.com/1.0/configns"
 
+STRING_BOOL = STRING_ATTR, ("true" | "false")
```
mvidner, line 5: we should still
> include "common.rnc"


even if we don't run the validation yet

jreidinger: how it will behave in final build which include everything? will be common.rng included many times?

jreidinger:
@mvidner - to your question, no dependency, as modules should not depend on autoyast ( you do not want it on running system ). Also new references is not needed for trang, so only in yast2-schema it is good to have it together.

for BSC I created https://bugzilla.suse.com/show_bug.cgi?id=1170886

PR #602, opened 2020-05-07
Set autoinst clonable

imobachgs:
Sets the `AutoInstClonable` attribute as the special handling for this section was removed at https://github.com/yast/yast-autoinstallation/pull/604.


imobachgs:
> version bump?

Yes, you are right. I guess there is no reason to change the version. I did not the same in the autoyast2 counterpart because I expect we will release a new version soon. But in this case, it is different.

Thanks!

PR #605, opened 2020-05-27
Fix efivars workaround

jreidinger:
bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1172114

It is SLE15 SP2 only as it is intended as workaround till new shim will appear. Fix of https://github.com/yast/yast-bootloader/pull/600

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed May 27 12:07:31 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- fix detection of present of efivars causing grub2-install
+  failures on some arm boards (bsc#1172114)
+- 4.2.25
+
+-------------------------------------------------------------------
 Fri May 22 10:48:35 UTC 2020 - Steffen Winterfeldt <snwint@suse.com>
 
 - s390 secure boot: enhance disk type detection to cover multipath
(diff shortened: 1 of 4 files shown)


afaerber:
I note that I had previously suggested to tighten the condition of doing this to not just check for the sysfs dir but also for some, e.g., DMI system-level info that would only activate it on the one vendor's known affected systems. That would rule out side-effects on other systems such as Raspberry Pi or Jetson, but might need adjustments if further vendors were to ship an ... unexpected UEFI implementation.

jreidinger:
@afaerber well, I would try to keep it as simple as possible for yast and also perl-bootloader. I consider this only as workaround specific to SP2 which is not in TW, where I expect proper shim or handling it on grub2 level, because even for users that want to use grub2 itself, it will fail and they need to google what parameters they need to pass to grub, which is a bit user unfriendly for me.
Instead I expect that grub2-install just pick the right options and only when something goes wrong, there will be options to overwrite it like `--removable/no-removable` that force selection.

WenhuaChang:
IMHO, the cost of maintaining DMI as hardcoded table in the program "code" is not favorable, unless that could be supplied as aux data to the program that concerns it. But I believe grub or yast is lacking of such infrastructure in general of making DMI specific decisions ... implement that for workaround overkilled the idea of being as simple as possible (for not having other problems cracked in ..) 

Besides I believe providing default boot path as "fallback" location would benefit us for other circumstance in the long run. For x86, we experienced some systems would erase all boot variables after firmware update so the fallback helps in keeping us always bootable. :)

PR #606, opened 2020-05-29
Cleanup issue handling

schubi2:
Problem
==============

Each module (storage, firewall,network) has his own AutoYaST Issue and profile handling which is incompatible and doubles the code.
bsc#1171335

Solution
========

Bring base classes to yast2 package and cleanup the code of the other packages which will use
these common base classes:
yast/yast-yast2#1050
yast/yast-yast2#1055

### DIFF
--- a/src/lib/bootloader/autoinst_profile/bootloader_section.rb
+++ b/src/lib/bootloader/autoinst_profile/bootloader_section.rb
@@ -44,14 +44,14 @@ module Bootloader
       #   @see Bootloader::BootloaderFactory::SUPPORTED_BOOTLOADERS
 
       # @!attribute loader_device
-      #   @deprecated
+      #   @deprecated Replaced by `<boot_*>` elements in the `<global>` section.
 
       # @!attribute activate
       #   @see GlobalSection#activate
       #   @deprecated
 
       # @!attribute sections
-      #   @deprecated
+      #   @deprecated It still exists just to log a warning in AutoyastConverter.
 
       # @return [GlobalSection] 'global' section
       attr_accessor :global
--- a/src/lib/bootloader/autoinst_profile/bootloader_section.rb
+++ b/src/lib/bootloader/autoinst_profile/bootloader_section.rb
@@ -31,13 +31,9 @@ module Bootloader
       def self.attributes
         [
           { name: :loader_type },
-          { name: :loader_device }, # @deprecated
-          { name: :activate }, # @deprecated
-          { name: :sections }, # @deprecated
-          { name: :boot_boot_root }, # @deprecated
-          { name: :boot_boot_boot }, # @deprecated
-          { name: :boot_boot_mbr }, # @deprecated
-          { name: :boot_boot_extended } # @deprecated
+          { name: :loader_device }, # deprecated
+          { name: :activate }, # deprecated
+          { name: :sections } # deprecated
         ]
       end
 
@@ -55,18 +51,6 @@ module Bootloader
       #   @deprecated
 
       # @!attribute sections
-      #   @deprecated
-
-      # @!attribute boot_boot_root
-      #   @deprecated
-
-      # @!attribute boot_boot_boot
-      #   @deprecated
-
-      # @!attribute boot_boot_mbr
-      #   @deprecated
-
-      # @!attribute boot_boot_extended
       #   @deprecated
 
       # @return [GlobalSection] 'global' section
--- a/src/lib/bootloader/autoinst_profile/bootloader_section.rb
+++ b/src/lib/bootloader/autoinst_profile/bootloader_section.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # Copyright (c) [2020] SUSE LLC
 #
 # All Rights Reserved.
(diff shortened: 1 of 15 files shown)


src/lib/bootloader/autoinst_profile/bootloader_section.rb
```diff
@@ -0,0 +1,39 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "installation/autoinst_profile/section_with_attributes"
+
+module Bootloader
+  module AutoinstProfile
+    # This class represents an AutoYaST <bootloader> section
+    #
+    class BootloaderSection < ::Installation::AutoinstProfile::SectionWithAttributes
```
jreidinger, line 28: this looks a bit strange for me as it is not used for anything except issues. What about allowing to pass hash to issues and create instance of SectionWithAttributes there for Hash? Or adapt code to use it at more places instead of plain hash.

imobachgs: On of the objectives of the AutoYaST refactor is to improve how the profile is represented (using hashes everywhere is not that nice). So I will take the opportunity to extend this `BootloaderSection` so it can be used when importing a profile instead of a hash.

As far as I can see, the hash is used only in [::Bootloader::AutoyastConverter](https://github.com/yast/yast-bootloader/blob/038d3eecd16a6d24836b28a667723e2287eff5b8/src/lib/bootloader/autoyast_converter.rb), so perhaps we can  adapt that module

src/lib/bootloader/autoinst_profile/bootloader_section.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      # @!attribute loader_device
+      #   @deprecated
+
+      # @!attribute activate
+      #   @see GlobalSection#activate
+      #   @deprecated
+
+      # @!attribute sections
+      #   @deprecated
```
jreidinger, line 54: deprecated and also ignored, as in grub2 we do not support own sections.

src/lib/bootloader/autoinst_profile/bootloader_section.rb
```diff
@@ -0,0 +1,89 @@
...
+      end
+
+      define_attr_accessors
+
+      # @!attribute loader_type
+      #   @return [String] which boot loader to use (default, grub2, grub2-efi and none)
+      #   @see Bootloader::BootloaderFactory::SUPPORTED_BOOTLOADERS
+
+      # @!attribute loader_device
+      #   @deprecated
```
jreidinger, line 47: basically global/boot_* is replacement

src/lib/bootloader/autoinst_profile/device_map_entry_section.rb
```diff
@@ -0,0 +1,38 @@
...
+    # This class represents an AutoYaST <global> section within a <bootloader> one
+    class DeviceMapEntrySection < ::Installation::AutoinstProfile::SectionWithAttributes
+      def self.attributes
+        [
+          { name: :firmware },
+          { name: :linux }
+        ]
+      end
+
+      define_attr_accessors
```
jreidinger, line 35: I think both attributes need documentation :)

src/lib/bootloader/autoinst_profile/global_section.rb
```diff
@@ -0,0 +1,125 @@
...
+      define_attr_accessors
+
+      # @!attribute activate
+      #   @return [Boolean,nil] whether to set the _boot_ flag on the boot partition.
+
+      # @!attribute append
+      #   @return [String,nil] kernel parameters to add at the end of the boot entries.
+
+      # @!attribute boot_boot
+      #   @return [String,nil] write GRUB 2 to a separate `/boot` partition if it exists.
```
jreidinger, line 63: and if not exist , then it writes to "/"

src/lib/bootloader/autoinst_profile/global_section.rb
```diff
@@ -0,0 +1,125 @@
...
+      #     ("true" or "false")
+
+      # @!attribute boot_custom
+      #   @return [String,nil] name of device to write GRUB 2 to (e.g., "/dev/sda3").
+
+      # @!attribute boot_extended
+      #   @return [String,nil] write GRUB 2 to the extended partition ("true" or "false").
+
+      # @!attribute boot_mbr
+      #   @return [String,nil] write GRUB 2 to the MBR of the first disk in the device map
```
jreidinger, line 73: I am not sure if that condition about first disk in the device map is still valid. In fact it writes to all disk where "/boot" lives. So if it is in LVM or raid, it writes to all disk where such LVM or raid lives.

src/lib/bootloader/autoinst_profile/bootloader_section.rb
imobachgs: Yes, that's it. Actually, the method is there just because of [this warning](https://github.com/yast/yast-bootloader/blob/ef272ca255c45c06fc4fbe6a4ec9f1aef564e5d0/src/lib/bootloader/autoyast_converter.rb#L49). If you are fine with dropping the warning, we can drop this code too.

src/lib/bootloader/autoinst_profile/global_section.rb
imobachgs: OK, I will have to update the AutoYaST guide then.

src/lib/bootloader/autoinst_profile/bootloader_section.rb
jreidinger: I am fine without it. It should just not crash if it is there ( even if it s removed in SLE12 already )

PR #622, opened 2020-10-22
Nvram fixes

jreidinger:
bsc: https://bugzilla.suse.com/show_bug.cgi?id=1177986 and https://bugzilla.suse.com/show_bug.cgi?id=1157550#c60


hramrach:
Thanks for fixing

I was wondering if we need a test for this as well

dgdavid (approved):

PR #624, opened 2020-11-10
use more consistent wording

jreidinger:
Goal of this pull request is to make installation summary and UI of bootloader more consistent.

Lets start from the end. We discuss two approaches describe below and result of discussion with UX team is to use the first one and just adjust a bit wording. So final result is this one:

summary3

From discussion in yast team we try two approaches. One is current one just with better wording and the second one is to use in summary checkbox like approach which reduce text and make it more looking like checkbox configuration.

Example of approach 1:
qt 
qt_summary1
ncurses
ncurses_summary1

example of approach 2:
qt
qt_summary2
ncurses
ncurses_summary2

if we use approach 2 we probably should also make current defails more consistent and remove that write into from
ncurses_details1

but as said above approach 1 wins as it is more explicit for users what exactly happen and what is the change.

### DIFF
--- a/src/lib/bootloader/grub2.rb
+++ b/src/lib/bootloader/grub2.rb
@@ -188,79 +188,99 @@ module Bootloader
       )
     end
 
+    # Gets all location where stage1 will be written
     def locations
       locations = []
 
       partition_location = Yast::BootStorage.boot_partitions.map(&:name).join(", ")
-      locations << partition_location + _(" (/boot)") if stage1.boot_partition?
+      locations << partition_location if stage1.boot_partition?
       if stage1.extended_boot_partition?
         partitions = Yast::BootStorage.boot_partitions.map do |partition|
           Yast::BootStorage.extended_for_logical(partition).name
         end
-        locations << partitions.join(", ") + _(" (/boot)")
+        locations << partitions.join(", ")
       end
-      if stage1.mbr?
-        # TRANSLATORS: MBR is acronym for Master Boot Record, if nothing locally specific
-        # is used in your language, then keep it as it is.
-        locations << Yast::BootStorage.boot_disks.map(&:name).join(", ") + _(" (MBR)")
-      end
+      locations << Yast::BootStorage.boot_disks.map(&:name).join(", ") if stage1.mbr?
       locations << stage1.custom_devices if !stage1.custom_devices.empty?
 
       locations
     end
 
     def mbr_line
-      if stage1.mbr?
+      # TRANSLATORS: summary line where %s is disk specified, can be more disks, separated by comma
+      res = if stage1.mbr?
         _(
-          "Install bootcode into MBR (<a href=\"disable_boot_mbr\">do not install</a>)"
+          "Write into MBR of %s (<a href=\"disable_boot_mbr\">do not write</a>)"
         )
+      # TRANSLATORS: summary line where %s is disk specified, can be more disks, separated by comma
       else
         _(
-          "Do not install bootcode into MBR (<a href=\"enable_boot_mbr\">install</a>)"
+          "Do not write into MBR of %s (<a href=\"enable_boot_mbr\">write</a>)"
         )
       end
+
+      format(res, Yast::BootStorage.boot_disks.map(&:name).join(", "))
     end
 
     def partition_line
-      if stage1.boot_partition?
+      # TRANSLATORS: summary line where %s is partition specified, can be more disks,
+      # separated by comma
+      res = if stage1.boot_partition?
         _(
-          "Install boot code into a partition with /boot " \
-            "(<a href=\"disable_boot_boot\">do not install</a>)"
+          "Write into a partition with /boot - %s" \
+            "(<a href=\"disable_boot_boot\">do not write</a>)"
         )
+      # TRANSLATORS: summary line where %s is partition specified, can be more disks,
+      # separated by comma
       else
         _(
-          "Do not install boot code into a partition with /boot " \
-            "(<a href=\"enable_boot_boot\">install</a>)"
+          "Do not write into a partition with /boot - %s" \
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/grub2.rb
```diff
@@ -188,87 +188,107 @@ def disk_order_summary
...
+          "Write into an extended partition with /boot - %s" \
+            "(<a href=\"disable_boot_extended\">do not write</a>)"
         )
+      # TRANSLATORS: summary line where %s is partition specified, can be more disks,
+      # separated by comma
       else
         _(
-          "Do not install boot code into a extended partition with /boot " \
-            "(<a href=\"enable_boot_extended\">install</a>)"
+          "Do not into an extended partition with /boot - %s" \
```
wfeldt, line 275: write?

wfeldt:
Generally, two comments:

1. I would drop the article before 'partition', like "~~a~~ logical partition"
2. "Write into [~~an~~] extended partition with /boot" - I'm not sure that transposrts any obvious meaning to the non-initiated :-)

That doesn't mean I know better - just my feeling.

ancorgs:
So the proposal will look now like:

- Boot Code Location
  - Write into MBR of /dev/sda (**do not write**)
  - Do not write into an extended partition with /boot - /dev/sda4 (**write**)
  - Do not write into a logical partition with /boot - /dev/sda5 (**write**)

Apart from what @wfeldt already commented, I kind of miss an "it" next to "write". And maybe there is too much write, don't write... Maybe using just "change" for the action handler could be better.

 Something like this (just a proposal):

- Boot Code Location
  - Write it into MBR of /dev/sda (**change**)
  - Do not write it into extended partition with /boot - /dev/sda4 (**change**)
  - Do not write it into logical partition with /boot - /dev/sda5 (**change**)

ancorgs:
Or maybe I would just simplify the sentences about partitions. Something like this:

- Boot Code Location
  - Write it into MBR of /dev/sda (**change**)
  - Do not write it into extended partition /dev/sda4 (**change**)
  - Do not write it into logical partition /dev/sda5 (**change**)

src/lib/bootloader/grub2.rb
jreidinger: good catch, thanks

wfeldt:
I've the feeling we are doing something unnatural here. The expected thing would be

- boot code location
  - [x] mbr - dev/sda
  - [ ] logical partition with /boot - /dev/sda5
  - [x] extended partition with /boot - /dev/sda4

What's missing is the on/off semantics.

BTW, no 'it' ('install into /dev/sa', 'write to /dev/sda') - these are not complete sentences.

jreidinger:
@wfeldt do not forget that it is summary richtext, so any click/change result in recomputing proposal ( e.g. when you click write to partition and not write to MBR, it also add to software proposal e.g. syslinux. So checkboxes does not fit well.

But what we can have is something like

- boot code location
  - ✓mbr - dev/sda (change)
  - ✗logical partition with /boot - /dev/sda5 (change)
  - ✓extended partition with /boot - /dev/sda4 (change)

wfeldt:
The checkboxes were contributed by markdown and actually unintended. What I wanted to say was that this is a classical list of checkbox items and all the wording we put around it is just there to verbosely explain that.

wfeldt (approved):
There's the missing space Ancor already mentioned and I personally would drop the 'it' in 'write it into'. But if the UX team is fine with it, Im not the one to argue.

So, lgtm (but add the space).

src/lib/bootloader/grub2.rb
```diff
@@ -188,87 +188,107 @@ def disk_order_summary
       )
     end
 
+    # Gets all location where stage1 will be written
```
ancorgs, line 191: NP: locations, in plural

src/lib/bootloader/grub2.rb
```diff
@@ -188,87 +188,107 @@ def disk_order_summary
...
+      # TRANSLATORS: summary line where %s is partition specified, can be more disks,
+      # separated by comma
+      res = if stage1.extended_boot_partition?
         _(
-          "Install boot code into a extended partition with /boot " \
-            "(<a href=\"disable_boot_extended\">do not install</a>)"
+          "Write it into extended partition with /boot - %s " \
+            "(<a href=\"disable_boot_extended\">do not write</a>)"
         )
+      # TRANSLATORS: summary line where %s is partition specified, can be more disks,
```
ancorgs, line 271: To make sure the translators get this comment, shouldn't it be after the "else"?

src/lib/bootloader/grub2.rb
```diff
@@ -188,87 +188,107 @@ def disk_order_summary
...
+          "Do not write it into logical partition with /boot - %s " \
+            "(<a href=\"enable_boot_boot\">write</a>)"
         )
       end
+      format(res, Yast::BootStorage.boot_partitions.map(&:name).join(", "))
     end
 
     def extended_partition_line
-      if stage1.extended_boot_partition?
+      # TRANSLATORS: summary line where %s is partition specified, can be more disks,
```
ancorgs, line 264: Can you please check if this comment is actually copied to the .po in the expected place?

ancorgs:
Looks good in general. I just added a couple of comments, but feel free to merge if they are not relevant.

src/lib/bootloader/grub2.rb
jreidinger: yes, it is there. Check is easy. just run `rake pot` and it will generate pot file that is used by translators

src/lib/bootloader/grub2.rb
jreidinger: our pot generator use the closest comment before. so it works fine.

jreidinger:
typo in comment fixed.

jreidinger:
image update to the final version

PR #629, opened 2021-01-21
Fix test

joseivanlopez:
### Problem

The logic to calculate the udev device names was delegated to *yast-storage-ng*, see https://github.com/yast/yast-bootloader/pull/628. Such logic depends on the default *mount_by* option. That value is set in *storage-ng* according to the architecture. There are some test failures because the different architectures were not considered.

* https://bugzilla.suse.com/show_bug.cgi?id=1181175

### Solution

Force the same default *mount_by* value in the unit tests, independently on the real architecture. Note that *yast-storage-ng* already contains a full set of exhaustive tests for this logic.


ancorgs (approved):
LGTM

PR #630, opened 2021-01-26
Fix recalculation when storage has changed (bsc#1180218, bsc#1180976)

ancorgs:
## Problem

YaST bootloader cannot properly update its proposal to reflect changes in the swap configuration. That's visible in two situations reported as separate bug reports:

- If the swap partition is substituted with another with a different name, the `resume=` parameter gets duplicated (one with the previous name and another with the current one) - [bsc#1180218](https://bugzilla.suse.com/show_bug.cgi?id=1180218)
- If the swap partition is deleted, the `resume=` parameter is not removed - [bsc#1180976](https://bugzilla.suse.com/show_bug.cgi?id=1180976)

In theory, the bootloader proposal should be totally discarded and recreated from scratch every time the storage setup changes. But there was an error in the code and the change was generally not detected.

## Solution

Fixed the detection of changes in storage. Let's see how.

The call to `Yast::BootStorage.boot_filesystem` can have the side effect of updating the `@storage_revision` variable. That was causing the subsequent call to `Yast::BootStorage.storage_changed?` to always return `false`.

Now both calls are properly sorted, so the possible side effect of `Yast::BootStorage.boot_filesystem` does not longer affect the detection.

## Testing

- Added new regression unit test
- Checked manually that this fixes [bsc#1180976](https://bugzilla.suse.com/show_bug.cgi?id=1180976)
- Checked manually that this fixes [bsc#1180218](https://bugzilla.suse.com/show_bug.cgi?id=1180218)

### DIFF
--- a/src/lib/bootloader/proposal_client.rb
+++ b/src/lib/bootloader/proposal_client.rb
@@ -133,6 +133,9 @@ module Bootloader
     def make_proposal_raising(attrs)
       force_reset = attrs["force_reset"]
       storage_read = Yast::BootStorage.storage_read?
+      # This must be checked at the beginning because the call to BootStorage.boot_filesystem
+      # below can trigger a re-read and change the result of BootStorage.storage_changed?
+      # See bsc#1180218 and bsc#1180976.
       storage_changed = Yast::BootStorage.storage_changed?
 
       if Yast::BootStorage.boot_filesystem.is?(:nfs)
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/proposal_client.rb
```diff
@@ -131,13 +131,15 @@ def description
 
     # make proposal without handling of exceptions
     def make_proposal_raising(attrs)
+      force_reset = attrs["force_reset"]
+      storage_read = Yast::BootStorage.storage_read?
+      storage_changed = Yast::BootStorage.storage_changed?
```
jreidinger, line 136: please add note to code about that order and side effects. Otherwise in future it will be again accidentally broken.

jreidinger (approved):

PR #631, opened 2021-01-27
Do not add resume when swap is too small

joseivanlopez:
### Problem

When the swap is smaller than the RAM size, suspending to RAM could not work. But the *resume=* kernel parameter is being added even in that cases.

* https://bugzilla.suse.com/show_bug.cgi?id=1180977
* https://trello.com/c/VkjmNIRr/2279-3-yast-bootloader-adds-the-resume-parameter-when-it-shouldnt

### Solution

The *resume=* kernel parameter is not added if the swap size is smaller than the RAM size.

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jan 27 10:59:26 UTC 2021 - José Iván López González <jlopez@suse.com>
+
+- Do not propose resume kernel parameter when the swap is smaller
+  than the RAM size (bsc#1180977).
+- 4.3.20
+
+-------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


package/yast2-bootloader.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jan 27 10:59:26 UTC 2021 - José Iván López González <jlopez@suse.com>
+
+- Do not propose resume kernel parameter when the swap is smaller
+  than the RAM size (bsc#1180977).
```
jreidinger, line 5: it makes sense to not propose it. Just description that it is not possible is a bit strong as not always whole ram is filled.

jreidinger (approved):

PR #632, opened 2021-02-09
use shim also on aarch (jsc#SLE-15020)

jreidinger:
trello: https://trello.com/c/E4U5cNa9/2289-3-feature-musthave-secure-boot-support-in-yast-on-aarch64

jsc: https://jira.suse.com/browse/SLE-15823

Testing: Only unit tested as failed to test secure boot in qemu and do not have hardware around that allows such test.


wfeldt (approved):
lgtm

I guess I'll have to adjust perl-Bootloader as well, then.

PR #633, opened 2021-02-15
Fix popup

jreidinger:
This pull request in yast-yast2 (https://github.com/yast/yast-yast2/pull/1122) changed the behavior during unit tests of `Yast::Report` and `Yast2::Popup`, making necessary to add some extra mocking to prevent YaST from trying to open windows during the test execution.

As far as we know, that affected the test-suites of: yast-storage-ng, yast-country, yast-firewall, yast-nfs-server, yast-nis-client, yast-pam, yast-security, yast-services-manager and yast-sysconfig.

This should fix the problem for this repository.

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Feb 15 14:49:00 UTC 2021 - Josef Reidinger <jreidinger@suse.com>
+
+- Adapted unit test to recent changes in Yast::Report (related to
+  bsc#1179893).
+- 4.3.22
+
+-------------------------------------------------------------------
 Tue Feb  9 15:44:42 UTC 2021 - Josef Reidinger <jreidinger@suse.com>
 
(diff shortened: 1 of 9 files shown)


ancorgs:
Taking into account that yast-bootloader is expected to have a solid test-suite (with 88.5% test coverage at the moment), I'm not sure if just adding `allow(Yast::Report).to receive(:Error)` on relative wide contexts is the right solution here. Looks a bit like just saying the testsuite to shut up instead of taking the opportunity to improve it.

jreidinger:
@ancorgs well, generic issue with bootloader is that majority of code expects grub2 installed, so it call some harmless detection of default states and so on. It works fine if that detection failed ( aka return empty result ), but Yast::Execute opens popup if that detection failed in unexpected ways like that grub2-env is missing. So I tried more localized fixes, but it is not trivial. I will try tomorrow if it can be done somewhere in between.

ancorgs (approved):
LGTM

PR #635, opened 2021-03-03
Ro efivars

jreidinger:
Issue: Checking just for presence of efivars is not enough as it can be mounted as read only for whatever reason.
Fix: Check also if it is mounted as readwrite.

Tested also manually with rw efivars and works.

related PBIs: 
https://trello.com/c/GSWoJy3g/2305-sles15-sp3-p1-1182749-secure-boot-is-enabled-on-aarch64-even-when-disabled-on-test-machine
https://trello.com/c/ru1DFOGu/2313-bug-1174111-jetson-agx-xavier-grub2-install-could-not-prepare-boot-variable-invalid-argument

### DIFF
--- a/test/grub_install_test.rb
+++ b/test/grub_install_test.rb
@@ -12,15 +12,7 @@ describe Bootloader::GrubInstall do
     end
 
     def stub_efivars(removable: false)
-      efivardirs = if removable
-        []
-      else
-        ["Boot0000-8be4df61-93ca-11d2-aa0d-00e098032b8c",
-         "BootCurrent-8be4df61-93ca-11d2-aa0d-00e098032b8c",
-         "BootOrder-8be4df61-93ca-11d2-aa0d-00e098032b8c"]
-      end
-
-      allow(Dir).to receive(:glob).and_return(efivardirs)
+      allow(Bootloader::Systeminfo).to receive(:writable_efivars?).and_return(!removable)
     end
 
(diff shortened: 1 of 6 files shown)


test/grub_install_test.rb
```diff
@@ -12,15 +12,7 @@ def stub_arch(arch)
...
-      efivardirs = if removable
-        []
-      else
-        ["Boot0000-8be4df61-93ca-11d2-aa0d-00e098032b8c",
-         "BootCurrent-8be4df61-93ca-11d2-aa0d-00e098032b8c",
-         "BootOrder-8be4df61-93ca-11d2-aa0d-00e098032b8c"]
-      end
-
-      allow(Dir).to receive(:glob).and_return(efivardirs)
+      allow(Bootloader::Systeminfo).to receive(:writable_efivars?).and_return(!removable)
```
dgdavid, line 15: :thinking: So, if they are _removable_ they are _not writable_?

wfeldt (approved):
lgtm

test/grub_install_test.rb
jreidinger: well, it is kind of other way around. grub2-install has `--removable` flag that means do not write to efibootmanager. So basically do not modify efivars. This way we avoid problem with non writable efivars

dgdavid: Ok, thanks for the explanation :)

PR #636, opened 2021-03-03
fix wrong summary when there are no boot partition/disks as on s390 (…

jreidinger:
…bsc#1181801)

Issue: on s390 it shows "boot locations: , "
Fix: it is caused by fact that on s390 there are no boot partition/disk used in code neither any stage1 devices. So it use include? on empty array that always return true. This is fixed now, so `stage1.boot_partition?` return false if there are no stage1 devices and no boot partition detected.

Testing: manual on s390

### DIFF
--- a/src/lib/bootloader/stage1.rb
+++ b/src/lib/bootloader/stage1.rb
@@ -114,11 +114,17 @@ module Bootloader
     end
 
     def boot_partition?
-      include_real_devs?(boot_partition_names)
+      names = boot_partition_names
+      return false if names.empty?
+
+      include_real_devs?(names)
     end
 
     def mbr?
-      include_real_devs?(boot_disk_names)
+      names = boot_disk_names
+      return false if names.empty?
+
+      include_real_devs?(names)
     end
 
     def logical_boot?
@@ -139,9 +145,11 @@ module Bootloader
     end
 
     def extended_boot_partition?
-      return false if boot_partition_names == extended_boot_partitions_names
+      names = extended_boot_partitions_names
+      return false if names.empty?
+      return false if boot_partition_names == names
 
-      include_real_devs?(extended_boot_partitions_names)
+      include_real_devs?(names)
     end
 
     def custom_devices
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/stage1.rb
```diff
@@ -139,9 +145,11 @@ def extended_boot_partitions_names
     end
 
     def extended_boot_partition?
-      return false if boot_partition_names == extended_boot_partitions_names
+      names = extended_boot_partitions_names
+      return false if names.empty?
```
imobachgs, line 149: np: why not merging this line and the next one?

imobachgs (approved):

ancorgs (approved):
It would be good to not decrease the test coverage. But other than that, LGTM

src/lib/bootloader/stage1.rb
```diff
@@ -114,11 +114,17 @@ def boot_disk_names
...
+
+      include_real_devs?(names)
     end
 
     def mbr?
-      include_real_devs?(boot_disk_names)
+      names = boot_disk_names
+      return false if names.empty?
+
+      include_real_devs?(names)
```
imobachgs, line 127: Why it is not possible to make `include_real_devs` return false if `names` is empty?

src/lib/bootloader/stage1.rb
jreidinger: I like to have non related checks separated for readability

src/lib/bootloader/stage1.rb
jreidinger: Well, it is possible, but inconsistent with ruby include which I would like to avoid:

```
irb(main):001:0> "abc".include?("")
=> true
irb(main):002:0> "abc".include?("a")
=> true
```

imobachgs: Well, I was more thinking about an array:

```ruby
irb(main):001:0> ["a", "b", "c"].include?("")
=> false
```

jreidinger: this is kind of different as it is if array include empty array and not empty string. Here names are also array and not string.

PR #637, opened 2021-03-10
Device map dialog validation

jreidinger:
trello: https://trello.com/c/LOBnWzcU/2338-publicsles-15-sp3-p5-1183218-build-1585-no-grub-related-warnings-where-number-of-devices-exceeds-8
bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1183218

### DIFF
--- a/src/lib/bootloader/device_map_dialog.rb
+++ b/src/lib/bootloader/device_map_dialog.rb
@@ -63,6 +63,7 @@ module Bootloader
           end
           max_dev = Bootloader::DeviceMap::BIOS_LIMIT
           if disks.size > max_dev
+            # TRANSLATORS: an error message where %i is the number of devices.
             Yast::Popup.Error(_("Device map can have at maximum %i devices") % max_dev)
             next
           end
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/device_map_dialog.rb
```diff
@@ -61,6 +61,11 @@ def controller_loop
             Yast::Popup.Error(_("Device map must contain at least one device"))
             next
           end
+          max_dev = Bootloader::DeviceMap::BIOS_LIMIT
+          if disks.size > max_dev
+            Yast::Popup.Error(_("Device map can have at maximum %i devices") % max_dev)
```
dgdavid, line 66: NP: I'd use `format` since I found better having _placeholders_ in translations than %i or %s.

```ruby

format(
  # TRANSLATORS: an error message where %{max_devices} is the number of max devices a device can map.
  _("Device map can have at maximum %{max_devices} devices"),
 max_devices: max_dev
)
```

dgdavid (approved):
LGTM, just a suggestion that you can dismiss.

src/lib/bootloader/device_map_dialog.rb
jreidinger: btw format is just alias to %...and I agree with you when there are more places as it allows translators to move it freely around. But when there is just one, I prefer short notation.

dgdavid (approved):
LGTM

PR #639, opened 2021-04-19
Fix cpu mitigation sp3

jreidinger:
bsc: https://bugzilla.suse.com/show_bug.cgi?id=1184968

issue is that cpu mitigation does not expect to be used outside of grub2 bootloader. So I adapt it.

### DIFF
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -159,11 +159,16 @@ module Bootloader
     end
 
     def init
-      self.value = grub2.cpu_mitigations.value
+      if grub2.respond_to?(:cpu_mitigations)
+        self.value = grub2.cpu_mitigations.value
+      else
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/grub2_widgets.rb
```diff
@@ -159,11 +159,17 @@ def help
     end
 
     def init
-      self.value = grub2.cpu_mitigations.value
+      if grub2.respond_to?(:cpu_mitigations)
+        self.value = grub2.cpu_mitigations.value.to_s
+      else
+        # do not crash when use no bootloader. This widget is also used in security dialog.
+        # (bsc#1184968)
+        disable
```
dgdavid, line 167: Do you mean `:disable`?

jreidinger: no, I want disable widget

dgdavid (approved):

PR #641, opened 2021-05-12
arm can boot on uefi

ggardet:


ggardet:
Cc @dirkmueller

dirkmueller (approved):
LGTM

src/lib/bootloader/systeminfo.rb
```diff
@@ -85,7 +85,7 @@ def efi_used?(bootloader_name)
       #
       # @return [Boolean] true if system can (in principle) boot via UEFI
       def efi_supported?
-        Yast::Arch.x86_64 || Yast::Arch.i386 || Yast::Arch.aarch64
+        Yast::Arch.x86_64 || Yast::Arch.i386 || Yast::Arch.aarch64 || Yast::Arch.arm
```
jreidinger, line 88: the change itself looks reasonable. How it is with rest conditions like secure boot or available shim? and removable installation?

ggardet: There is no shim for 32-bit arm. There is no SecureBoot for 32-bit arm so far.
What do you mean by "removable installation"?

jreidinger: if there is no secure boot, then it should be disabled at https://github.com/yast/yast-bootloader/blob/e2b65506c069f56c4eb6f47b8811eed4bc898c02/src/lib/bootloader/systeminfo.rb#L35

regarding removable installation it is described at https://github.com/yast/yast-bootloader/blob/e2b65506c069f56c4eb6f47b8811eed4bc898c02/src/lib/bootloader/grub_install.rb#L109 and aarch 64 is specific that we always run removable and if efivars are rw then also non removable. In other archs we run it only once.

ggardet: Well, there is no shim but the kernel is signed, so we may use SecureBoot.

`src/lib/bootloader/grub_install.rb` has been fixed.

jreidinger: great, so the last thing missing to give it final blessing. You adapt grub_install, but just at one place. It needs to be adapted also at https://github.com/yast/yast-bootloader/pull/641/files#diff-7267da9b3618b7386967b687f12bd1ea546a56b9da16f073e07320cd2dc3d979R45 so arm workaround works properly also for 32 bit.

ggardet: > great, so the last thing missing to give it final blessing. You adapt grub_install, but just at one place. It needs to be adapted also at https://github.com/yast/yast-bootloader/pull/641/files#diff-7267da9b3618b7386967b687f12bd1ea546a56b9da16f073e07320cd2dc3d979R45 so arm workaround works properly also for 32 bit.

I fail to see where the last update is missing.

jreidinger: @ggardet well, you add condition for arm here to this part of workaround but the same change is also needed at  `src/lib/bootloader/grub_install.rb` line 45 which runs arm also for non-removables, so it is for sure written to both if efivars are rw.

ggardet: Ah yes. Pushed a fixed version.

jreidinger (approved):

PR #642, opened 2021-05-16
Avoid conflicts with filesystem and hicolor-icon-theme (bsc#1184786)

dirkmueller:


package/yast2-bootloader.spec
```diff
@@ -81,17 +81,17 @@ This package contains the YaST2 component for bootloader configuration.
...
 %{yast_scrconfdir}
-%{yast_fillupdir}
 %{yast_schemadir}
 %{yast_libdir}
-%{yast_icondir}
-%doc %{yast_docdir}
-%license COPYING
+%{_fillupdir}/sysconfig.bootloader.*
+%{_datadir}/metainfo/org.opensuse.yast.Bootloader.metainfo.xml
+%{_datadir}/icons/hicolor/*/apps/yast-bootloader*
```
lslezak, line 95: I'd still keep the YaST macros in use, e.g.:

```
%{yast_icondir}/hicolor/*/apps/yast-bootloader*
```

We define the target path in a single place (https://github.com/yast/yast-rake/blob/master/lib/tasks/install.rake#L31), if we (though very unlikely) change it we will not need to touch all packages again, just update `%{yast_icondir}`.

I think keeping one indirection level is OK in this case.

dirkmueller: I moved away from the macro because there is a pull request to change the definition of the `%yast_icondir` macro, so my PR would break. 

I can certainly change it back to yast_icondir if we agree to not change the definition, but whats the value of using a yast_ prefixed macro for something that isn't yast specific?

dirkmueller: @lslezak just to confirm, is your request specifically about going back to `%yast_icondir` or is it a more general remark? e.g. what about `%yast_fillupdir` ? there is no such thing as a yast specific fillup dir, it is hardcoded in the fillup macros.

lslezak (approved):
Approved, thank you for the fix!

PR #644, opened 2021-06-07
Fixes for 'arm can boot on uefi'

ggardet:
This is a follow-up of https://github.com/yast/yast-bootloader/pull/641

### DIFF
--- a/src/modules/BootSupportCheck.rb
+++ b/src/modules/BootSupportCheck.rb
@@ -79,11 +79,11 @@ module Yast
       return true if type == "none"
 
       # grub2 is sooo cool...
-      return true if type == "grub2" && !Arch.aarch64
+      return true if type == "grub2" && !Arch.aarch64 && !Arch.arm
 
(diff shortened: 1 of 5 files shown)


src/modules/BootSupportCheck.rb
```diff
@@ -79,11 +79,11 @@ def correct_loader_type(type)
       return true if type == "none"
 
       # grub2 is sooo cool...
-      return true if type == "grub2" && !Arch.aarch64
+      return true if type == "grub2" && !Arch.aarch64 && !Arch.arm
```
dirkmueller, line 82: you can remove line 81 and 82 when you move line 86 here..

jreidinger: it does not help, as it is support check. So if you select on arm grub2 it should do warning. This line basically says that anything non arm can use non-efi grub2 and line on 86 says that efi is good on arm. So it looks good for me.

jreidinger (approved):

PR #646, opened 2021-06-09
Enable linuxefi/initrdefi on x86 only

ggardet:


Vogtinator:
Technically it can be set to `false` everywhere now because of https://build.opensuse.org/package/view_file/Base:System/grub2/0001-efi-linux-provide-linux-command.patch?expand=1.

src/lib/bootloader/grub2efi.rb
```diff
@@ -59,7 +59,7 @@ def propose
       # for UEFI always remove PMBR flag on disk (bnc#872054)
       self.pmbr_action = :remove
 
-      grub_default.generic_set("GRUB_USE_LINUXEFI", Yast::Arch.aarch64 ? "false" : "true")
+      grub_default.generic_set("GRUB_USE_LINUXEFI", (Yast::Arch.aarch64 || Yast::Arch.arm) ? "false" : "true")
```
Vogtinator, line 62: IMO should be left as `false` everywhere.
If that's too risky, it should check for x86 specifically though as only that has `linuxefi`.

ggardet: So something like:
```
grub_default.generic_set("GRUB_USE_LINUXEFI", (Yast::Arch.x86_64 || Yast::Arch.i386) ? "true" : "false")
```
?

Vogtinator: Yep

ggardet: Done

src/lib/bootloader/grub2efi.rb
```diff
@@ -59,7 +59,7 @@ def propose
       # for UEFI always remove PMBR flag on disk (bnc#872054)
       self.pmbr_action = :remove
 
-      grub_default.generic_set("GRUB_USE_LINUXEFI", Yast::Arch.aarch64 ? "false" : "true")
+      grub_default.generic_set("GRUB_USE_LINUXEFI", (Yast::Arch.x86_64 || Yast::Arch.i386) ? "true" : "false")
```
jreidinger, line 62: well, when you touch it and also be an arm expert, can you please add a comment line above this line of code explaining why it needs to be done only on intel and not on arm? Thanks

ggardet: Ok, I added `# linuxefi/initrdefi are available on x86 only`

jreidinger (approved):

PR #648, opened 2021-06-18
riscv64 enablement

dirkmueller:


src/lib/bootloader/bootloader_factory.rb
```diff
@@ -59,8 +59,9 @@ def supported_names
           system_bl = nil
         end
         ret = system_bl ? [system.name] : [] # use current as first
-        ret << "grub2" unless Yast::Arch.aarch64 # grub2 everywhere except aarch64
-        ret << "grub2-efi" if Yast::Arch.x86_64 || Yast::Arch.aarch64
+        # grub2 everywhere except aarch64
```
jreidinger, line 62: please update also comment here that risc is also efi only.

src/lib/bootloader/grub_install.rb
```diff
@@ -42,7 +42,8 @@ def execute(devices: [], secure_boot: false, trusted_boot: false, update_nvram:
         Yast::Execute.on_target(cmd)
         # workaround for arm on SLE15 SP2 (bsc#1167015)
         # run grub2-install also non-removable if efi is there
-        if (Yast::Arch.aarch64 || Yast::Arch.arm) && Systeminfo.writable_efivars?
+        if (Yast::Arch.aarch64 || Yast::Arch.arm || Yast::Arch.riscv64) &&
```
jreidinger, line 45: so this workaround is also needed on risc? To be honest I am not much happy about it as it is just workaround that is a bit strange.

src/lib/bootloader/bootloader_factory.rb
```diff
@@ -59,8 +59,9 @@ def supported_names
...
+        ret << "grub2" unless Yast::Arch.aarch64 || Yast::Arch.riscv64
+        ret << "grub2-efi" if Yast::Arch.x86_64 || Yast::Arch.aarch64 || Yast::Arch.riscv64
```
jreidinger, line 64: I would use here `SystemInfo.efi_supported?` to simplify it.

src/modules/BootSupportCheck.rb
```diff
@@ -79,11 +79,11 @@ def correct_loader_type(type)
       return true if type == "none"
 
       # grub2 is sooo cool...
-      return true if type == "grub2" && !Arch.aarch64 && !Arch.arm
+      return true if type == "grub2" && !Arch.aarch64 && !Arch.arm && !Arch.riscv64
 
       return true if (Arch.i386 || Arch.x86_64) && type == "grub2-efi" && efi?
 
-      return true if type == "grub2-efi" && (Arch.aarch64 || Arch.arm)
+      return true if type == "grub2-efi" && (Arch.aarch64 || Arch.arm || Arch.riscv64)
```
jreidinger, line 86: As I see duplicities here, I would say that it makes sense to define method efi_only? in SystemInfo class and use it from there as I see it is used in multiple places.

src/lib/bootloader/bootloader_factory.rb
dirkmueller: good point, done!

src/modules/BootSupportCheck.rb
dirkmueller: good point, changed.

src/lib/bootloader/grub_install.rb
dirkmueller: I don't know for sure, the approach taken here is to make it like aarch64, as the booting path is quite similar.

src/lib/bootloader/bootloader_factory.rb
dirkmueller: done

src/lib/bootloader/grub_install.rb
jreidinger: basically that workaround is needed as some arm expect removable installation for efi and does not work without removable, so we install it to both locations. Not sure about risc as I am not familiar with that arch ( well, even with arm I am not so much familiar and depends mainly on experts ).

src/lib/bootloader/bootloader_factory.rb
```diff
@@ -59,8 +59,9 @@ def supported_names
           system_bl = nil
         end
         ret = system_bl ? [system.name] : [] # use current as first
-        ret << "grub2" unless Yast::Arch.aarch64 # grub2 everywhere except aarch64
-        ret << "grub2-efi" if Yast::Arch.x86_64 || Yast::Arch.aarch64
+        # grub2 everywhere except aarch64 or riscv64
+        ret << "grub2" unless Yast::Arch.aarch64 || Yast::Arch.riscv64
```
jreidinger, line 63: I think here it also makes sense to use `Systeminfo.efi_mandatory?` I think missing Arch.arm is just bug

dirkmueller: done

src/lib/bootloader/grub_install.rb
dirkmueller: can we do this in a followup?

jreidinger: yes, fine for me.

jreidinger (approved):

PR #649, opened 2021-07-05
Add os-prober package (SLE-15-SP3)

shundhammer:
_**This is the fix for the SLE-15-SP3 branch. The port to the master branch will follow.**_

## Trello

https://trello.com/c/OVhSxr2e/2493-3-sle15-sp3-1186369-yast2-bootloader-probe-foreign-os-doesnt-detect-add-windows

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1186369

## Problem

No selection for a parallel-boot Windows in the Grub boot menu after a SLES-15-SP3 installation.

## Cause

The `os-prober` package is no longer installed by default, unlike in previous releases.

## Fix

Now explicitly adding the `os-prober` package to the set of packages to install, unless

- the package is not available (not provided by the current product)
- on s/390

In the UI also no longer offering that _Probe Foreign OS_ check box if the package is not available.

## Grub2 for EFI Boot

Does it make sense to do that on EFI boot as well? Not sure.

However, the _Probe Foreign OS_ check box was not only available in the EFI scenario, it was even checked by default (unlike in the non-EFI case), so the old behaviour was retained, i.e. you get the `os-prober` package installed (if it's available and on a non-s/390 architecture) in both the EFI and the non-EFI case.

### DIFF
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -7,6 +7,7 @@ require "bootloader/device_map_dialog"
 require "bootloader/serial_console"
 require "bootloader/cpu_mitigations"
 require "bootloader/systeminfo"
+require "bootloader/os_prober"
 require "cfa/matcher"
 
 Yast.import "BootStorage"
@@ -1103,7 +1104,7 @@ module Bootloader
           TimeoutWidget.new(hiden_menu_widget),
           HSpacing(1),
           VBox(
-            Left(Yast::Arch.s390 ? CWM::Empty.new("os_prober") : OSProberWidget.new),
+            os_prober_widget,
             VSpacing(1),
             Left(hiden_menu_widget)
           ),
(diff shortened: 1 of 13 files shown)


src/lib/bootloader/grub2_widgets.rb
```diff
@@ -1115,5 +1116,15 @@ def contents
...
       )
     end
+
+  private
+
+    def os_prober_widget
+      if OsProber.available? # Checks !Arch.s390 and if package is available
+        Left(OSProberWidget.new)
+      else
+        CWM::Empty.new("os_prober")
```
wfeldt, line 1126: Originally, there was a Left() around it; does it make a difference?

wfeldt (approved):
lgtm

src/lib/bootloader/grub2_widgets.rb
shundhammer: No, empty is empty, no matter if it's left-aligned or not. ;-)

PR #653, opened 2021-09-15
Avoid jsc#PM-2612 reference

afaerber:
The changelog should reference jsc#SLE- ticket number.


afaerber:
Relates to https://github.com/yast/yast-yast2/pull/1196. No further PM- references here.

jreidinger (approved):

PR #654, opened 2021-10-08
Custom device path validation / conversion. 

mchf:
bnc#1092950

## Problem ##

User entered device specification as UUID="<uuid_of_a_device>" and bootloader module crashed with internal error

## Solution ##

1) User's input is validated and an error is reported if the path is invalid
bootloader1

2) strings UUID="<uuid_of_a_device>" or LABEL="<device_label>" are translated to an absolute path (/dev/...)
bootloader2

### DIFF
--- a/src/lib/bootloader/device_path.rb
+++ b/src/lib/bootloader/device_path.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require "yast"
 
 module Bootloader
@@ -29,7 +31,7 @@ module Bootloader
 
     # @return [Boolean] true if the @path exists in the system
     def exists?
-      File.exists?(path)
+      File.exist?(path)
     end
 
     alias_method :valid?, :exists?
--- a/src/lib/bootloader/device_path.rb
+++ b/src/lib/bootloader/device_path.rb
@@ -19,10 +19,10 @@ module Bootloader
     def initialize(dev)
       @path =  if dev_by_uuid?(dev)
         # if defined by uuid, convert it
-        File.realpath(dev.sub(/UUID="([-a-zA-Z0-9]*)"/, '/dev/disk/by-uuid/\1'))
+        dev.sub(/UUID="([-a-zA-Z0-9]*)"/, '/dev/disk/by-uuid/\1')
       elsif dev_by_label?(dev)
         # as well for label
-        File.realpath(dev.sub(/LABEL="(.*)"/, '/dev/disk/by-label/\1'))
+        dev.sub(/LABEL="(.*)"/, '/dev/disk/by-label/\1')
       else
         # add it exactly (but whitespaces) as specified by the user
         dev.strip
--- a/src/lib/bootloader/device_path.rb
+++ b/src/lib/bootloader/device_path.rb
@@ -1,0 +1,47 @@
+require "yast"
+
+module Bootloader
+  # Class for device path
+  #
+  # @example device path can be defined explicitly
+  #   DevicePath.new("/devs/sda")
+  # @example definition by UUID is translated to device path
+  #   dev = DevicePath.new("UUID=\"0000-00-00\"")
+  #   dev.path -> "/dev/disk/by-uuid/0000-00-00"
+  class DevicePath
+    attr_reader :path
+
+    # Performs initialization
+    #
+    # @param dev [<String>] either a path like /dev/sda or special string for uuid or label
+    def initialize(dev)
+      @path =  if dev_by_uuid?(dev)
+        # if defined by uuid, convert it
+        File.realpath(dev.gsub(/UUID="([-a-zA-Z0-9]*)"/, '/dev/disk/by-uuid/\1'))
+      elsif dev_by_label?(dev)
+        # as well for label
+        File.realpath(dev.gsub(/LABEL="(.*)"/, '/dev/disk/by-label/\1'))
+      else
+        # add it exactly (but whitespaces) as specified by the user
+        dev.strip
+      end
+    end
+
+    # @return [Boolean] true if the @path exists in the system
+    def exists?
+      File.exists?(path)
+    end
+
+    alias :valid? :exists?
+
+  private
+
+    def dev_by_uuid?(dev)
+      dev =~ /UUID=".+"/
+    end
+
+    def dev_by_label?(dev)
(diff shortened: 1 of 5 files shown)


src/lib/bootloader/device_path.rb
```diff
@@ -0,0 +1,47 @@
...
+  class DevicePath
+    attr_reader :path
+
+    # Performs initialization
+    #
+    # @param dev [<String>] either a path like /dev/sda or special string for uuid or label
+    def initialize(dev)
+      @path =  if dev_by_uuid?(dev)
+        # if defined by uuid, convert it
+        File.realpath(dev.gsub(/UUID="([-a-zA-Z0-9]*)"/, '/dev/disk/by-uuid/\1'))
```
jreidinger, line 20: ugh, bootloader should not read real path on system. It is storage responsibility. that gsub is fine ( even I would use simple `sub`, but calling realpath is problematic. E.g. consider autoyast config mode.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -911,6 +918,28 @@ def init_custom_devices(custom_devices)
...
+      # particular byte sequence is valid or not
+      return true if Mode.config
+
+      devs_list.split(",").all? do |d|
+        dev_path = DevicePath.new(d)
+
+        if Yast::Mode.installation
+          # uuids are generated later by mkfs, so not known in time of installation
+          # so whatever can be true
+          dev_path.uuid? || dev_path.label?
```
jreidinger, line 936: ugh, what? This means that only uuid or label is valid in custom devices, which is definitively correct as you can specify even kernel names or by-id or by-path here.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -880,16 +883,20 @@ def store
...
       if Yast::UI.QueryWidget(:custom, :Value)
         devs = Yast::UI.QueryWidget(:custom_list, :Value)
         if devs.strip.empty?
-          Yast::Report.Error(_("Custom boot device have to be specied if checked"))
+          Yast::Report.Error(_("Custom boot device has to be specified if checked"))
+          Yast::UI.SetFocus(Id(:custom_list))
+          return false
+        end
+        if !valid_custom_devices?(devs)
+          Yast::Report.Error(_("One of custom partitions is invalid"))
```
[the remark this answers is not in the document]
kobliha: In fact a little _[Choose...]_ button that would open up a new dialog for selecting a partition from the list does not seem to be that bad idea. In all places where possible. Using the same functionality everywhere.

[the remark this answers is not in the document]
jreidinger: it should be possible. Maybe storage already provide such dialog? I think in general it is not just about partitions, it can be any disk, partition, LVM volume, raid device, etc.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -880,16 +883,32 @@ def store
...
+        Yast::UI.SetFocus(Id(:custom_list))
+        return false
+      end
+
+      invalid_devs = invalid_custom_devices(devs)
+      if !invalid_devs.empty?
+        ret = Yast::Popup.ContinueCancel(
+          _(
+            "These custom devices can be invalid: #{invalid_devs.join(",")}" \
+            "Please check if exist and spelled correctly" \
```
jreidinger, line 906:
```suggestion
            "Please check if exist and spelled correctly." \
```

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -880,16 +883,32 @@ def store
...
+        Yast::Report.Error(_("Custom boot device has to be specified if checked"))
+        Yast::UI.SetFocus(Id(:custom_list))
+        return false
+      end
+
+      invalid_devs = invalid_custom_devices(devs)
+      if !invalid_devs.empty?
+        ret = Yast::Popup.ContinueCancel(
+          _(
+            "These custom devices can be invalid: #{invalid_devs.join(",")}" \
```
jreidinger, line 905:
```suggestion
            "These custom devices can be invalid: #{invalid_devs.join(", ")}." \

File: src/lib/bootloader/grub2_widgets.rb [hunk 6]
```diff
@@ -911,6 +930,30 @@ def init_custom_devices(custom_devices)
...
+      # AY profile can be generated for whatever system so we cannot decite if
+      # particular byte sequence is valid or not
+      return [] if Mode.config
+
+      devs_list.split(",").reject do |d|
+        dev_path = DevicePath.new(d)
+
+        if Yast::Mode.installation
+          # uuids are generated later by mkfs, so not known in time of installation
+          # so whatever can be true
```
jreidinger, line 949: is the statement correct also for labels? or does label have to be specified in advance and known?

jreidinger (approved):

src/lib/bootloader/udev_mapping.rb
```diff
@@ -58,8 +58,12 @@ def staging
     end
 
     def udev_to_kernel(dev)
+      dev_path = DevicePath.new(dev)
+
       # in mode config if not found, then return itself
       return dev if Yast::Mode.config
+      # in installation do not care of uuids - not know at this time
+      return dev if Yast::Mode.installation && dev_path.uuid?
```
jreidinger, line 66: this sadly does not work as expected, as udev mapping is called also in inst_finish, where partitioning and all uuid is known and needs to be handled properly.

mchf: well, it worked as expected, bud i didn't know that it is effective even in the finish script ;-)

jreidinger (approved):

src/lib/bootloader/device_path.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    def dev_by_uuid?(dev)
+      dev =~ /UUID=".+"/
+    end
+
+    def dev_by_label?(dev)
+      dev =~ /LABEL=".+"/
+    end
+
+    def devicegraph
```
ancorgs, line 60: It should be ok to always use staging here. Isn't it?

DimStar77:
Results in https://bugzilla.opensuse.org/show_bug.cgi?id=1193016

PR #657, opened 2021-12-08
rely on yast-storage-ng to detect UEFI boot support status (bsc#937067)

wfeldt:
## Problem

- https://trello.com/c/ZWMQnKom
- https://bugzilla.suse.com/show_bug.cgi?id=937067

EFI detection code is all over the place. Consolidate in one spot.

## Solution

Use latest `Y2Storage::Arch` class to determine UEFI boot support.

## See also

- https://github.com/yast/yast-storage-ng/pull/1256

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Dec 10 12:05:57 UTC 2021 - Steffen Winterfeldt <snwint@suse.com>
+
+- rely on yast-storage-ng to detect UEFI boot support status (bsc#937067)
+- 4.4.10
+
+-------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


jreidinger:
Well, I kind of expect it will need some test adaptation, but otherwise code looks good. I am just curious why it is in Y2Storage::Arch when EFI is not much about storage and more about boot. So why not in Yast::Arch?

wfeldt:
The EFI code is actually in libstorage-ng in Storage::Arch and it cannot be moved out as libstorage-ng itself also uses it.
Y2Storage::Arch is the wrapper around that class. So that's how it ended up there.

That there's also Yast::Arch (with even slightly different names for the architectures) historically is indeed not so nice, but I wouldn't address that atm.

The code is atm a first try and I'll need some more testing.

jreidinger (approved):

PR #658, opened 2021-12-13
Honors the user decision when aborting after reporting probing issues

dgdavid:
TODO:

* Update requried yast-storage-ng version.

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Dec 15 08:31:38 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Honors the user decision when aborting after reporting
+  probing issues (bsc#1193749).
+- 4.4.10
+
+-------------------------------------------------------------------
 Thu Nov 25 08:29:34 UTC 2021 - Michal Filka <mfilka@suse.com>
 
 - bnc#1193016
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/config_dialog.rb
```diff
@@ -26,6 +26,8 @@ def initialize(initial_tab: :boot_code)
 
     def run
       guarded_run
+    rescue Yast::AbortException
+      :abort
```
jreidinger, line 30: If you want to introduce new general Abort Exception then I find having it handled in each client a bit error prone and over kill. Why not handle it in WFM.Call itself like other general exceptions? - https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/wfm.rb#L358 Just do not mix AbortException from ruby and Yast::AbortException...maybe better name like UserAbort or UIAbortException would be better to distinguish,

dgdavid: I have checked it today. About allowing `WFM.Call` to handle the situation, you're completely right @jreidinger. In fact, looks like `Yast::AbortException` was introduced for that (or similar) purpose. See https://github.com/yast/yast-ruby-bindings/pull/215.

Thanks a lot!

dgdavid:
This change is not necessary after new solution proposed at https://github.com/yast/yast-storage-ng/pull/1265 (see https://github.com/yast/yast-storage-ng/pull/1265/commits/b971949acf0b72907144d82c4e21ee370fa709a2).

PR #660, opened 2022-01-06
New rubocop

jreidinger:
draft for review of changes done in new rubocop that works with ruby 3.1


src/lib/bootloader/grub2base.rb
```diff
@@ -414,36 +414,39 @@ def propose_encrypted
...
-        else
-          "<a href=\"enable_secure_boot\">(" + _("enable") + ")</a>"
-        end
+      link = if secure_boot
+        "<a href=\"disable_secure_boot\">(#{_("disable")})</a>"
+      else
+        "<a href=\"enable_secure_boot\">(#{_("enable")})</a>"
+      end
+
+      "#{_("Secure Boot:")} #{secure_boot ? _("enabled") : _("disabled")} #{link}"
```
lslezak, line 423: NP: this pattern is repeated several times in the code, what about defining some helper like

```ruby
def status_string(status)
  status ? _("enabled") : _("disabled")
end

# and then use in
"#{_("Secure Boot:")} #{status_string(secure_boot)} #{link}"

```

test/grub2pwd_test.rb
```diff
@@ -5,6 +5,23 @@
 
 require "bootloader/grub2pwd"
 
+ENCRYPTED_PASSWORD = "grub.pbkdf2.sha512.10000.774E325959D6D7BCFB7384A0245674D83D0D540A89C02FEA81E35489F8DE7ADFD93988190AD9857A0FFF363825DDF97C8F4E658D8CC49FC4A22C053B08AB3EFE.6FB19FF26FD03D85C40A33D8BA7C04E72EDE3DD5D7080C177553A4FED370F71C579AF0B15B3B93ECECEA355469A4B6D0560BFB53ED35DDA0B80F5363BFBD54E4"
```
lslezak, line 8: Too long line?

lslezak (commented):
Looks OK to me

src/lib/bootloader/grub2base.rb
jreidinger: yeap, also notice it.

test/grub2pwd_test.rb
jreidinger: strange that it does not complain. Not sure why

src/lib/bootloader/sysconfig.rb
```diff
@@ -35,14 +35,14 @@ def initialize(bootloader: nil, secure_boot: false, trusted_boot: false, update_
     end
 
     def self.from_system
-      bootloader = Yast::SCR.Read(AGENT_PATH + "LOADER_TYPE")
+      bootloader = Yast::SCR.Read("#{AGENT_PATH}LOADER_TYPE")
```
jreidinger, line 38: and here is wrong auto correct. Because AGENT_PATH is not string, but Yast::Path, then result is different.

teclator:
It LGTM but maybe @dgdavid found something I overlooked :stuck_out_tongue_winking_eye:

src/modules/Bootloader.rb
```diff
@@ -418,7 +418,6 @@ def modify_kernel_params(*args)
 
           case value
           when false then next # already done
-          when true then param.add_parameter(key, value)
           when Array
             value.each { |val| param.add_parameter(key, val) }
```
dgdavid, line 422: Is is `true.each { ...` now? If so, you know,  `undefined method each' for true:TrueClass)`

jreidinger: no, true is not Array...it goes to else section which is `param.add_parameter(key, value)`

src/lib/bootloader/grub2base.rb
```diff
@@ -446,7 +446,15 @@ def update_nvram_summary
...
-      "#{_("Update NVRAM:")} #{update_nvram ? _("enabled") : _("disabled")} #{link}"
+      "#{_("Update NVRAM:")} #{status_string(update_nvram)} #{link}"
+    end
+
+    def status_string(status)
+      if status
+        _("enabled")
+      else
+        _("disabled")
+      end
```
dgdavid, line 457: NP: I'd use ternary operator here.

dgdavid (approved):
LGTM

src/modules/Bootloader.rb
dgdavid: Right! I overlooked else and mixed branches at the time of review :)

PR #661, opened 2022-01-10
Support powerpc secure boot (bsc#1192764 jsc#SLE-18271).

hramrach:
This adds powerpc secure boot detection support to systeminfo and grub
install.

Currently we have secure boot support indicated by ibm,secure-boot file
presence but in many configurations this file is present but secure boot
cannot work.

We should propose secure boot on powerpc when and only when content of
this file is non-zero because then booting without secure boot will
fail.

On the other hand we should enable the user to manually turn on secure 
boot when ibm,secure-boot is present or even in all cases.

In the future it is expected that the firmware grows something like 
efivarfs allowing the OS to pass additional parameters to the secure 
bootloader which should make the secure boot reliable.

See also https://github.com/openSUSE/perl-bootloader/pull/137

### DIFF
--- a/src/lib/bootloader/grub_install.rb
+++ b/src/lib/bootloader/grub_install.rb
@@ -101,6 +101,14 @@ module Bootloader
       cmd << "--removable" if removable_efi?
       cmd << "--no-nvram" if !update_nvram
 
+      if target == "powerpc-ieee1275" then
+        if secure_boot then
+          cmd << "--suse-force-signed"
+        else
+          cmd << "--suse-inhibit-signed"
+        end
+      end
+
       cmd
     end
 
--- a/src/lib/bootloader/systeminfo.rb
+++ b/src/lib/bootloader/systeminfo.rb
@@ -18,14 +18,15 @@ module Bootloader
       #
       # @return [Boolean] true if secure boot is currently active
       def secure_boot_active?
-        (efi_supported? || s390_secure_boot_supported?) && Sysconfig.from_system.secure_boot
+        (efi_supported? || s390_secure_boot_supported? || ppc_secure_boot_active?) &&
+          Sysconfig.from_system.secure_boot
       end
 
       # Check if secure boot is in principle supported.
       #
       # @return [Boolean] true if secure boot is (in principle) supported on this system
       # def secure_boot_supported?
-      #  efi_supported? || s390_secure_boot_supported?
+      #  efi_supported? || s390_secure_boot_supported? || ppc_secure_boot_supported?
       # end
 
       # Check if secure boot is configurable with a bootloader.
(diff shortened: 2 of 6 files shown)


hramrach:
rubocop test seems generally broken but unit tests for ppc are missing hence the coverage decrease

hramrach:
There are some failures in secure_boot_available? - needs investigation.

hramrach:
I added tests for the new powerpc code and they pass so I will call this finished.

As mentioned this is not complete because on powerpc we need to distinguish 'supported' and 'available' secure boot state which is not supported by the upper layers.

We need to propose secure boot when 'supported' but allow the user to turn it on when 'available' - or however are these states named in the end.

hramrach:
Looks like this should just work. The proposal code uses secure_boot_active? and the enablement code secure_boot_available?

src/lib/bootloader/grub_install.rb
```diff
@@ -85,6 +86,9 @@ def report_failure(exception)
     # creates basic command for grub2 install without specifying any stage1
     # locations
     def basic_cmd(secure_boot, trusted_boot, update_nvram)
+      if secure_boot && (target == "powerpc-ieee1275") then
+        return ["/usr/bin/dd", "status=none", "if=/usr/share/grub2/powerpc-ieee1275/grub.elf"]
```
jreidinger, line 90: so it means that grub2-install call is skipped completely? As you replace it with dd. And as I write in bug report I kind of expect that grub2 itself will do this dd, not all tools using grub

src/lib/bootloader/systeminfo.rb
```diff
@@ -137,6 +138,50 @@ def s390_secure_boot_active?
         false
       end
 
+      # Return secure boot status on ppc
+      #
+      # nil - no support
+      # 0   - disabled
+      # 1   - enabled in audit-only mode
+      # 2+  - enabled in enforcing mode
+      def ppc_secure_boot
```
jreidinger, line 147: all ppc specific methods should have something like
```
      def ppc_secure_boot
          return nil unless Yast::Arch.ppc
```

so it does not affect non ppc code.

hramrach: It reads a ppc-specific file so this is pretty safe.

jreidinger: yeap, but it do some unnecessary file read and also makes code more obvious, that it is ppc specific. as ibm,secure-boot can be e.g. added to s390x which can do some side effects.

hramrach: s390 has /sys/firmware/ipl/secure

jreidinger: it is current state, but show know what brings future? So I prefer more exact code that survive potential changes on other architectures.

jreidinger:
change looks good, so please just merge master branch to your branch, bump changelog and version and I can merge it.
Thanks

jreidinger (approved):

jreidinger:
merged, thanks for patch

PR #665, opened 2022-01-28
use product specific boot timeout proposal (jsc#SLE-22667)

jreidinger:
jira:  https://jira.suse.com/browse/SLE-22667

related: https://github.com/yast/yast-yast2/pull/1240 and https://github.com/yast/yast-installation-control/pull/122

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Jan 28 22:37:00 UTC 2022 - Josef Reidinger <jreidinger@suse.com>
+
+- Allow product/system role to select boot timeout (jsc#SLE-22667)
+- 4.4.16
+
+-------------------------------------------------------------------
 Tue Jan 25 13:11:58 UTC 2022 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Use Package module instead of PackageSystem (bsc#1194886).
--- a/src/lib/bootloader/grub2base.rb
+++ b/src/lib/bootloader/grub2base.rb
@@ -361,7 +361,7 @@ module Bootloader
     def propose_timeout
       return if grub_default.timeout
 
-      grub_default.timeout = "8"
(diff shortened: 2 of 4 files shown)


jreidinger:
tested manually and works as expected

src/lib/bootloader/grub2base.rb
```diff
@@ -361,7 +361,7 @@ def propose_terminal
     def propose_timeout
       return if grub_default.timeout
 
-      grub_default.timeout = "8"
+      grub_default.timeout = Yast::ProductFeatures.GetIntegerFeature("globals", "boot_timeout").to_s
```
teclator, line 364: So, what is default if not defined?

jreidinger: It is always defined. See - https://github.com/yast/yast-yast2/pull/1240/files#diff-e3d91269dec42511487bf9598e632fe2da743b05435e0cf3c77da23e520c0f77R55

package/yast2-bootloader.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Jan 28 22:37:00 UTC 2022 - Josef Reidinger <jreidinger@suse.com>
+
+- Allow product/system role to select boot timeout (jsc#SLE-22667)
+- 4.4.16
```
teclator, line 5: Just miss the version bump in the spec file.

jreidinger: good catch

src/lib/bootloader/grub2base.rb
teclator: .. so, maybe the dependency should be updated requiring that specific version? https://github.com/yast/yast-bootloader/blob/0e1d4068e86aaf04a602a62ee30a4a72395aa1c2/package/yast2-bootloader.spec#L46

teclator: https://github.com/yast/yast-bootloader/blob/0e1d4068e86aaf04a602a62ee30a4a72395aa1c2/package/yast2-bootloader.spec#L29

teclator (approved):
LGTM

PR #666, opened 2022-04-05
Resume boot parameter fixes for SLE15 SP3

jreidinger:
## Problem

Issue is that resume boot parameter is cloned, but partitioning is not cloned, so often device is invalid. This is especially true for default device by uuid which should be always unique.

- https://bugzilla.suse.com/show_bug.cgi?id=1187690
- https://bugzilla.suse.com/show_bug.cgi?id=1197192

## Solution

Solution is on two fronts.

1. Do not export resume parameter so it is not part of cloned profile
2. During import if non existing  device is in resume, remove resume part to keep old profiles working even with new dracut.

## Testing

- [x] *Added a new unit test*
- [ ]  *Tested manually*

## Screenshots

### DIFF
--- a/src/lib/bootloader/autoyast_converter.rb
+++ b/src/lib/bootloader/autoyast_converter.rb
@@ -101,6 +101,12 @@ module Bootloader
           val = data.global.public_send(key)
           next unless val
 
+          # import resume only if device exists (bsc#1187690)
+          resume = val[/[\s^]resume=(\S+)/, 1]
+          if resume && !Yast::BootStorage.staging.find_by_any_name(resume)
+            val = val.gsub(/[\s^]resume=#{Regexp.escape(resume)}/, "")
+          end
+
           default.public_send(method).replace(val)
         end
 
--- a/src/lib/bootloader/autoyast_converter.rb
+++ b/src/lib/bootloader/autoyast_converter.rb
@@ -277,7 +277,11 @@ module Bootloader
 
         DEFAULT_KERNEL_PARAMS_MAPPING.each do |key, method|
           val = default.public_send(method)
-          res[key] = val.serialize unless val.empty?
+          result = val.serialize
+          # do not export resume parameter as it depends on storage which is
+          # not cloned by default and if missing it is proposed (bsc#1187690)
+          result.gsub!(/[\s^]resume=\S+/, "")
+          res[key] = result unless result.empty?
         end
 
         DEFAULT_STRING_MAPPING.each do |key, method|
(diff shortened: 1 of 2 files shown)


src/lib/bootloader/autoyast_converter.rb
```diff
@@ -101,6 +101,13 @@ def import_default(data, default)
           val = data.global.public_send(key)
           next unless val
 
+          # import resume only if device exists (bsc#1187690)
+          resume = val[/(?:\s|\A)resume=(\S+)/, 1]
+          if resume && !Yast::BootStorage.staging.find_by_any_name(resume)
+            log.warn "remove resume parameter for non existing #{resume}"
+            val = val.gsub(/(?:\s|\A)resume=#{Regexp.escape(resume)}/, "")
```
dgdavid, line 108: Question: is not valid to simply use the same as above here? I.e., `val.gsub(/(?:\s|\A)resume=\S+/, "")`

jreidinger: yes, it can be used. I just want to protect here if there are multiple resumes, to not remove valid ones ( even if multiple resumes does not make much sense, but due to some perioud of buggy behavior some autoyast profiles, can contain it )

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -101,6 +101,13 @@ def import_default(data, default)
           val = data.global.public_send(key)
           next unless val
 
+          # import resume only if device exists (bsc#1187690)
+          resume = val[/(?:\s|\A)resume=(\S+)/, 1]
+          if resume && !Yast::BootStorage.staging.find_by_any_name(resume)
+            log.warn "remove resume parameter for non existing #{resume}"
```
ancorgs, line 107: I know it's only a warning in the logs, but it could be phrased to explain the problem better

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -277,7 +284,11 @@ def export_default(res, default)
 
         DEFAULT_KERNEL_PARAMS_MAPPING.each do |key, method|
           val = default.public_send(method)
-          res[key] = val.serialize unless val.empty?
+          result = val.serialize
+          # do not export resume parameter as it depends on storage which is
+          # not cloned by default and if missing it is proposed (bsc#1187690)
```
ancorgs, line 289: I would break the sentence in two to make it more understandable.

```
# Do not export the 'resume' parameter as it depends on storage, which is not
# cloned by default. Anyways, 'resume' will be proposed if it's missing (bsc#1187690).
```

jreidinger: thanks for note. In the end I found that it is cloned, but it is hard to apply as uuid, id neither path is cloned. Only label one.

ancorgs (approved):
LGTM

PR #669, opened 2022-07-25
Adaptations to work inside a container

ancorgs:
## Problem

yast2-bootloader does not work inside a container out of the box (see [bsc#1199840](https://bugzilla.suse.com/show_bug.cgi?id=1199840)). There are several things that need to be fixed:

1) storage-ng probing must work. See https://github.com/yast/yast-in-container/pull/10
2) The command `grub2-mkpasswd-pbkdf2` is executed locally (ie. in the container) instead of the target system due to [historical reasons](https://bugzilla.suse.com/show_bug.cgi?id=900039) that do not longer apply.
3) Problems installing the needed packages.

## Solution

As mentioned, the first problem is addressed in a separate pull request at another repository.

This fixes the second problem by making sure `grub2-mkpasswd-pbkdf2` is executed in the target system.

Problem 3 remains to be fixed (and likely will be addressed by an upcoming pull request at yast-yast2).

## Testing

- Tested manually that containerized execution gets fixed by this.
- Tested manually that installation is not broken by this because now `grub2-mkpasswd-pbkdf2` is executed during the proposal phase, when `Execute.locally` and `Execute.on_target` are still equivalent.

## Notes

During manual testing I found the grub2 password is logged plain at the YaST logs. Bug to be reported.

### DIFF
--- a/src/lib/bootloader/grub2pwd.rb
+++ b/src/lib/bootloader/grub2pwd.rb
@@ -131,7 +131,7 @@ module Bootloader
     end
 
     def encrypt(password)
-      result = Yast::Execute.locally("/usr/bin/grub2-mkpasswd-pbkdf2",
+      result = Yast::Execute.on_target("/usr/bin/grub2-mkpasswd-pbkdf2",
         env:    { "LANG" => "C" },
         stdin:  "#{password}\n#{password}\n",
         stdout: :capture)
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Jul 25 15:26:05 UTC 2022 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Execute the command grub2-mkpasswd-pbkdf2 in the target system
+  so the module can run in a minimal container (bsc#1199840).
+- 4.5.2
+
+-------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


src/lib/bootloader/grub2pwd.rb
```diff
@@ -131,7 +131,7 @@ def disable
     end
 
     def encrypt(password)
-      result = Yast::Execute.locally("/usr/bin/grub2-mkpasswd-pbkdf2",
+      result = Yast::Execute.on_target("/usr/bin/grub2-mkpasswd-pbkdf2",
```
jreidinger, line 134: well, another option I see is to have that mkpasswd installed locally as it does not need to be on target system if it is decided to split it from grub2. We need on target system mainly tools used to create config and install stage1 due to new kernel and new grub2 modules - so grub2-mkconfig and grub2-install.

In general I am not against change if it is properly tested during upgrade from old SLE...do we still support migration from SLE11?

ancorgs: I decided to go this way because is simpler and I don't think grub2-mkpasswd-pbkdf2 will be split from the grub2 package in the foreseeable future.

That tool is included in the grub2 package even for SLE-11. So if you are upgrading such a system the tool is already in the target.

dgdavid (approved):
Bearing in mind your comment at https://github.com/yast/yast-bootloader/pull/669/files#r929612013, it LGTM.

Please, do not forget to report the bug mentioned in the description

> During manual testing I found the grub2 password is logged plain at the YaST logs. Bug to be reported.

PR #671, opened 2022-09-08
Transactional

jreidinger:
## Problem

Bootloader failing on transactional system as /boot is read only.

## Solution

Call specific transactional-update calls and avoid direct modification of that content.

## Testing

- *Tested manually*

## Screenshots

### DIFF
--- a/src/lib/bootloader/bootloader_base.rb
+++ b/src/lib/bootloader/bootloader_base.rb
@@ -36,7 +36,7 @@ module Bootloader
     end
 
     # writes configuration to target disk
-    def write; end
+    def write(etc_only: false); end
 
     # reads configuration from target disk
     def read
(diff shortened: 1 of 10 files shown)


lslezak:
Looks good so far...

src/lib/bootloader/bootloader_base.rb
```diff
@@ -36,7 +36,7 @@ def prepare
     end
 
     # writes configuration to target disk
-    def write; end
+    def write(etc_only: false); end
```
mvidner, line 39:
```suggestion
    # @param etc_only [Boolean] true on transactional systems
    #   because /boot is read-only there
    def write(etc_only: false); end
```

jreidinger: thanks

imobachgs (approved):

PR #672, opened 2022-09-21
Fixed hidden timeout proposal

mchf:
## Problem

We should propose "hidden timeout" by default. [bsc#1203418](https://bugzilla.suse.com/show_bug.cgi?id=1203418)

### DIFF
--- a/src/lib/bootloader/grub2base.rb
+++ b/src/lib/bootloader/grub2base.rb
@@ -216,20 +216,6 @@ module Bootloader
 
   private
 
-    def propose_grub_default
-      if grub_default.kernel_params.empty?
-        kernel_line = Yast::BootArch.DefaultKernelParams(propose_resume)
-        grub_default.kernel_params.replace(kernel_line)
-      end
-      grub_default.gfxmode ||= "auto"
-      grub_default.recovery_entry.disable unless grub_default.recovery_entry.defined?
-      grub_default.distributor ||= ""
-      grub_default.default = "saved"
-      # always propose true as grub2 itself detect if btrfs used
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/grub2base.rb
```diff
@@ -227,6 +216,20 @@ def serial_console?
 
   private
 
+    def propose_grub_default
```
jreidinger, line 219: I would move the code closer to other propose private methods at https://github.com/yast/yast-bootloader/pull/672/files#diff-5797ed1429ad469f007496587dfc38e58cb54f82bc44c3bf99d81b9bef6b58beR338

src/lib/bootloader/grub2base.rb
```diff
@@ -227,6 +216,20 @@ def serial_console?
...
+        kernel_line = Yast::BootArch.DefaultKernelParams(propose_resume)
+        grub_default.kernel_params.replace(kernel_line)
+      end
+      grub_default.gfxmode ||= "auto"
+      grub_default.recovery_entry.disable unless grub_default.recovery_entry.defined?
+      grub_default.distributor ||= ""
+      grub_default.default = "saved"
+      # always propose true as grub2 itself detect if btrfs used
+      grub_default.generic_set("SUSE_BTRFS_SNAPSHOT_BOOTING", "true")
+      grub_default.hidden_timeout = "0"
```
jreidinger, line 230: as timeout and hidden timeout are quite related I would add this code to propose_timeout method https://github.com/yast/yast-bootloader/pull/672/files#diff-5797ed1429ad469f007496587dfc38e58cb54f82bc44c3bf99d81b9bef6b58beR366

jreidinger (approved):

PR #674, opened 2022-10-05
Pwd leak sp1

jreidinger:
## Problem

Grub2 password is logged to yast logs in plain text when doing encryption.

- https://bugzilla.suse.com/show_bug.cgi?id=1201962

Note: SLE15-SP1 is the first affected code stream as previously is used WFM.Execute which does not logs. And it is still in some LTSS supported.

## Solution

Do not log stdin for command to generate encrypted password.

## Testing

- *Tested manually* as unit tests for writting to logs is quite fragile.

PR #676, opened 2022-10-06
Merge pwd leak sp3

jreidinger:
port of https://github.com/yast/yast-bootloader/pull/675 and https://github.com/yast/yast-bootloader/pull/674

a bit tricky as previously merge from SP2 was not done, so it needs a bit more conflict solving


mvidner:
OK, I've checked that the other change on SLE-15-SP2 is slready present in SP3 as
```
Wed Mar  3 09:30:42 UTC 2021 - Josef Reidinger <jreidinger@suse.com>

- Respect if efivars is mounted read only (bsc#1174111,
  bsc#1182749)
- 4.3.23
```

mvidner (approved):

PR #677, opened 2022-10-06
Merge pwd leak sp4

jreidinger:
merge of https://github.com/yast/yast-bootloader/pull/676 https://github.com/yast/yast-bootloader/pull/675 https://github.com/yast/yast-bootloader/pull/674

PR #678, opened 2022-10-07
Merge pwd leak master

jreidinger:
merge of https://github.com/yast/yast-bootloader/pull/677  https://github.com/yast/yast-bootloader/pull/676 https://github.com/yast/yast-bootloader/pull/675 , description -> https://github.com/yast/yast-bootloader/pull/674

+ use the new Yast::Reducer class from https://github.com/yast/yast-yast2/pull/1277

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Oct  5 21:35:19 UTC 2022 - Josef Reidinger <jreidinger@suse.com>
+
+- prevent leak of grub2 password to logs(bsc#1201962)
+- 4.5.7
+
+-------------------------------------------------------------------
 Wed Sep 21 20:42:31 UTC 2022 - Josef Reidinger <jreidinger@suse.com>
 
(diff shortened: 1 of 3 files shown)


mvidner:
@jreidinger I have tested successfully. `-I path` doesn't work because `require "yast"` overrides `-I` by prepending /usr/share/YaST2/lib

```
cd bootloader
irb -ryast

$LOAD_PATH.unshift "../yast2/library/system/src/lib"
$LOAD_PATH.unshift "src/lib"
require "bootloader/grub2pwd.rb"
pwd = Bootloader::GRUB2Pwd.new
pwd.password = "test"
```

mvidner (changes requested):
Well, the code is fine, but the single new commit is just a 'merge', which is a lie.

We have merged the code using NoStdinRecorder and then silently replaced it with ReducedRecorder. The replacement should ideally be a separate commit, but if you don't want to redo it, amend the merge commit message:

> Note: also replaced the inlined NoStdinRecorder with a more general ReducedRecorder (in yast2.rpm)

mvidner (approved):

PR #679, opened 2023-01-12
Ppc secure

jreidinger:
## Problem

After selecting secure boot on ppc64 when it is not enabled in HMC (hardware management console), it is shown after rerun as disabled, which is confusing.

## Solution

After discussion with IBM we show properly checked checkbox and add note to enable secure boot also in HMC. 

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

( note screenshots due to some difficulties with our ppc comes from mocked code on x86_64, so it other widgets and values can look a different than on ppc64 )

unchecked checkbox 
ppc_sec_boot_unchecked
after checking it
ppc_sec_boot_checked

### DIFF
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -346,7 +346,7 @@ module Bootloader
       textdomain "bootloader"
 
       self.handle_all_events = true
-      @secure_boot_widget = SecureBootWidget.new
+      @secure_boot_widget = SecureBootWidget.new(notify: true)
 
       super
     end
@@ -359,6 +359,7 @@ module Bootloader
     end
 
     def init
+      @secure_boot_widget.init # ensure it is init before this widget so we can read value
       handle_label
     end
 
@@ -366,6 +367,7 @@ module Bootloader
       return unless events["ID"] ==  @secure_boot_widget.widget_id
 
       handle_label
+      nil
     end
 
   private
@@ -387,10 +389,15 @@ module Bootloader
   class SecureBootWidget < CWM::CheckBox
     include Grub2Widget
 
-    def initialize
+    def initialize(notify: false)
       textdomain "bootloader"
+      @notify = notify
 
-      super
+      super()
+    end
+
+    def opt
+      @notify ? [:notify] : []
     end
 
     def label
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -339,6 +339,50 @@ module Bootloader
     end
   end
 
+  # Specific secure boot widget for ppc as we need to handle also
+  # secure boot state in hardware. For details see https://bugzilla.suse.com/show_bug.cgi?id=1206295
+  class SecureBootWidgetPPC < CWM::CustomWidget
+    def initialize
+      textdomain "bootloader"
+
+      self.handle_all_events = true
+      @secure_boot_widget = SecureBootWidget.new
+
+      super
+    end
+
+    def contents
+      VBox(
+        @secure_boot_widget,
+        ReplacePoint(Id(:ppc_sec_label), Empty())
+      )
+    end
+
+    def init
+      handle_label
+    end
+
+    def handle(events)
+      return unless events["ID"] ==  @secure_boot_widget.widget_id
+
+      handle_label
+    end
+
+  private
+
+    def handle_label
+      # no work if already enabled in HMC
+      return if Systeminfo.ppc_secure_boot_active?
+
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/grub2_widgets.rb
```diff
@@ -339,14 +339,65 @@ def store
     end
   end
 
+  # Specific secure boot widget for ppc as we need to handle also
+  # secure boot state in hardware. For details see https://bugzilla.suse.com/show_bug.cgi?id=1206295
```
mvidner, line 343: Thanks for the Bugzilla link but it should be in the PR description upfront :)

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -339,14 +339,65 @@ def store
...
+    end
+
+  private
+
+    def handle_label
+      # no work if already enabled in HMC
+      return if Systeminfo.ppc_secure_boot_active?
+
+      if @secure_boot_widget.value
+        label_widget = Label(_("Enable also in HMC!"))
```
mvidner, line 380: What who?!

Without context, I as the user wouldn't be sure who does the HMC enabling: YaST or myself?

Make it more explicit: "Currently disabled in HMC, remember to enable it also there"
Or, if we can't see the HMC value: "Make sure to enable it also in HMC"

mvidner (changes requested):

src/lib/bootloader/grub2_widgets.rb
mvidner: About the "HMC" term: 

- UAAB!!
- the related code https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/systeminfo.rb#L159 should use same naming, so mention "HMC" in a comment there


(Unexplained Abbreviations Are Bad)

hramrach: s/HMC/firmware/

HMC is one way to manage firmware settings on the pSeries platform

We also support PowerVC, and powernv platform.

src/lib/bootloader/systeminfo.rb
```diff
@@ -18,7 +18,7 @@ class << self
       #
       # @return [Boolean] true if secure boot is currently active
       def secure_boot_active?
-        (efi_supported? || s390_secure_boot_supported? || ppc_secure_boot_active?) &&
+        (efi_supported? || s390_secure_boot_supported? || ppc_secure_boot_supported?) &&
           Sysconfig.from_system.secure_boot
```
hramrach, line 22: I suppose this is the actual fix although the commit message is very misleading

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -339,6 +339,50 @@ def store
     end
   end
 
+  # Specific secure boot widget for ppc as we need to handle also
+  # secure boot state in hardware. For details see https://bugzilla.suse.com/show_bug.cgi?id=1206295
```
hramrach, line 343: So this is not testing secure boot widget on ppc, this is adding special secure boot widget for ppc. Again, very misleading commit message.

And what for?

What is the difference between the spacial ppc widget and the secure boot widget for other platforms?

It should be the same.

The only special behavior intended was that secure boot should be disabled by default on ppc when platform supports it but does not enable it.

src/lib/bootloader/systeminfo.rb
jreidinger: well, fix contain two parts first is here that state of checkbox reflect OS settings only. The second is special widget which I will explain at second part.

src/lib/bootloader/grub2_widgets.rb
jreidinger: Well, the specific functionality of ppc widget is that it say in case that secure boot is enabled and not enabled in firmware, to display label with reminder to enabled secure boot also in hardware. See screenshots in pull request description. We do not do it for other architectures as e.g. on x86_64 you do not need to enable it, you can only enforce it in firmware, so it behaves differently ( at least what I understand so far ).

hramrach: How is that different? You enable/disable SB enforcement in firmware that supports it.

hramrach: Isn't that the same on x86? It either checks and enforces the signature, or does neither.

The IBM firmware actually has all options (do not care, check but not enforce, check and enforce) but the 'check but not enforce' option is disregarded, it's mostly for debugging or other special use.

jreidinger:
obsoleted by https://github.com/yast/yast-bootloader/pull/680

PR #680, opened 2023-02-08
Secure ppc2

jreidinger:
## Problem

secure boot on ppc works differently than on other architectures. It checks if hardware has enforced signed kernel and only in that case propose to use it and also only in that case is checkbox checked even if OS is set to install signed kernel.

bug(private): https://bugzilla.suse.com/show_bug.cgi?id=1206295
replaces https://github.com/yast/yast-bootloader/pull/679

## Solution

Make it consistent with other architectures and reflect only OS state and propose secure boot when it is supported.

## Testing

- *Adapted unit test*

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Feb  8 15:30:25 UTC 2023 - Josef Reidinger <jreidinger@suse.com>
+
+- make secure boot for ppc64 consistent with how secure boot works
+  on other architectures (bsc#1206295)
+- 4.5.8
+
+-------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


hramrach:
This likely enables secure boot on some systems that don't really need it (possible but not activated) but it's better than inability to turn on secure boot.

Also people who already have an installed system should inherit the existing configuration that sets secure boot to false.

jreidinger:
@hramrach yeap, it will run same as on other architectures, where you get it even if you do not enforce it in BIOS. if we found problematic case, we can cover it as we do for s390 where it is proposed only for scsi https://github.com/yast/yast-bootloader/pull/680/files#diff-db6b868f6a66626401a3bbd83ec5412746a89f38cdd2c0eaa35648e8005b280eR135

imobachgs (approved):

PR #684, opened 2023-04-25
Fix migration sp5

jreidinger:
## Problem

User selection was ignored during migration.

- https://bugzilla.suse.com/show_bug.cgi?id=1210811

## Solution

Drop legacy code for SLE11 migration and reset proposal only if it is forced.

## Testing

- manually


src/lib/bootloader/proposal_client.rb
```diff
@@ -226,7 +226,8 @@ def propose_for_update(force_reset)
       # old one is grub2, but mismatch of EFI and non-EFI (bsc#1081355)
       elsif old_bootloader =~ /grub2/ && old_bootloader != current_bl.name
         raise MismatchBootloader.new(old_bootloader, current_bl.name)
-      elsif !current_bl.proposed? || force_reset
+      elsif force_reset
+        log.warn "forced proposel during migration. Will change existing config"
```
shundhammer, line 230: Typo: s/proposel/proposal/

shundhammer (approved):
LGTM

PR #686, opened 2023-05-24
Systemd boot feature

schubi2:
## Problem

Offer a GPLv3-free boot path for ALP (https://jira.suse.com/browse/PED-1906)

## Solution

Implementing systemd-boot support in yast-bootloader.
Systemd-boot will only supported by products which have explicit enabled it in the product definition file.
Additional, systemd-boot is currently a work in process. The user will be warned that there is currently
no official support (see screenshot).
Maybe there are still pending discussion how and when this PR can be used ;-)

## Testing

- Added a new unit test
- Tested manually while normal installation, AY installation and in the installed system.

## AY Docu

https://github.com/SUSE/doc-sle/pull/1509

## Screenshots
systemdboot

## Needed PRs
The following PR is needed for it too:
https://github.com/yast/yast-installation/pull/1093


kobliha:
I somehow thought this would not be merged to master. It's an experimental feature and Benjamin Brunner explicitly says in https://jira.suse.com/browse/PED-1906 that systemd-boot is not supported.

schubi2:
> I somehow thought this would not be merged to master. It's an experimental feature and Benjamin Brunner explicitly says in https://jira.suse.com/browse/PED-1906 that systemd-boot is not supported.

Yes,  that's why I have said to wait with the PR merge. There is still an ongoing discussion with Bennies team how to go on here....

SUPPORTED_SCENARIOS.md
```diff
@@ -8,6 +8,8 @@ The goal of this document is to have a single source of information  which scena
 * grub2-efi
   * only for UEFI boot
   * only with GPT (see [bug](https://bugzilla.novell.com/show_bug.cgi?id=889733#c8))
+* systemd-boot
+  * only for UEFI boot
```
jreidinger, line 12: Does it mean it should work for every UEFI on all architectures like ppc or aarch?

jreidinger:
note: new svg has Sysconfig background transparent and it is inconsistent. It looks like mistake.

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -33,19 +33,26 @@ def import(data)
...
-        import_stage1(data, bootloader)
-        import_default(data, bootloader.grub_default)
-        import_device_map(data, bootloader)
-        import_password(data, bootloader)
-        # always nil pmbr as autoyast does not support it yet,
-        # so use nil to always use proposed value (bsc#1081967)
-        bootloader.pmbr_action = nil
-        cpu_mitigations = data.global.cpu_mitigations
-        bootloader.cpu_mitigations = CpuMitigations.from_string(cpu_mitigations) if cpu_mitigations
+        if ["grub2", "grub2-efi"].include?(bootloader.name)
```
jreidinger, line 36: prefered syntax should be 
```ruby
case bootloader.name
when "grub2", "grub2-efi"
  ...
when "systemd-boot"
  ...
```

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -33,19 +33,26 @@ def import(data)
...
+          # so use nil to always use proposed value (bsc#1081967)
+          bootloader.pmbr_action = nil
+          cpu_mitigations = data.global.cpu_mitigations
+          if cpu_mitigations
+            bootloader.cpu_mitigations = CpuMitigations.from_string(cpu_mitigations)
+          end
+        end
+        if bootloader.name == "systemd-boot"
+          bootloader.menue_timeout = data.global.timeout
+          bootloader.secure_boot = data.global.secure_boot
```
jreidinger, line 53: this should be definitively documented in autoyast documentation in SUSE/sle-doc to avoid surprise of users/customers.

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -61,12 +68,19 @@ def export(config)
...
 
         res["global"] = {}
-        global = res["global"]
-        export_grub2(global, config) if config.name == "grub2"
-        export_grub2efi(global, config) if config.name == "grub2-efi"
-        export_default(global, config.grub_default)
-        export_password(global, config.password)
-        res["global"]["cpu_mitigations"] = config.cpu_mitigations.value.to_s
+
+        if ["grub2", "grub2-efi"].include?(config.name)
```
jreidinger, line 72: same here for using `case` instead of set of ifs.

src/lib/bootloader/systemdboot.rb
```diff
@@ -0,0 +1,304 @@
...
+
+    # @!attribute secure_boot
+    #   @return [Boolean] current secure boot setting
+    attr_accessor :secure_boot
+
+    def initialize
+      super
+
+      textdomain "bootloader"
+      self.menue_timeout = Yast::ProductFeatures.GetIntegerFeature("globals", "boot_timeout").to_i
```
jreidinger, line 33: I think that this part should be in propose method and not in initialize.

src/lib/bootloader/systemdboot.rb
```diff
@@ -0,0 +1,304 @@
...
+      install_bootloader
+      create_menue_entries
+      write_menue_timeout
+
+      true
+    end
+
+    def propose
+      super
+      log.info("Propose settings...")
```
jreidinger, line 74: I am missing here proposal for secure boot

src/lib/bootloader/systemdboot.rb
```diff
@@ -0,0 +1,304 @@
...
+      self.secure_boot = other.secure_boot unless other.secure_boot.nil?
+    end
+
+    def cpu_mitigations
+      Bootloader::CpuMitigations.new(:manual)
+    end
+
+    def cpu_mitigations=(value)
+      log.info "ignoring set of cpu_mitigations (#{value}) in systemd bootloader."
+    end
```
jreidinger, line 50: as here it is ignored. Do we even need this methods? It should behave same as for "none" bootloader and basically do not show cpu mitigation when it is not written.

src/lib/bootloader/systemdboot.rb
```diff
@@ -0,0 +1,304 @@
...
+    def name
+      "systemd-boot"
+    end
+
+    def packages
+      res = super
+
+      case Yast::Arch.architecture
+      when "x86_64", "aarch64"
+        res << "shim" << "mokutil" if secure_boot
```
jreidinger, line 120: so shim can work with systemd boot? and what about risc64? it also support EFI see https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/systeminfo.rb#L98

jreidinger: And also does it mean that systemd does not need any software for its installation? where lives bootctl?

src/lib/bootloader/systemdboot.rb
```diff
@@ -0,0 +1,304 @@
...
+                     "Cannot create systemd-boot menue entry:\n" \
+                     "Command `%{command}`.\n" \
+                     "Error output: %{stderr}"
+                   ), command: e.commands.inspect, stderr: e.stderr)
+          )
+            break
+          end
+        end
+      end
+      File.delete(cmdline_file) if Yast::Stage.initial
```
jreidinger, line 178: why delete is needed? does not it prevent installation of params for kernel update? In code comment should be rationale

src/lib/bootloader/systemdboot.rb
```diff
@@ -0,0 +1,304 @@
...
+    end
+
+    def read_menue_timeout
+      filename = File.join(Yast::Installation.destdir, LOADERCONF)
+      return unless File.file?(filename)
+
+      f = File.open(filename, "r")
+      f.each_line do |line|
+        line.lstrip!
+        next if line.start_with?("#")
```
jreidinger, line 188: this looks like quite fragile manual parsing. Maybe augeas can be used to parse that conf?

src/lib/bootloader/systemdboot.rb
```diff
@@ -0,0 +1,304 @@
...
+      del_files.each do |f|
+        filename = File.join(Yast::Installation.destdir, f)
+        File.delete(filename) if File.exist?(filename)
+      end
+    end
+
+    def secure_boot_available
+      ret = false
+      begin
+        ret = Yast::Execute.on_target(MOKUTIL, "--sb-state", allowed_exitstatus: 1) == 0
```
jreidinger, line 236: I am not sure, but cannot be Systeminfo module used for checking secure boot availability?

src/lib/bootloader/systemdboot.rb
```diff
@@ -0,0 +1,304 @@
...
+      )
+        return
+      end
+      remove_secure_boot_settings
+    end
+
+    def set_secure_boot
+      log.info("Enabling secure boot options")
+      src = File.join(Yast::Installation.destdir, "/boot/efi/EFI/systemd/systemd-bootx64.efi")
+      dest = File.join(Yast::Installation.destdir, "/boot/efi/EFI/systemd/grub.efi")
```
jreidinger, line 266: hmm, this will need some comments **why** mv is needed and why we cannot use default systemd file name.

src/lib/bootloader/systemdboot_widgets.rb
```diff
@@ -0,0 +1,169 @@
...
+            HSpacing(1),
+            TimeoutWidget.new,
+            HSpacing(1)
+          ),
+          VStretch()
+        )
+      end
+    end
+  end
+end
```
jreidinger, line 169: good as POC, but I think if more settings like kernel params or cpu mitigation appears we should somehow abstract widgets to be able to reuse ones we have for grub2 as they looks almost identical.

src/modules/BootArch.rb
```diff
@@ -63,7 +63,7 @@ def DefaultKernelParams(resume)
         ret << " quiet"
         ret
       elsif Arch.s390
-        termparm = if ENV["TERM"] == "linux"
+        termparm = if ENV.fetch("TERM", nil) == "linux"
```
jreidinger, line 66: ugh why it is needed?

src/modules/BootStorage.rb
```diff
@@ -240,17 +249,24 @@ def boot_partitions
       stage1_partitions_for(boot_filesystem)
     end
 
+    # shortcut to get stage1 partitions for /
+    def root_partitions
```
jreidinger, line 253: do you test it with some more fancy partitioning like having root on MD raid over LVM? As I am not sure if systemd boot wants RAID partition or physical one where it lives.

test/systemdboot_test.rb
```diff
@@ -0,0 +1,86 @@
...
+      subject.secure_boot = false
+      subject.menue_timeout = 10
+
+      subject.merge(other)
+
+      expect(subject.secure_boot).to eq true
+      expect(subject.menue_timeout).to eq 12
+    end
+  end
+end
```
jreidinger, line 86: :+1: for tests

SUPPORTED_SCENARIOS.md
bbrunner: PPC does not support UEFI, therefore it should be out of the picture.
For aarch64, while in theory it should be possible, AFAICS it's currently completely untested in the PoC for openSUSE MicroOS (https://build.opensuse.org/project/show/devel:microos:systemd-boot).

IMHO, we should focus on x86_64 UEFI only, but this needs some additional input from the Micro OS side.

src/lib/bootloader/systemdboot.rb
bbrunner: bootctl is currently located in udev (sub-package of systemd).
Just as a little warning, in Factory we are planning to move everything systemd-boot related into a new sub-package, for better separation soonish.

src/lib/bootloader/systemdboot.rb
bbrunner: Unfortunately, shim currently only checks hardcoded for grub.efi.
See also https://en.opensuse.org/Systemd-boot#Secure_Boot

src/lib/bootloader/bootloader_factory.rb
```diff
@@ -50,6 +53,10 @@ def clear_cache
       def supported_names
         if Yast::Mode.config
           # default means bootloader use what it think is the best
+          if Yast::ProductFeatures.GetBooleanFeature("globals", "enable_systemd_boot")
+            return BootloaderFactory::SUPPORTED_BOOTLOADERS + [SYSTEMDBOOT, DEFAULT_KEYWORD]
+          end
```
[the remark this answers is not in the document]
bbrunner: I'm not completely sure, if I fully understand the comment.
Nevertheless, to be on the save side. ;) We have to make sure that this enablement doesn't accidentally sneak into SLE and is only available in the openSUSE MicroOS installer.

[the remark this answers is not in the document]
jreidinger: well, the code is about `Mode.config` which is specific mode when you want to create autoyast profile using UI and it kind of broken now as we are still not sure how much hardware/product on which it runs affect resulting autoyast profile. So if we allow to create TW profile on SLE autoyast and vice versa. So here it is mainly note that this change will strict to create autoyast profile using config mode only to products where it is supported. @imobachgs maybe you can clarify how it is expected that config mode works across products.

src/lib/bootloader/systemdboot.rb
jreidinger: ok, so when it happens, we will need to add such package here, so we are sure it is installed.

src/lib/bootloader/systemdboot.rb
jreidinger: ok, so @schubi2 please document rationale for this step in comment in code, otherwise in few years we will forget it and maybe when shim start supporting it, we will be able to remove it.
Thanks

lnussel:
meanwhile I've created https://github.com/lnussel/sdbootutil to handle both installation of systemd-boot itself as well as installing kernels into the ESP. So the logic for that doesn't have to be implemented by yast.

jreidinger:
@schubi2 as discussed on slack, please add also limitation to x86_64 only.

jreidinger:
@lnussel thanks. Less logic in yast means easier reuse if customers wants to use it elsewhere or also to implement it in perl-bootloader where it is needed for kernel upgrades.

schubi2:
Thanks for all the input ! I will try to adapt them.....

src/lib/bootloader/autoyast_converter.rb
schubi2: It is already documented in this PR
https://github.com/SUSE/doc-sle/pull/1509/files#diff-b3b3b12399b1094b281a3537befddb0f672fe039dc546a93f714b333c1d49ab8R60

SUPPORTED_SCENARIOS.md
schubi2: I have restricted it

src/lib/bootloader/bootloader_factory.rb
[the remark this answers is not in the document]
schubi2: Well, that could be a problem. But the same problem we have here with grub and grub-efi and no one has really blamed it in the past. So I hope this problem is more "theoretically" :-)

src/lib/bootloader/grub2base.rb
```diff
@@ -143,6 +143,10 @@ def write(etc_only: false)
       @password.write
       return if etc_only
 
+      # reset eventl. already installed systemd bootloader
+      systemd_boot = BootloaderFactory.bootloader_by_name("systemd-boot")
+      systemd_boot&.delete
```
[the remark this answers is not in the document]
schubi2: Yes, good point !

src/modules/BootArch.rb
schubi2: rubocop has requested it :-)

src/lib/bootloader/systemdboot.rb
schubi2: I have restricted it to x86_64 only

src/lib/bootloader/systemdboot.rb
schubi2: Yes, I have moved it.

src/lib/bootloader/systemdboot.rb
schubi2: I have done it like "none" bootloader. So, I have removed it.

src/lib/bootloader/systemdboot.rb
schubi2: Well, because it will be created here too. But maybe you are right and I should not delete it......

schubi2: I have checked it. /etc/kernel/cmdline will be used by the administrator in an installed system only. As we have created it in this function we should delete it again.

src/lib/bootloader/systemdboot.rb
schubi2: Yes, good argument. Do you have any example which I can use for it ?

src/lib/bootloader/systemdboot.rb
schubi2: That has been my first thought too :-) . But in that state Systeminfo has not returned the correct value. So I have used mokutil. If it is worth I can check it again.....

schubi2:
> @schubi2 as discussed on slack, please add also limitation to x86_64 only.

I have added this restriction.

src/lib/bootloader/systemdboot.rb
schubi2: I have described it.

src/modules/BootStorage.rb
schubi2: Quite difficult to test all scenarios, isn't it ? Josef, how have you tested it ?

src/lib/bootloader/systemdboot.rb
schubi2: I have used augeas now. Looks saver :-)

src/lib/bootloader/systemdboot.rb
bbrunner: JFYI, since sr#1096633, bootctl (and everything else related to sd-boot) has been moved to a new subpackage systemd-boot now

schubi2:
I have refactored the PR in order to use sdbootutil

lnussel:
So any blockers left that would prevent merging this?

src/modules/BootStorage.rb
jreidinger: well, usually it is tested by openQA and if I touch code that affects it then I give it a manual run with combination ( so usually LVM and software raid together ) and test if it works as expected.

src/lib/bootloader/finish_client.rb
```diff
@@ -65,21 +65,25 @@ def write
...
-      # fate #303395: Use kexec to avoid booting between first and second stage
-      # copy vmlinuz, initrd and flush kernel option into /var/lib/YaST2
-      if Yast::Linuxrc.InstallInf("kexec_reboot") == "1"
-        kexec = ::Bootloader::Kexec.new
-        kexec.prepare_environment
-      else
-        log.info "Installation started with kexec_reboot set 0"
-      end
+      if !BootloaderFactory.current.is_a?(SystemdBoot)
+        # not for SystemdBoot bootloader
```
jreidinger, line 69: can you also add reason why kexec and dracut regeneration is not for systemdboot ?

src/lib/bootloader/systemdboot.rb
```diff
@@ -0,0 +1,190 @@
...
+
+    def packages
+      res = super
+      res << "sdbootutil" << "systemd-boot"
+
+      case Yast::Arch.architecture
+      when "x86_64"
+        res << "shim" if secure_boot
+      else
+        log.warn "Unknown architecture #{Yast::Arch.architecture} for EFI"
```
jreidinger, line 117: I would say here  is typo as it is not unknown architecture for EFI but systemdboot

schubi2: Yes, I have fixed it.

src/lib/bootloader/finish_client.rb
schubi2: A good point. I have added it.

aplanas:
> currently the biggest question is if/how to hide this feature for some products 

I suggest to use the skelcd-control file, maybe some flag in `<globals>` like `<enable_sdboot config:type="boolean">true</enable_sdboot>`. By default is `false`, and should not be presented as an option

SUPPORTED_SCENARIOS.md
```diff
@@ -77,5 +79,9 @@ This option requires packages based on the architecture of the system:
 * arm architecture requires: <b>grub2-arm-efi</b>.
 * aarch64 architecture requires: <b>grub2-arm64-efi</b>.
 
+## systemd-boot
+If you're running a multiboot EFI system, systemd-boot can provide easier boot management and may even reduce your boot times.
+Systemd-boot will be supported on x86_64 EFI architecture only.
```
aplanas, line 84: Just to note that should work on aarch64 too if EFI is used via uboot (like grub2 today)

src/lib/bootloader/autoyast_converter.rb
```diff
@@ -33,19 +33,28 @@ def import(data)
...
+          import_stage1(data, bootloader)
+          import_default(data, bootloader.grub_default)
+          import_device_map(data, bootloader)
+          import_password(data, bootloader)
+          # always nil pmbr as autoyast does not support it yet,
+          # so use nil to always use proposed value (bsc#1081967)
+          bootloader.pmbr_action = nil
+          cpu_mitigations = data.global.cpu_mitigations
+          if cpu_mitigations
+            bootloader.cpu_mitigations = CpuMitigations.from_string(cpu_mitigations)
```
aplanas, line 49: should this be also shared with the systemd-boot option?

schubi2:
> > currently the biggest question is if/how to hide this feature for some products
> 
> ~I suggest to use the skelcd-control file, maybe some flag in `<globals>` like `<enable_sdboot config:type="boolean">true</enable_sdboot>`. By default is `false`, and should not be presented as an option~
> 
> Edit: I see that it is already using `enable_systemd_boot`. Should this resolve the question?

Yes :-)

SUPPORTED_SCENARIOS.md
schubi2: Currently we are supporting x86_64 only. But good point....

src/lib/bootloader/autoyast_converter.rb
schubi2: Hm, what do you mean with this ?

aplanas: Those kernel parameters should be shared in the case of Grub2 and also in the case of sd-boot. Shouldn't?

jreidinger: well, what I see so far is that sd-boot support in this PR as it is now does not allow kernel params modification and cpu_mitigations depends on it.

jreidinger (approved):

src/lib/bootloader/autoyast_converter.rb
lnussel: for systemd-boot /etc/kernel/cmdline is used when creating new boot entries

lnussel:
fwiw the submission declined due to staging fail on i586. Not sure whether due to this change or a previous one.

lnussel:
test suite fail also in devel project

PR #687, opened 2023-06-15
Persist zfcp.allow_lun_scan kernel option

joseivanlopez:
## Problem

The zFCP kernel module enables the *allow_lun_scan* option by default (since SLE 12). But that option can be disabled by means of the *zfcp.allow_lun_scan=0* kernel parameter. If *auto_lun_scan* is configured, then the kernel option has to be persisted in the target system.

See https://github.com/openSUSE/agama/pull/626.

## Solution

Adds *zfcp.allow_lun_scan* kernel option to the list of options to persist on s390 arch.

## Testing

* Added a new unit test


teclator (approved):
LGTM

jreidinger (approved):

PR #689, opened 2023-08-31
fixed test for checking -supported_names-

schubi2:
## Problem

Testsuite was broken for architectures other than x86_64 while checking the supported bootloaders.


jreidinger:
LGTM just please bump version and changes

jreidinger (approved):
thanks

PR #683, opened 2023-02-14
[PoC] support 32 bit UEFI firmware on x86_64/i386 architecture (bsc#1208003, jsc#PED-2569)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1208003
- https://jira.suse.com/browse/PED-2569

PR #690, opened 2023-09-07
Backport sp6

jreidinger:
## Problem

SLE 15 SP6 branch is based on SLE15 SP5 maintenance branch. So fixes done in master can be sometimes useful also for SP6.

## Solution

- https://github.com/yast/yast-bootloader/pull/683 backported
- https://github.com/yast/yast-bootloader/pull/687 as it is agama only is not backported to avoid side-effect in SLE15
- https://github.com/yast/yast-bootloader/pull/685 is already in SP5


mvidner (approved):
LGTM

also CI is green again, yay

PR #691, opened 2024-01-10
Fix nil factory

jreidinger:
## Problem

insts-sys does not have sysconfig for bootloader. The code first try to get bootloader name from system and then translate it to object. Sadly that name is not checked if it is defined and it result in unnecessary error in yast logs.

## Solution

Check if bootloader name is defined and if not, then just return nil to activate condition at https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/bootloader_factory.rb#L42 to use proposed bootloader instead.

## Testing

- *Tested manually* ( official SP6 medium and yupdate from this branch and then on proposal screen check yast logs that it does not contain `/tmp/var/log/YaST2/y2log-2.gz:2023-12-16 06:53:34 <3> install(4251) [Ruby] bootloader/bootloader_factory.rb(bootloader_by_name):84 Factory receive nil name`

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jan 10 23:16:39 UTC 2024 - Josef Reidinger <jreidinger@suse.com>
+
+- Do not try finding undefined bootloader name to avoid error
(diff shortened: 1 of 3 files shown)


shundhammer (approved):
LGTM

PR #693, opened 2024-01-25
Allow cio_ignore kernel argument on s390 (bsc#1210525)

teclator:
## Problem

The cio_ignore kernel argument is ignored on s390 when the parameters are read and later it is not written properly (see https://github.com/yast/yast-installation/pull/1104 for more details).

- https://bugzilla.suse.com/show_bug.cgi?id=1210525

## Solution

Whitelist the cio_ignore parameter on s390.

## Testing

- *Tested manually*

### DIFF
--- a/package/yast2-bootloader.changes
+++ b/package/yast2-bootloader.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jan 25 11:18:12 UTC 2024 - Knut Anderssen <kanderssen@suse.com>
+
+- Add cio_ignore to allowed kernel options on s390 (bsc#1210525)
+- 5.0.4
+
+-------------------------------------------------------------------
 Wed Jan 10 23:16:39 UTC 2024 - Josef Reidinger <jreidinger@suse.com>
 
 - Do not try finding undefined bootloader name to avoid error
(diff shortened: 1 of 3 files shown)


package/yast2-bootloader.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Jan 25 11:18:12 UTC 2024 - Knut Anderssen <kanderssen@suse.com>
+
+- Add cio_ignore to allowed kernel options on s390 (bsc#1210525)
```
jreidinger, line 4: I would not say `allowed` and more likely mention that it will be kept from installation environment.

src/modules/BootArch.rb
```diff
@@ -36,13 +36,14 @@ def main
...
-    # from installation to running kernel on s390 (bsc#1086665)
+    # from installation to running kernel on s390 (bsc#1086665, bsc#1210525)
     S390_WHITELIST = [
       /net\.ifnames=\S*/,
       /fips=\S*/,
       /mitigations=\S*/,
       /rd\.zdev=\S*/,
-      /zfcp\.allow_lun_scan=\S*/
+      /zfcp\.allow_lun_scan=\S*/,
+      /cio_ignore=\S*/
```
jreidinger, line 46: I would add here commend why we keep cio_ignore ( so point to bug there and also mention that we prefer one from installation and otherwise we propose it ).

package/yast2-bootloader.changes
teclator: Fine with me just reused the previous entry for rd.zdev.

jreidinger (approved):

PR #694, opened 2024-02-20
New flag in the product description file:

schubi2:
Due
https://github.com/yast/yast-installation/pull/1109
the user can set systemd-boot und luks2 for the proposal.
The communication will be done via the product description (global/prefered_bootloader = "systemd-boot")
This has the advantage that the product can also define which bootloader should be installed.


jreidinger:
I will do review on friday due to time off

jreidinger (approved):
looks good to me. It is still product specific, so it should not affect other products if they do not want systemd boot. And splitting enablement and preferred looks good to me.

PR #695, opened 2024-02-27
Feature: Do not timeout boot menue in systemd-boot. (bsc#1216366)

schubi2:
Bildschirmfoto_2024-02-27_11-31-51

### DIFF
--- a/src/lib/bootloader/systemdboot_widgets.rb
+++ b/src/lib/bootloader/systemdboot_widgets.rb
@@ -25,7 +25,7 @@ module Bootloader
 
         super()
 
-        @minimum = -1
+        @minimum = 0
         @maximum = 600
         @default = 10
       end
@@ -35,7 +35,7 @@ module Bootloader
       def contents
         CheckBoxFrame(
           Id(:cont_boot),
-          _("Finishing systemd-boot menu after a certain amount of time"),
+          _("Automatically boot the default entry after a timeout"),
           false,
           HBox(
             IntField(Id(:seconds), _("&Timeout in Seconds"), @minimum, @maximum,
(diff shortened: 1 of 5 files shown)


Vogtinator:
I'm not sure whether it makes sense to translate to `-1` internally. That way there are three states:

a) Checkbox checked, timeout >= 0
b) Checkbox unchecked, timeout  ignored
c) Checkbox checked, timeout -1

c shouldn't be possible IMO

src/lib/bootloader/systemdboot_widgets.rb
```diff
@@ -27,25 +27,52 @@ def initialize
...
 
-      attr_reader :minimum, :maximum
+      attr_reader :minimum, :maximum, :default
 
-      def label
-        _("&Timeout in Seconds")
+      def contents
+        CheckBoxFrame(
+          Id(:cont_boot),
+          _("Finishing systemd-boot menu after a certain amount of time"),
```
Vogtinator, line 38:
```suggestion
          _("Automatically boot the default entry after a timeout"),
```

src/lib/bootloader/systemdboot_widgets.rb
```diff
@@ -27,25 +27,52 @@ def initialize
...
           "specifies the time the boot loader will wait until the default kernel is loaded.</p>\n")
       end
 
+      def validate
+        if Yast::UI.QueryWidget(Id(:cont_boot), :Value) &&
+            Yast::UI.QueryWidget(Id(:seconds), :Value) == -1
+          systemdboot.menue_timeout = Yast::ProductFeatures.GetIntegerFeature("globals",
+            "boot_timeout").to_i
+          systemdboot.menue_timeout = @default if systemdboot.menue_timeout <= 0
+          Yast::UI.ChangeWidget(Id(:seconds), :Value, systemdboot.menue_timeout.to_i)
```
jreidinger, line 60: this should not be in validate. Validate returns true/false if value is valid or not. It should not  do auto corrections. I would simple change in init minimum to `0`.

jreidinger: and if needed to use default, do it in `init` method

schubi2:
> I'm not sure whether it makes sense to translate to `-1` internally. That way there are three states:
> 
> a) Checkbox checked, timeout >= 0 b) Checkbox unchecked, timeout ignored c) Checkbox checked, timeout -1
> 
> c shouldn't be possible IMO

The user does not see -1 and cannot change the value to -1 anymore now.

src/lib/bootloader/systemdboot_widgets.rb
schubi2: Looks better. Thanks!

src/lib/bootloader/systemdboot_widgets.rb
schubi2: Yes, there has been a better place than validate :-)

src/lib/bootloader/systemdboot_widgets.rb
```diff
@@ -17,35 +17,63 @@ def systemdboot
...
+        Yast::UI.ChangeWidget(Id(:cont_boot), :Value, systemdboot.menue_timeout >= 0)
+        systemdboot.menue_timeout = default_value if systemdboot.menue_timeout < 0
+        Yast::UI.ChangeWidget(Id(:seconds), :Value, systemdboot.menue_timeout)
       end
 
       def store
-        systemdboot.menue_timeout = value.to_s
+        if Yast::UI.QueryWidget(Id(:cont_boot), :Value)
+          systemdboot.menue_timeout = Yast::UI.QueryWidget(Id(:seconds), :Value)
+          systemdboot.menue_timeout = default_value if systemdboot.menue_timeout == -1
```
Vogtinator, line 63: Can this ever be -1 here, given that it's read from the widget here?

schubi2: Yes it could if Yast::UI.QueryWidget(Id(:cont_boot), :Value) is set. menue_timeout is set to the default value in orde r not to see -1 for the user anymore. I have tested it. 
You have requested it ;-)

Vogtinator: `Yast::UI.QueryWidget(Id(:seconds), :Value)` is controlled by yast though and has a minimum of 0

Vogtinator: Not resolved yet, but apparently I can't reopen this thread

jreidinger (approved):

Vogtinator:
Pls don't ignore https://github.com/yast/yast-bootloader/pull/695#discussion_r1506134457

schubi2:
> Pls don't ignore [#695 (comment)](https://github.com/yast/yast-bootloader/pull/695#discussion_r1506134457)

Yes, you are right. I have removed this useless command and it is still working.

PR #696, opened 2024-03-07
improve warning for secure boot on s390 to reflect changes done on ha…

jreidinger:
…rdware side (bsc#1219989)

## Problem

Hardware start supporting more disk types for secure boot on s390, so warning is no longer accurate.

## Solution

Try to improve warning text and also help text to be more accurate.

### DIFF
--- a/src/lib/bootloader/grub2_widgets.rb
+++ b/src/lib/bootloader/grub2_widgets.rb
@@ -359,8 +359,8 @@ module Bootloader
           "<p><b>Secure Boot Support</b> if checked enables Secure Boot support.<br>" \
           "This does not turn on secure booting. " \
           "It only switches to the new secure-boot enabled boot data format. " \
-          "Note that this new format works only on z15 or later and only for zFCP disks. " \
-          "The system does not boot if these requirements are not met.</p>"
+          "Note that this new format works only on z15 or later and disk type supports depends on hardware. " \
+          "For details consult requirements at https://www.ibm.com/docs/en/linux-on-systems?topic=introduction-requirements</p>"
         )
       else
         _(
@@ -388,9 +388,11 @@ module Bootloader
 
       Yast::Popup.ContinueCancel(
         _(
-          "The new secure-boot enabled boot data format works only on z15 " \
-          "and later and only for zFCP disks.\n\n" \
-          "The system does not boot if these requirements are not met."
+          "The secure boot IPL works only on IBM z15, IBM LinuxONE III or later.\n" \
+          "NVMe disks works from IBM z15.\n" \
+          "FC-attached SCSI disks needs at least IBM LinuxONE III.\n" \
+          "ECKD DASDs with CDL layout works on IBM z16 or newer.\n" \
+          "If these requirements are not met, the system will not IPL in secure mode."
         )
       )
     end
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/grub2_widgets.rb
```diff
@@ -359,8 +359,10 @@ def help
           "<p><b>Secure Boot Support</b> if checked enables Secure Boot support.<br>" \
           "This does not turn on secure booting. " \
           "It only switches to the new secure-boot enabled boot data format. " \
-          "Note that this new format works only on z15 or later and only for zFCP disks. " \
-          "The system does not boot if these requirements are not met.</p>"
+          "Note that this new format works only on z15 or later and "\
+          "disk type supports depends on hardware. " \
+          "For details consult requirements at " \
+          "https://www.ibm.com/docs/en/linux-on-systems?topic=introduction-requirements</p>"
```
joseivanlopez, line 365: I would rephase it a bit:

```suggestion
          "Note that this new format works only on z15 or later and "\
          "only for some types of disk. The supported disks depends on " \
          "the hardware. For more details see the requirements at " \
          "https://www.ibm.com/docs/en/linux-on-systems?topic=introduction-requirements</p>"
```

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -388,9 +390,11 @@ def validate
...
       Yast::Popup.ContinueCancel(
         _(
-          "The new secure-boot enabled boot data format works only on z15 " \
-          "and later and only for zFCP disks.\n\n" \
-          "The system does not boot if these requirements are not met."
+          "The secure boot IPL works only on IBM z15, IBM LinuxONE III or later.\n" \
+          "NVMe disks works from IBM LinuxONE III.\n" \
+          "FC-attached SCSI disks needs at least IBM LinuxONE III or a IBM z15.\n" \
+          "ECKD DASDs with CDL layout works on IBM z16, LinuxONE 4 or newer.\n" \
+          "If these requirements are not met, the system will not IPL in secure mode."
```
joseivanlopez, line 397: And here, what about something like this?

```suggestion
          "The secure boot IPL works only on IBM z15, IBM LinuxONE III or later.\n" \
          "And also note the following restrictions:" \
          "NVMe disk works only from IBM LinuxONE III.\n" \
          "FC-attached SCSI disk needs at least IBM LinuxONE III or a IBM z15.\n" \
          "ECKD DASD with CDL layout works on IBM z16, LinuxONE 4 or newer.\n" \
          "If these requirements are not met, then the system will not boot in\n" \
          "secure mode."
```

src/lib/bootloader/grub2_widgets.rb
mvidner, line 365:
```suggestion
          "Note that this new format works only on z15 or later and "\
          "only for some disk types. " \
          "For more details see the requirements at  " \
          "https://www.ibm.com/docs/en/linux-on-systems?topic=introduction-requirements</p>"
```

src/lib/bootloader/grub2_widgets.rb
mvidner, line 397:
```suggestion
          "The secure boot IPL works only on IBM z15, IBM LinuxONE III or later.\n" \
          "Also note the following restrictions:" \
          "NVMe disks work since IBM LinuxONE III.\n" \
          "FC-attached SCSI disks need at least IBM LinuxONE III or IBM z15.\n" \
          "ECKD DASDs with CDL layout work on IBM z16, LinuxONE 4 or newer.\n" \
          "If these requirements are not met, the system will not IPL in secure mode."
```

(Grammar: "disk**s** need**∅**" and IMHO plural is better here)

joseivanlopez (approved):
LGTM

PR #700, opened 2024-04-26
Kernel parameters for systemd-boot

schubi2:
## Kernel parameters for systemd-boot

Up to now no kernel parameters can be defined if systemd-boot is used.

- https://bugzilla.opensuse.org/show_bug.cgi?id=1220892

## Solution

Changed bootloader UI.

## Testing

- Updated unit test
- Tested manually

### DIFF
--- a/src/lib/bootloader/bootloader_base.rb
+++ b/src/lib/bootloader/bootloader_base.rb
@@ -7,8 +7,6 @@ Yast.import "BootStorage"
 Yast.import "Linuxrc"
 Yast.import "Mode"
 Yast.import "Package"
-Yast.import "Kernel"
-Yast.import "BootStorage"
 
 module Bootloader
   # Represents base for all kinds of bootloaders
@@ -100,24 +98,6 @@ module Bootloader
     end
 
   private
-
-    def propose_resume
-      swap_parts = Yast::BootStorage.available_swap_partitions
-      largest_swap_name, lagest_swap_size = (swap_parts.max_by { |_part, size| size } || [])
-
-      propose = Yast::Kernel.propose_hibernation? && largest_swap_name
-
-      return "" unless propose
-
-      if lagest_swap_size < Yast::BootStorage.ram_size
-        log.info "resume parameter is not added because swap (#{largest_swap_name}) is too small"
-
-        return ""
-      end
-
-      # try to use label or udev id for device name... FATE #302219
-      UdevMapping.to_mountby_device(largest_swap_name)
-    end
 
     # @return [Boolean] true when kexec-tools package should be included; false otherwise
     def include_kexec_tools_package?
--- a/src/lib/bootloader/bootloader_base.rb
+++ b/src/lib/bootloader/bootloader_base.rb
@@ -7,6 +7,8 @@ Yast.import "BootStorage"
 Yast.import "Linuxrc"
 Yast.import "Mode"
 Yast.import "Package"
+Yast.import "Kernel"
+Yast.import "BootStorage"
 
 module Bootloader
   # Represents base for all kinds of bootloaders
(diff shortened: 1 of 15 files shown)


src/lib/bootloader/bootloader_base.rb
```diff
@@ -99,6 +101,24 @@ def merge(other)
...
 
+    def propose_resume
+      swap_parts = Yast::BootStorage.available_swap_partitions
+      largest_swap_name, lagest_swap_size = (swap_parts.max_by { |_part, size| size } || [])
+
+      propose = Yast::Kernel.propose_hibernation? && largest_swap_name
+
+      return "" unless propose
+
+      if lagest_swap_size < Yast::BootStorage.ram_size
```
jreidinger, line 112: Typo `lagest` instead of `largest`

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -1070,7 +1012,7 @@ def contents
         VBox(
           VSpacing(1),
           MarginBox(1, 0.5, KernelAppendWidget.new),
```
jreidinger, line 1014: Here it should also use Grub2KernelAppendWidget, not?

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -269,26 +228,9 @@ def store
     end
 
     # represents kernel command line
-    class KernelAppendWidget < CWM::InputField
+    class Grub2KernelAppendWidget < KernelAppendWidget
```
jreidinger, line 231: inconsistent naming, one is called Grub and here it is Grub2

src/lib/bootloader/bootloader_base.rb
```diff
@@ -99,6 +101,24 @@ def merge(other)
 
   private
 
+    def propose_resume
```
jreidinger, line 104: as it is not specific to bootloader and it is more about finding suitable resume swap partition I would move whole code to `Yast::BootStorage`

src/lib/bootloader/systemdboot_widgets.rb
```diff
@@ -105,6 +106,38 @@ def store
...
+    # Represents decision if smt is enabled
+    class SdCpuMitigationsWidget < CpuMitigationsWidget
+      include SystemdBootHelper
+
+      def init
+        if systemdboot.respond_to?(:cpu_mitigations)
+          self.value = systemdboot.cpu_mitigations.value.to_s
+        else
+          # do not crash when use no bootloader. This widget is also used in security dialog.
+          # (bsc#1184968)
```
jreidinger, line 118: yeah and this shows that it can lead to potential issues with naming as new name is used in security dialog. So renaming security_boot is problematic and can be even more when we do not know if it is systemd boot or grub2. And also code is almost identical. So what I proper here is to not have divided widgets  which differes just in name of bootloader nad instead of including that helpers define something like:
```
def settings
  case BootloaderFactory.current
  when Grub2Base then BootloaderFactory.current.grub_default
  when SystemdBoot then BootloaderFactory.current
  else
     nil
  end
end
```
and in code have something like
```
def init
  if settings
    self.value =settings.cpu_mitigations.value.to_s
  else
    disable
  end
end
```

and similar for rest of code. Basically allow widget to work with both dynamically instead of copying code here and there.
```

```

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -123,50 +123,9 @@ def store
     end
 
     # Represents decision if smt is enabled
-    class CpuMitigationsWidget < CWM::ComboBox
+    class GrubCpuMitigationsWidget < CpuMitigationsWidget
```
jreidinger, line 126: and btw if you do as I suggested below with single widget in generic widgets, then to avoid changes in yast2-installation you can add just alias here with simple
```
CpuMitigationsWidget = ::Bootloader::CpuMitigationsWidget
```
and it will stay backward compatible :wink:

src/lib/bootloader/grub2_widgets.rb
schubi2: That's a BIG bug. Thanks !

src/lib/bootloader/bootloader_base.rb
schubi2: Renamed

src/lib/bootloader/systemdboot_widgets.rb
schubi2: Hm, I have not managed to go exactly that way, but I hope that I have mainly hit your requirements.

src/lib/bootloader/grub2_widgets.rb
schubi2: Yes, I have "unified" it. Good idea .....

jreidinger (approved):

PR #701, opened 2024-05-03
Writing "root" entry into /etc/kernel/cmdline which is needed by sdbootutil

schubi2:
* sdbootutil needs the root entry in /etc/kernel/cmdline in order to write the correct bootloader menu entry.
* /etc/kernel/cmdline has to be written before sdbootutil will be called.

## Testing

- Adapted test cases.
- Tested manually in MicroOS(including transactional-update) and TW (while installation and in the installed system).

### DIFF
--- a/src/lib/bootloader/systemdboot.rb
+++ b/src/lib/bootloader/systemdboot.rb
@@ -205,7 +205,11 @@ module Bootloader
     def create_menue_entries
       # writing kernel parameter to /etc/kernel/cmdline
       File.open(File.join(Yast::Installation.destdir, CMDLINE), "w+") do |fw|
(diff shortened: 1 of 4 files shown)


src/lib/bootloader/systemdboot.rb
```diff
@@ -227,7 +220,6 @@ def create_menue_entries
                  ), command: e.commands.inspect, stderr: e.stderr)
         )
       end
-      File.delete(cmdline_file) if Yast::Stage.initial # see above
```
joseivanlopez, line 230: Has the file to be kept?

jreidinger (approved):

PR #702, opened 2024-07-19
Rename menue_timeout to menu_timeout

aplanas:
I think that Menü has been used as menue in the code as a German translation, creating a variable menue_timeout. Also menue is used in error messages.

This commit rename menue to menu, including the error messages presented to the user.


aplanas:
cc @schubi2

jreidinger (approved):
looks good to me. As schubi do not complain I will merge it.

jreidinger:
Thanks for patch

PR #703, opened 2024-08-14
Disable secure boot 32b arm

jreidinger:
## Problem

32 bit arm and also riscv64 does not have shim and due to this we do not support secure boot on them (as opposite to s390 and ppc with own secure boot way ). But bootloader propose secure boot there, resulting in perl-bootloader failure.

https://bugzilla.suse.com/show_bug.cgi?id=1229070

## Solution

disable secure boot support on given architectures.

## Testing

- *Added a new unit test*


ancorgs (approved):
LGTM

PR #704, opened 2024-11-05
S390 sec boot2

jreidinger:
## Problem

The text in proposal popup regarding s390 is still not precise. On running system it is correct.

## Solution

Keep the text in sync and also note it in code to avoid such issue next time.

## Testing

- no testing as it is just text change and text is identical and both parts use dialog, so it should also fit.


shundhammer (approved):
LGTM

PR #707, opened 2024-11-13
Supporting Grub2-BLS

schubi2:
Added grub2-bls support which has been announced:
https://news.opensuse.org/2024/10/08/grub2-bls

[jsc#PED-10703](https://jira.suse.com/browse/PED-10703)

## Testing

- Added new unit tests
- Tested manually


src/lib/bootloader/generic_widgets.rb
```diff
@@ -73,20 +72,6 @@ def handle
         return :redraw if !Yast::Popup.ContinueCancel(popup_msg)
       end
 
-      if new_bl == "systemd-boot"
-        # popup - Continue/Cancel
-        popup_msg = _(
-          "\n" \
-          "Systemd-boot support is currently work in progress and\n" \
```
schubi2, line 80: We have implemented systemd-boot last year and have not gotten strong errors.
As our team (Ähm, mainly I am :-) ) is maintainer of the systemd-boot part in yast-bootloader we would suggest to remove this warning.

PR #708, opened 2024-11-14
Supporting Grub2-BLS

schubi2:
Added grub2-bls support which has been announced:
https://news.opensuse.org/2024/10/08/grub2-bls

[jsc#PED-10703](https://jira.suse.com/browse/PED-10703)

## Testing

- Added new unit tests

- Tested manually
-- New installation
-- In installed system
-- AutoYaST installation


src/lib/bootloader/generic_widgets.rb
```diff
@@ -73,20 +72,6 @@ def handle
         return :redraw if !Yast::Popup.ContinueCancel(popup_msg)
       end
 
-      if new_bl == "systemd-boot"
-        # popup - Continue/Cancel
-        popup_msg = _(
-          "\n" \
-          "Systemd-boot support is currently work in progress and\n" \
```
schubi2, line 80: We have implemented systemd-boot last year and have not gotten strong errors.
As our team (Ähm, mainly I am :-) ) is maintainer of the systemd-boot part in yast-bootloader we would suggest to remove this warning.

src/lib/bootloader/grub2bls.rb
```diff
@@ -0,0 +1,193 @@
...
+        "set", "timeout=#{grub_default.timeout}")
+    end
+
+    def merge_sections(other)
+      return if !other.sections.default || other.sections.default.empty?
+
+      @sections.default = other.sections.default
+    end
+
+    def create_menu_entries
```
aplanas, line 168: Most of this code can be shared to the systemd-boot case, as they are using the same interface (sdbootutil).  Would make case to unify it?

src/lib/bootloader/grub2bls.rb
```diff
@@ -0,0 +1,193 @@
...
+      log.info "merging result: timeout: #{grub_default.timeout}"
+      log.info "                mitigations: #{cpu_mitigations.to_human_string}"
+      log.info "                kernel_params: #{grub_default.kernel_params.serialize}"
+    end
+    # rubocop:enable Metrics/AbcSize
+
+    # @return [Array<String>] packages required to configure given bootloader
+    def packages
+      res = super
+      res << ("grub2-" + Yast::Arch.architecture + "-efi-bls")
```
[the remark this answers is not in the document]
schubi2: From my point of view it is a packaging issue. Each "grub2-<Yast::Arch.architecture>-efi-bls" package should provide a "grub2-efi-bls" provides and the solver should pick the right one (depending on architecture)

src/lib/bootloader/grub2bls.rb
```diff
@@ -0,0 +1,193 @@
...
+      res << "sdbootutil"
+      res << "grub2"
```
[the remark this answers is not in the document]
schubi2: I have needed for the "fallback" (described above) in order to call grub2-editenv. You are right, with the removing of the fallback I can remove it too.

src/lib/bootloader/bls_sections.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    # write default entry
+    def write_default
+      ret = Yast::Execute.on_target("/usr/bin/sdbootutil",
+        "set-default", @default,
+        allowed_exitstatus: [0, 1])
+
+      return unless ret != 0
+
+      # fallback directly over grub2-editenv
```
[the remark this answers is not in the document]
schubi2: I have removed it.

src/lib/bootloader/grub2bls.rb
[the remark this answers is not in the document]
schubi2: removed

src/lib/bootloader/bootloader_factory.rb
```diff
@@ -24,6 +25,7 @@ class BootloaderFactory
     # Keyword used in autoyast for default bootloader used for given system.
     DEFAULT_KEYWORD = "default"
     SYSTEMDBOOT = "systemd-boot"
+    GRUB2BLS = "grub2-bls"
```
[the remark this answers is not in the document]
schubi2: It will not be shown in the selection list. Is this wrong ?

src/lib/bootloader/bls_sections.rb
```diff
@@ -0,0 +1,95 @@
...
+      Yast::Execute.on_target("/usr/bin/grub2-editenv", grubenv_path,
+        "set", "default=" + @default)
+    end
+
+    # @return [Array] return array of entries or []
+    def read_entries
+      output = Yast::Execute.on_target(
+        "/usr/bin/bootctl", "--json=short", "list", stdout: :capture
```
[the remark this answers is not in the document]
aplanas: The title is presented with `sdbootutil -v list-entries`

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -1055,66 +1064,82 @@ def loader_location_widget?
       end
 
       def generic_mbr_widget?
-        (Yast::Arch.x86_64 || Yast::Arch.i386) && grub2.name != "grub2-efi"
+        (Yast::Arch.x86_64 || Yast::Arch.i386) && !["grub2-efi", "grub2-bls"].include?(grub2.name)
       end
 
       def secure_boot_widget?
-        Systeminfo.secure_boot_available?(grub2.name)
+        Systeminfo.secure_boot_available?(grub2.name) && grub2.name != "grub2-bls"
```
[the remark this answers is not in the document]
schubi2: Michael Chang has written:
---I don’t think we need "secure boot" since it’s specific to ordinary grub----
That's why I have not implemented. But we still can do it afterwards ;-)

src/lib/bootloader/bootloader_factory.rb
```diff
@@ -108,6 +114,10 @@ def use_systemd_boot?
           (Yast::Arch.x86_64 || Yast::Arch.aarch64) # only these architectures are supported.
       end
 
+      def use_grub2_bls?
+        (Yast::Arch.x86_64 || Yast::Arch.aarch64) # only these architectures are supported.
```
jreidinger, line 118: so no plan yet to try it on s390x? I read a bit that IBM people trying somehow BLS on s390

src/lib/bootloader/generic_widgets.rb
```diff
@@ -43,6 +43,7 @@ def localized_names(name)
       names = {
         "grub2"        => _("GRUB2"),
         "grub2-efi"    => _("GRUB2 for EFI"),
+        "grub2-bls"    => _("GRUB2 with BLS"),
```
jreidinger, line 46: I think it would deserve comment for translators as BLS akak BootLoader Specification is not much widely known term

src/lib/bootloader/bootloader_factory.rb
aplanas: The bootloader team is working on s390, but for now makes sense to make a step in the grub2-bls direction for intel and arm

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -1055,66 +1064,82 @@ def loader_location_widget?
...
       end
 
       def trusted_boot_widget?
-        Systeminfo.trusted_boot_available?(grub2.name)
+        Systeminfo.trusted_boot_available?(grub2.name) && grub2.name != "grub2-bls"
       end
 
       def update_nvram_widget?
-        Systeminfo.nvram_available?(grub2.name)
+        Systeminfo.nvram_available?(grub2.name) && grub2.name != "grub2-bls"
```
jreidinger, line 1079: in general I do not like to have it in widgets...this check for grub2.name should be done in given Systeminfo modules which do decision in single place regarding archs and bootloader types.

src/lib/bootloader/grub2base.rb
```diff
@@ -435,14 +435,6 @@ def update_nvram_summary
 
       "#{_("Update NVRAM:")} #{status_string(update_nvram)} #{link}"
     end
-
-    def status_string(status)
```
jreidinger, line 439: ugh, why it is removed? I see it is used upper.

jreidinger:
I add some comments that needs to be clarified...but more importantly to make it work I worry it needs also changes in yast2-storage-ng as currently EFI partition is 256MiB which is quite small for BLS which in the latest specification mention EFI of 1GiB to be able to have inside enough kernels.
We discuss it on daily and it will be problematic to increase it unconditionally for all installation (users already complain it is too big and on aarch with some small storage it is even more critical), so we probably need some flag for installation to adapt storage proposal if given flag is given....flag like passing `bls=1` to kernel command line....and in such case it should be possible to even change default bootloader to grub2-bls or systemd boot when such flag is specified.

aplanas:
> I add some comments that needs to be clarified...but more importantly to make it work I worry it needs also changes in yast2-storage-ng as currently EFI partition is 256MiB which is quite small for BLS which in the latest specification mention EFI of 1GiB to be able to have inside enough kernels.

This 1GiB is for UKIs, and we are not using it now. For the current systemd-boot (already merged) and this grub2-bls, 512MiB should be OK (is what MicroOS is using). The kernel and the initrds are reused, so we can have multiple active snapshots represented in the ESP.  But this decision should be done by the bootloader team, in case that they have a different vision here.

About the mechanism ... is this bls=1 set by the user? That is dangerous IMHO, as I select the bootloader much later

jreidinger:
> > I add some comments that needs to be clarified...but more importantly to make it work I worry it needs also changes in yast2-storage-ng as currently EFI partition is 256MiB which is quite small for BLS which in the latest specification mention EFI of 1GiB to be able to have inside enough kernels.
> 
> This 1GiB is for UKIs, and we are not using it now. For the current systemd-boot (already merged) and this grub2-bls, 512MiB should be OK (is what MicroOS is using). The kernel and the initrds are reused, so we can have multiple active snapshots represented in the ESP. But this decision should be done by the bootloader team, in case that they have a different vision here.
> 
> About the mechanism ... is this bls=1 set by the user? That is dangerous IMHO, as I select the bootloader much later

As said, now it is 256MiB with 128MiB minimum, which is too low from my understanding of BLS. So it will need to adapt to whatever value and as mentioned even with that 256MiB users complain that it is too much space.

mechanism is like with fips. It is set on kernel cmdline during boot of medium and then it changes what installer propose. In agama it will be more straightforward as it do not use wizard approach, but yast installer use wizard and storage is done very soon and bootloader later. So idea is that user set bls=1 on kernel cmdline and it will propose big enough EFI partition to fit BLS based bootloader and then user can change it to even common one, but EFI proposed size will remain.

aplanas:
> So idea is that user set bls=1 on kernel cmdline and it will propose big enough EFI partition to fit BLS based bootloader and then user can change

That is the part that does not make sense to me. The selection of the bootloader comes later. I select grub2-bls or systemd-boot in YaST with some graphical interface, and I would expect that the proposal changes to the expected sizes and layout that make sense from my selection.

Maybe something that makes a bit more sense, from the user PoV, is that if the system has EFI then the selected layout would be such that mount the ESP in /boot/efi with "enough" size, whatever means means for this version of the installer.

I do not know if the XML that defines the disk layout and some of the defaults (the skel file) can condition the size of the ESP under the condition of the selected boot loader, but this can be another path that can work.

ancorgs:
> That is the part that does not make sense to me. The selection of the bootloader comes later. I select grub2-bls or systemd-boot in YaST with some graphical interface, and I would expect that the proposal changes to the expected sizes and layout that make sense from my selection.

That's a historical problem we have with YaST that makes hard to fit this case.

YaST is a wizard and the storage (partitioning) setup is one of the first steps that is configured by the user. But tuning the configuration of the boot loader is something that gets done almost at the end of the process, right before starting the installation. The partitioning setup has already been reviewed, adapted and accepted by the user, we cannot simply change it (for a variety of reasons).

> Maybe something that makes a bit more sense, from the user PoV, is that if the system has EFI then the selected layout would be such that mount the ESP in /boot/efi with "enough" size, whatever means means for this version of the installer.

That means we grow the EFI for everyone. Both people using the default boot loader and those using Grub2-BLS. I can tell you there will be a lot of people unhappy with that. For whatever reason, people tend to be very sensitive about having too big UEFI partitions (for their own individual definitions of "too big").

> I do not know if the XML that defines the disk layout and some of the defaults (the skel file) can condition the size of the ESP under the condition of the selected boot loader, but this can be another path that can work.

The size of those boot-related partitions are calculated based on different so-called "strategies". We have the following strategies: Legacy (for BIOS-based systems), NFSRoot (for systems without local storage), PReP (for powerPC systems), Raspi (for Raspberry Pi), UEFI (for standard EFI-based systems) and Zipl (for s390). We add new strategies as we need to cover more scenarios.

We could perfectly develop a separate strategy for Grub2-BLS+EFI. But the original problem remains - the installer needs to choose a strategy and stick to it. And the selection happens during the step that sets the initial storage configuration. So we use criteria like the architecture, the product, the location of "/" and any other aspect that is known at that point. If we are going to use the chosen boot-loader as a criteria to choose the strategy, then such a decision needs to be known during the partitioning step.

src/lib/bootloader/grub2bls.rb
schubi2: I have moved all sdbootutil calls to one class bls.rb

src/lib/bootloader/grub2base.rb
schubi2: I have moved it to bootloader_base.rb because it will be used by different bootloaders now.

src/lib/bootloader/grub2_widgets.rb
```diff
@@ -1055,66 +1064,82 @@ def loader_location_widget?
...
       end
 
       def pmbr_widget?
         (Yast::Arch.x86_64 || Yast::Arch.i386) &&
-          Yast::BootStorage.gpt_boot_disk?
+          Yast::BootStorage.gpt_boot_disk? &&
+          grub2.name != "grub2-bls"
```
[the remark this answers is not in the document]
schubi2: @aplanas Alberto, what do you think about it ?

src/lib/bootloader/bls.rb
```diff
@@ -0,0 +1,105 @@
...
+      format(_(
+               "Cannot install bootloader:\n" \
+               "Command `%{command}`.\n" \
+               "Error output: %{stderr}"
+             ), command: e.commands.inspect, stderr: e.stderr)
+    )
+    end
+
+    def self.write_menu_timeout(timeout)
+      Yast::Execute.on_target(SDBOOTUTIL, "set-timeout", timeout)
```
jreidinger, line 50: here and also below this code it is using non`!` varient. It means it will return nil and not raise exception, but the code below expect it. See https://github.com/yast/yast-yast2/blob/717d99f214a56645216bb9c5f4f32d1244d4fe35/library/system/src/lib/yast2/execute.rb#L69 for diff between `on_target` and `on_target!`

schubi2: Ups, yes you are right ! Fixed.

src/lib/bootloader/grub2_widgets.rb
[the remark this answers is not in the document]
aplanas: In this case grub2-efi, grub2-bls and systemd-boot should share this common flag. IIUC this is used to protect the GPT partition table from some tools that change the MBR, so would make sense to do this for all the EFI bootloaders

thkukuk:
> As said, now it is 256MiB with 128MiB minimum, which is too low from my understanding of BLS. So it will need to adapt to whatever value and as mentioned even with that 256MiB users complain that it is too much space.

Of course users do complain **today** that 256MB is too much space if you only use 6MB.
But: with SLFO we don't support legacy boot anymore for SL Micro, only EFI. SLES16 may be different because of VMs and Cloud, but I was told that even there everything is now finally moving to EFI for a good reason: Confidential Computing.

With FDE+TPM, which should become the default for our products, BLS is mandatory.
So **after** this change, people will no longer complain that we waste that many space, because it is in use. And using `/boot/efi` means, we need less space in `/boot`, so `/`.

And we need 1GB for `/boot/efi`, looking at my reference machines with daily updates, Tumbleweed uses 480MB, no other OS in parallel installed.

src/lib/bootloader/grub2_widgets.rb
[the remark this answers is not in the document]
schubi2: OK, thanks. So they are supporting it too now. Fixed.

schubi2:
> 2. If the user is in a supported architecture for grub2-bls (the other one is aarch64), and select the grub2-bls bootloader, present some complain about the EFI partition size

For the first step I would prefer this solution. While implementation and testing I have not had any issue regarding disk space. So I am wondering why we are insisting on the 1GByte size so strictly.

aplanas:
> why we are insisting on the 1GByte size so strictly.

The current model (shared kernel and initrd between snapshots, as sdbootutil uses) do not requires 1Gb of EFI space, but if the idea is moving in the future to UKIs, then this size seems more reasonable.

UKIs can be shared now, as certain components can live outside them with new systemd versions (like the cmdline), but I suspect that the re-usage index will be lower than the current approach.

jreidinger:
Also 1GiB is specified in BLS itself even thought quite vague as `Otherwise, if on GPT and an ESP is found and it is large enough (let’s say at least 1G) it should be used as $BOOT and used as primary location to place boot loader menu resources in.` ( source https://uapi-group.org/specifications/specs/boot_loader_specification/ )

ancorgs:
> Can next idea fit in YaST?

It's surely doable and sounds like a sensible approach to move forward.

>     1. Make this Grub2-BLS strategy, and enable if by default if we are in Intel (for now) with UEFI. This will select the default EFI partition for 1Gb and select the grub2-bls bootloader by default.

For the yast2-storage-ng part, that's just a matter of adding the related logic to https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/boot_requirements_checker.rb#L125

For the yast2-bootloader part I guess it's also possible to define the default bootloader using any reasonable criteria (I'm not that familiar with that part of the code).

>     2. If the user is in a supported architecture for grub2-bls (the other one is aarch64), and select the grub2-bls bootloader, present some complain about the EFI partition size

That should be doable as well. We would need to decide which mechanism to use for that check, but it doesn't sound hard to implement.

jreidinger:
For bootloader changing default bootloader is mainly about changing logic at https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/bootloader_factory.rb#L116 which is the part of code that is responsible for it.

ancorgs:
> [...] that's just a matter of adding the related logic to [...]

One important note. That logic to decide whether to go for grub2-bls should be configurable by "product". I'm pretty sure we don't want to change yet the behavior for SLES-16 or other enterprise products, neither for Agama-based installations (until we have tried it in YaST), etc.

So, at least for the time being, we should likely introduce a new configuration option. That would allow us to specify if we want to "try" (if all the other conditions are met) grub2-bls for a given product (only TW installed with YaST, in this case) or even per system role.

thkukuk:
> > [...] that's just a matter of adding the related logic to [...]
> 
> One important note. That logic to decide whether to go for grub2-bls should be configurable by "product". I'm pretty sure we don't want to change yet the behavior for SLES-16

Yes, we **want to change** the behavior for SLES-16 **now**: https://jira.suse.com/browse/PED-10703
It doesn't make any sense to do that only after the Alpha phase, this are things you are doing in the development/Alpha phase.

jreidinger:
> > > [...] that's just a matter of adding the related logic to [...]
> > 
> > 
> > One important note. That logic to decide whether to go for grub2-bls should be configurable by "product". I'm pretty sure we don't want to change yet the behavior for SLES-16
> 
> Yes, we **want to change** the behavior for SLES-16 **now**: https://jira.suse.com/browse/PED-10703 It doesn't make any sense to do that only after the Alpha phase, this are things you are doing in the development/Alpha phase.

So if I get that jira entry properly it is requested to have big enough EFI partition for x86_64 and aarch64 and to also have 1GiB /boot/zipl on s390 to allow future transition. And also no change on ppc. Do I get it correctly?
And having default bootloader according to product as for SLES and SLES 4 SAP it should not be default, but for micro and TW it should be default bootloader on supported architectures?

thkukuk:
> So if I get that jira entry properly it is requested to have big enough EFI partition for x86_64 and aarch64 and to also have 1GiB /boot/zipl on s390 to allow future transition. And also no change on ppc. Do I get it correctly? 

Ignore s390x for now. That's still not clearified and in evaluation.

> And having default bootloader according to product as for SLES and SLES 4 SAP it should not be default, but for micro and TW it should be default bootloader on supported architectures?

PM does not want a difference in SL Micro 6.x and SLES16, both should use the same bootloader setup.
But to answer all this questions: we need that in Tumbleweed now. Without that, there is no way to carry out a reliable evaluation.
And sooner or later we need BLS in SL Micro 6.x and SLES16 for x86-64 and aarch64. There is no way around that. Only question would be, if we use grub2-BLS or systemd-boot...

schubi2:
I will try to go Ancors way, defined in:

>For the yast2-storage-ng part, that's just a matter of adding the related logic to https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/boot_requirements_checker.rb#L125


> PM does not want a difference in SL Micro 6.x and SLES16, both should use the same bootloader setup. But to answer all this questions: we need that in Tumbleweed now. Without that, there is no way to carry out a reliable evaluation. And sooner or later we need BLS in SL Micro 6.x and SLES16 for x86-64 and aarch64. There is no way around that. Only question would be, if we use grub2-BLS or systemd-boot...

As Josef has already mentioned, the question could be answered around here:

https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/bootloader_factory.rb#L116

src/lib/bootloader/generic_widgets.rb
```diff
@@ -171,6 +164,57 @@ def store
...
+        [:nothing, _("do not change")]
+      ]
+    end
+
+    def store
+      current_bl = ::Bootloader::BootloaderFactory.current
+      if current_bl.respond_to?(:pmbr_action)
+        current_bl.pmbr_action = value
+      else
+        log.error("Bootloader #{current_bl} does not support PMBR.")
```
jreidinger, line 213: JFYI PMBR is not about bootloader support as it is set by parted. For more info https://uefi.org/specs/UEFI/2.10/05_GUID_Partition_Table_Format.html#protective-mbr Basically it is for backward compatibily for stuff that do not understand GPT and needs MS DOS table. But it has some disadvantages, so for EFI boot we usually unset it.

src/lib/bootloader/systeminfo.rb
```diff
@@ -62,6 +91,9 @@ def trusted_boot_active?
 
       # Check if the system is expected to have nvram - ie. update_nvram_active? makes a difference
       def nvram_available?(bootloader_name = nil)
+        # not for grub2-bls
+        return false if bootloader_name == "grub2-bls"
```
jreidinger, line 95: @aplanas Just to be sure, grub2-bls cannot skip writing to nvram? I remember that especially some aarch64 devices has problems to write to it and needs this "workaround" which skips write to it, otherwise a crash happen. I can find some bugzilla references, if you are interested.

src/lib/bootloader/systeminfo.rb
```diff
@@ -108,6 +142,30 @@ def efi_mandatory?
         Yast::Arch.aarch64 || Yast::Arch.arm || Yast::Arch.riscv64
       end
 
+      # Check if console settings are supported
+      #
+      # param bootloader_name [String] bootloader name
+      # @return [Boolean] true if supported
+      def console_supported?(bootloader_name)
+        !Yast::Arch.s390 && bootloader_name != "grub2-bls"
```
jreidinger, line 150: @aplanas Just to be sure grub2-bls does not support neither graphical console neither serial console? Especially later one can be a bit problematic on servers. Or it works in different way and is just not configurable and detect serial console automatic?

src/lib/bootloader/systeminfo.rb
```diff
@@ -108,6 +142,30 @@ def efi_mandatory?
...
+      def hiding_menu_supported?(bootloader_name)
+        bootloader_name != "grub2-bls"
+      end
+
+      # Check if setting password is supported
+      #
+      # param bootloader_name [String] bootloader name
+      # @return [Boolean] true if supported
+      def password_supported?(bootloader_name)
+        bootloader_name != "grub2-bls"
```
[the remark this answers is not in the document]
aplanas: grub2-bls can be extended via grubenv block, but I do not see there a way to set passwords, so I would say that grub2-bls cannot be protected with passwords today. But mchang will give the last word here.

IIUC the password protected grub2 is about avoiding users to edit the menu entry and such. This is already covered by the TPM2 and measured boot.

schubi2:
> I will try to go Ancors way, defined in:
> 
> > For the yast2-storage-ng part, that's just a matter of adding the related logic to https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/boot_requirements_checker.rb#L125
> 
I have made a PR:
https://github.com/yast/yast-storage-ng/pull/1396/files

jreidinger:
@schubi in general I would say PR looks good, I am just waiting to get answers from @aplanas what exactly grub2-bls supports. I would say that especially those console support is important to be sure.

aplanas:
> I am just waiting to get answers from @aplanas what exactly grub2-bls supports. I would say that especially those console support is important to be sure.

What is missing from my side?

jreidinger:
@aplanas answer to this question about grub2-bls support for consoles https://github.com/yast/yast-bootloader/pull/708#discussion_r1880943473

src/lib/bootloader/systeminfo.rb
aplanas: It is supporting both

schubi2: We have discussed it again and it is not really supported. I have tried it but without really success ;-)

jreidinger: hmm, especially that serial console can be a bit problematic on SLES as it is common way to get output from headless servers.

thkukuk: Servers today have all a web based management port, which does not only allow you to select a different boot entry, but also to enter the firmware. And with EFI, serial console with bootloader is coming too late anyways, as you select the boot entry in firmware upfront.
So is this really a problem today? Serial console for watching the boot process and login should still work, that's independent of the bootloader.
Since other distributions also switch to grub2-BLS, it looks like they don't see a problem with serial console, too. If it will really become one, somebody will implement it.
So ignore serial console for now, so that we can get this in and collect feedback, what people are really missing.

src/lib/bootloader/systeminfo.rb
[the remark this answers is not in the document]
thkukuk: We should not implement what customers requested ages ago, but what is really needed today. And with EFI, a password protected bootloader does not make really sense.
With SLE Micro we did: implement the minimal set we need to have a working OS, and only add, what customers are really missing, not what got requested ages ago and nobody is using anymore.
Not that surprising: no customer asked for things where developers claimed that all customers will need it...

jreidinger (approved):

jreidinger:
OK, I approved PR to give it a try. Discussion about missing features of grub2-bls would be probably best in jira, where the decision about defaults should be taken.

jreidinger:
@schubi2 I think you should have rights to merge. If not, just ping me when you want to have it merged.

schubi2:
> @schubi2 I think you should have rights to merge. If not, just ping me when you want to have it merged.

@jreidinger Thanks you Josef for your support ! I have merged it.

PR #709, opened 2024-12-20
Filter out root

jreidinger:
## Problem

In agama QA using PXE with root parameter to point to live image. After installation it pass "root=" parameter to target system and result in non-bootable system as it points to wrong root.

## Solution

Always filter out "root=" parameter as it points to installation system and not target one.

## Testing

- *Added a new unit test*


kobliha (approved):
Looks nice :) Thx!

PR #713, opened 2025-03-18
Supporting LUKS2 encryption by TPM2 device (jsc#PED-10703)

schubi2:
## Problem
We would like to give the user an easy way to use TPM2 devices for LUKS2 encrypted
partitions. (jsc#PED-10703)

The user can select TPM2 support for encryption in yast-storage-ng:
https://github.com/yast/yast-storage-ng/pull/1406

## Solution

Enrolling TPM2 with the given password, if the user has selected it.
Selecting yast2-bls bootloader which supports TPM2

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots
tpm2_3

### DIFF
--- a/src/lib/bootloader/bls.rb
+++ b/src/lib/bootloader/bls.rb
@@ -126,18 +126,8 @@ module Bootloader
 
       Yast::Execute.on_target("/usr/bin/sdbootutil",
                               "enroll", "--method=tpm2")
-      begin
-        Yast::Execute.on_target!("/usr/bin/sdbootutil",
-                                 "enroll", "--method=tpm2")
-      rescue Cheetah::ExecutionFailed => e
-        Yast::Report.Error(
-          format(_(
-                   "Cannot enroll TPM2 method:\n" \
-                   "Command `%{command}`.\n" \
-                   "Error output: %{stderr}"
-                 ), command: e.commands.inspect, stderr: e.stderr)
-        )
-      end
+      Yast::Execute.on_target("/usr/bin/sdbootutil",
+                              "enroll", "--method=tpm2")
     end
   end
 end
--- a/src/lib/bootloader/bls.rb
+++ b/src/lib/bootloader/bls.rb
@@ -107,7 +107,7 @@ module Bootloader
     # Enabe TPM2, if it is required
     def self.enable_tpm2
       return unless Y2Storage::StorageManager.instance.proposal
-      return unless Y2Storage::StorageManager.instance.proposal.settings.encryption_use_tpm2
+      return unless Y2Storage::StorageManager.instance.encryption_use_tpm2
 
       begin
         Yast::Execute.on_target!("keyctl", "padd", "user", "cryptenroll",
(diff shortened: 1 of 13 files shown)


src/lib/bootloader/bls.rb
```diff
@@ -99,7 +103,66 @@ def self.default_menu
         )
         output = ""
       end
-      output
+      output.strip
+    end
+
+    # Enabe TPM2, if it is required
```
jreidinger, line 109: typo `Enabe`

src/lib/bootloader/bls_sections.rb
```diff
@@ -16,27 +16,35 @@ class BlsSections
...
 
+    # @return [String] title of default boot section.
+    def default
+      return unless @data
+
+      entry = @data.select { |d| d["id"] == @default }
+      if entry.empty?
+        ""
+      else
+        entry.first["title"]
```
jreidinger, line 32: this looks strange as you basically search for first entry, so it should look like
```ruby
entry = @data.find { |d| d["id"] == @default }
entry ? entry["title"] : ""
```

src/lib/bootloader/bls_sections.rb
```diff
@@ -16,27 +16,35 @@ class BlsSections
...
     # @note to write it to system use #write later
     def default=(value)
-      log.info "set new default to '#{value.inspect}'"
-
-      # empty value mean no default specified
-      if !all.empty? && !all.include?(value) && !value.empty?
-        log.warn "Invalid value #{value} trying to set as default. Fallback to default"
-        value = ""
+      entry = @data.select { |d| d["title"] == value }
+      if entry.empty?
```
jreidinger, line 41: same here. The used method should not be select, but find that return first entry or nil.

src/lib/bootloader/grub2bls.rb
```diff
@@ -155,7 +157,7 @@ def packages
       res = super
       res << ("grub2-" + grub2bls_architecture + "-efi-bls")
       res << "sdbootutil"
-      res << "shim" if secure_boot
+      res << "shim"
```
jreidinger, line 160: why this change is needed? I think there are architectures where shim is no available, but as it is grub2-bls architectures are already limited, os not sure.

src/modules/BootSupportCheck.rb
```diff
@@ -174,26 +176,39 @@ def check_mbr
       false
     end
 
+    def check_tpm2
+      return true unless Y2Storage::StorageManager.instance.encryption_use_tpm2
+
+      add_new_problem(_("This bootloader cannot handle encryption supported by a TPM2 device. "))
```
jreidinger, line 182: so even our patched grub2 cannot handle tpm2 device?

aplanas: The text is confusing. Sure grub2 handle TPM2 devices since very long ago. I guess this text will not be present when EncryptionMethod is used.

schubi2: JFI: "EncryptionMethod" ....This will be a new PR based on the current.

schubi2:
BTW. All requested changes are going to the new PR......

PR #724, opened 2025-09-29
Bls update (WIP)

schubi2:
## Problem

*Short description of the original problem.*

- *Bugzilla link*
- *openQA link*
- *Links to other related pull requests*

## Solution

*Short description of the fix.*

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

### DIFF
--- a/src/lib/bootloader/proposal_client.rb
+++ b/src/lib/bootloader/proposal_client.rb
@@ -36,7 +36,9 @@ module Bootloader
           # TRANSLATORS: kind of boot. It is term for way how x86_64 can boot
           "grub2-efi"    => _("EFI boot"),
           # TRANSLATORS: kind of boot. It is term for way how can boot.
-          "systemd-boot" => _("Systemd boot")
+          "systemd-boot" => _("Systemd boot"),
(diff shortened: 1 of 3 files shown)


src/lib/bootloader/proposal_client.rb
```diff
@@ -36,7 +36,9 @@ def boot_map
           # TRANSLATORS: kind of boot. It is term for way how x86_64 can boot
           "grub2-efi"    => _("EFI boot"),
           # TRANSLATORS: kind of boot. It is term for way how can boot.
-          "systemd-boot" => _("Systemd boot")
+          "systemd-boot" => _("Systemd boot"),
+          # TRANSLATORS: kind of boot. It is term for way how can boot.
```
jreidinger, line 40: I would help translators with explanation what is BLS - something like `grub2 adapted to conform Boot Loader Specification (BLS)`.
And that reminds me that I do not like much that BLS boot as Systemd boot is also BLS boot...maybe something like `grub2 with BLS`  can be better?

jreidinger (approved):

PR #736, opened 2026-04-22
Remove trusted boot

mchf:
## Problem

Trusted boot was silently dropped from OS TW some time ago

- [*Bugzilla link*](https://bugzilla.suse.com/show_bug.cgi?id=1262249)


src/autoyast-rnc/bootloader.rnc
```diff
@@ -74,7 +74,6 @@ bl_global =
 
       element boot_custom { STRING }? &
       element generic_mbr { STRING }? &
-      element trusted_grub{ STRING }? &
```
jreidinger, line 77: I would not drop it as it means that any profile that has "<trusted_grub>false</trusted_grub>" will be invalid. Just add comment that option is ignored.

src/lib/bootloader/grub2base.rb
```diff
@@ -140,8 +135,7 @@ def propose
       propose_serial
       propose_xen_hypervisor
 
-      self.trusted_boot = false
-      self.secure_boot = Systeminfo.secure_boot_supported?
+      self.secure_boot = Systeminfo.secure_boot_active?
```
jreidinger: ugh, why? this change logic of proposing secure boot

src/autoyast-rnc/bootloader.rnc
mchf: yes i intentionally wanted to invoke that error ... but if you think it is bad approach I'll give it back

src/lib/bootloader/grub2base.rb
mchf: ups ... i slipped a bit ... will fix

src/autoyast-rnc/bootloader.rnc
jreidinger: well, issue with autoyast profiles is that when you generate it with clone, it contain all keys...so all cloned profile contained trusted_grub with false option.

src/autoyast-rnc/bootloader.rnc
```diff
@@ -74,6 +74,7 @@ bl_global =
 
       element boot_custom { STRING }? &
       element generic_mbr { STRING }? &
+      # trusted_grub is for backword compatibility only and is silently ignored. PED-7975
```
jreidinger: typo backward

jreidinger (approved):

PR #738, opened 2026-07-13
Dasd fba

jreidinger:
## Problem

Secure boot is also proposed for FBA DASD, which is not supported and zipl failed.

bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1270367
trello: https://trello.com/c/0dE6hwLT/5897-p2-bsc1270367-secure-boot-and-dasd

## Solution

Propose secure boot only for ECKD DASD.

## Testing

- *Added a new unit test*


src/lib/bootloader/systeminfo.rb
```diff
@@ -301,15 +301,24 @@ def scsi?(device)
...
+      # @see https://www.ibm.com/docs/en/linux-on-systems?topic=linux-secure-boot#secure_boot
       #
       # param device [Y2Storage::Partition, NilClass] partition device (or nil)
       #
       # @return [Boolean] true if device is a DASD device
       def dasd?(device)
-        device.name.start_with?("/dev/dasd")
-      rescue StandardError
-        false
+        return unless device
```
ancorgs: Being picky, a method with a name ending in `?` should always return true or false.

ancorgs (approved):
LGTM, although with a minor comment

src/lib/bootloader/systeminfo.rb
jreidinger: right, it was mistake. I plan to use return false.

